PR #17, opened 2024-01-05
deployment tests 

cooktheryan:
Test to ensure deployments are in an available state and always dump the operator logs to show any issues that may exist


cooktheryan:
@JasonPowr take a look at this and see if it covers what we walked about

JasonPowr (approved):
/lgtm

Yep, this looks good to me, handles both testing both the create cert workflow and deploying multiple instances of the operator :100:

Issue #18, opened 2024-01-05
Allow for bring your own database 


cooktheryan:
If variables are set then we should not create a db and should connect to an existing

bouskaJ:
The built-in database is automatically initialized with https://github.com/securesign/trillian/blob/redhat-v1.5/Dockerfile.database#L8. For the provided database we need to create schema with similar code https://github.com/sigstore/scaffolding/blob/main/cmd/trillian/createdb/main.go

cooktheryan:
Pr #28

cooktheryan:
closed by #30

PR #14, opened 2024-01-05
feat: add secrets, add cert creation, make optional

JasonPowr:
Pr to allow the operator to either create and use certs, or use certs created by user.

## Testing without cert and secrets
1. Launch and authenticate with cluster
2. Install Keycloak on the cluster
   1. `oc create -k https://github.com/securesign/sigstore-ocp/keycloak/operator/base`
   2. `oc create -k https://github.com/securesign/sigstore-ocp/keycloak/resources/base`
3. Run the operator
   1. `make install && make run` 
4. Add keycloak url to `config/samples/rhtas_v1alpha1_securesign.yaml`
5. Create a namespace: `oc new-project securesign`
6. Apply the yaml file: `oc apply -f config/samples/rhtas_v1alpha1_securesign_cert_create.yaml -n securesign`
7. Sign and verify

## Testing with cert and secrets
1. As Above (1-5)
2. Create secrets in namespace
 ``` 
oc -n securesign create secret generic fulcio-secret-rh --from-file=private=file_ca_key.pem --from-file=public=file_ca_pub.pem --from-file=cert=fulcio-root.pem --from-literal=password=secure  && oc -n securesign create secret generic rekor-private-key --from-file=private=rekor_key.pem
  ```
3. Apply the yaml file: `oc apply -f config/samples/rhtas_v1alpha1_securesign.yaml -n securesign`
4. Sign and verify

Completes https://github.com/securesign/secure-sign-operator/issues/1

PR #19, opened 2024-01-08
Remove scaffolding images

### LINKED ISSUE #3: Migrate Jobs to the operator workload
The operator create various `Job` resources that help to set the deployment up. Most of those jobs are based on [scaffolding](https://github.com/sigstore/scaffolding/tree/main) images and executes simple go app. That GO code could be easily migrated to the operator controllers and dependencies on those images could be removed.

bouskaJ:
Fixes #3 .

As wrote in the README.md file. This PR disable locally running operator because the operator connect directly to the trillian component. The communication is done through RPC. The RPC protocol does rely on HTTP/2 and this traffic is hard (but not impossible - see https://www.redhat.com/en/blog/grpc-or-http/2-ingress-connectivity-in-openshift) to route outside the cluster.

I will open new issue for fixing that.


cooktheryan:
@bouskaJ how does the helm charts handle this today? Also, if we have the `svc` address defined in the configuration do we have the capabilities of bypassing the OpenShift Haproxy?

bouskaJ:
> @bouskaJ how does the helm charts handle this today? Helm does not any active member so there is no capability to run outside the cluster. creates `Job` resources that runs always in the cluster. Also handling the configuration is quite hackish and it was solved by Jobs that are based on images from the https://github.com/securesign/scaffolding project and it brings additional cost of productization and maintenance. Also, it is hard to synchronize them so you can always see a lot of failed pods, restarting and waiting for something.

> Also, if we have the `svc` address defined in the configuration do we have the capabilities of bypassing the OpenShift Haproxy? I am not sure what you mean by this. The `svc` addresses are routed inside the cluster by default. Routing them outside is done by `Route` resource (that works only for `http` traffic and can be [enhanced for the RPC]( https://www.redhat.com/en/blog/grpc-or-http/2-ingress-connectivity-in-openshift)) or using the [nodeport](https://www.tkng.io/services/nodeport/).

I think both solutions (RPC route or nodeport) are possible, it just need some time to dig into it. The benefit of removing the dependency on the `scaffolding` images is worth that IMHO.

bouskaJ:
@lance do you want to comment?

bouskaJ:
I managed to workaround the issue using  https://www.redhat.com/en/blog/grpc-or-http/2-ingress-connectivity-in-openshift. I will make it part of this PR.

cooktheryan:
@sallyom or i can look at the code but @lance going to assign you as the main reviewer based on the above.

cc @lkatalin

cooktheryan:
@sallyom or i can look at the code but @lance going to assign you as the main reviewer based on the above.

cc @lkatalin

lance:
> Also handling the configuration is quite hackish and it was solved by Jobs that are based on images from the https://github.com/securesign/scaffolding project and it brings additional cost of productization and maintenance. Also, it is hard to synchronize them so you can always see a lot of failed pods, restarting and waiting for something.

I agree with this. I have never really liked all of the scaffolding images we are required to produce for helm installation. These do things like initialize the database and initial service configuration. With the exception of `backfill-redis` none of these images are used after a Helm install.

The fact that these same simple processes can be easily incorporated into a Go operator is nice and yes, it does significantly reduce our internal resource utilization for builds, and makes product releases simpler, as we will eliminate around a half dozen images that we need to ship.

controllers/trillian/create.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	db := trillianUtils.CreateTrillDb(instance.Namespace, constants.TrillianDbImage, dbDeploymentName, trillPVC, dbSecret.Name, dbLabels)
+	controllerutil.SetControllerReference(instance, db, i.Client.Scheme())
+	if err = i.Client.Create(ctx, db); err != nil {
+		instance.Status.Phase = rhtasv1alpha1.PhaseError
+		return instance, fmt.Errorf("could not create trillian DB: %w", err)
+	}
+
+	mysql := kubernetes.CreateService(instance.Namespace, "trillian-mysql", 3306, dbLabels)
```
lance, line 84: At some point, this should be configurable, shouldn't it? For example, an EKS deployment may use RDS as the backing database for Trillian. How will this be handled?

controllers/trillian/create.go
```diff
@@ -0,0 +1,144 @@
...
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "rhtas",
+			Namespace:    namespace,
+			Labels:       labels,
+		},
+		Type: "Opaque",
+		Data: map[string][]byte{
+			// generate a random password for the mysql root user and the mysql password
+			// TODO - use a random password generator
```
lance, line 137: admin/letmein123 🤣

controllers/trillian/create.go
```diff
@@ -0,0 +1,144 @@
...
+		if err = i.Client.Create(ctx, pvc); err != nil {
+			instance.Status.Phase = rhtasv1alpha1.PhaseError
+			return instance, fmt.Errorf("could not create pvc: %w", err)
+		}
+		trillPVC = pvc.Name
+	} else {
+		trillPVC = instance.Spec.PvcName
+	}
+
+	db := trillianUtils.CreateTrillDb(instance.Namespace, constants.TrillianDbImage, dbDeploymentName, trillPVC, dbSecret.Name, dbLabels)
```
lance, line 77: I don't see where `CreateTrillDb` creates all of the database resources (tables, indexes, etc.) as is done here https://github.com/securesign/scaffolding/blob/1372537c06f9ba199c42cdbb6d8faa9583b23b81/cmd/trillian/createdb/main.go#L43-L60, for example. Is that still TBD?

config/crd/bases/rhtas.redhat.com_securesigns.yaml
```diff
@@ -95,6 +95,8 @@ spec:
               trillian:
                 description: TrillianSpec defines the desired state of Trillian
                 properties:
+                  external:
+                    type: boolean
```
lance, line 99: I don't see this being used anywhere. What is the purpose of it?

controllers/common/utils/kubernetes/config_map.go
```diff
@@ -0,0 +1,18 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func InitConfigmap(namespace string, name string, labels map[string]string, data map[string]string) *corev1.ConfigMap {
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    labels,
```
lance, line 13: I wonder if utilities like this should always inject a `part-of` label. Or maybe that's what's always passed in... Just wondering if having the `part-of` labels always added at the point of object creation would help with consistency. Take it or leave it. :)

controllers/common/utils/kubernetes/secret.go
```diff
@@ -0,0 +1,17 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func CreateSecret(name string, namespace string, data map[string][]byte, labels map[string]string) *corev1.Secret {
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    labels,
```
lance, line 13: The same comment here as from `CreateConfigMap` regarding labels.

controllers/constants/constants.go
```diff
@@ -0,0 +1,5 @@
+package constants
+
+const (
+	AppName = "trusted-artifact-signer"
```
lance, line 4: 👍

controllers/constants/images.go
lance, line 1: It would be nice to eventually have a github action or something similar that keeps this file up to date with the latest snapshot images from Konflux

controllers/ctlog/create.go
lance, line 1: How will we handle customers who want to bring their own certificates?

lance (commented):
Looking good! I left a few comments inline. Also, I like that we are using the `app.kubernetes.io.part-of` label for the application name. 👍 

It looks like the trillian bits still need work on the initialization phases. I could be missing something, but I did not see where the database tables are created, for example.

I did not check every line of the ct_log initialization and compare with https://github.com/securesign/scaffolding/blob/redhat-v0.6/cmd/ctlog/createctconfig/main.go either. Is that scaffolding image the basis for the init code in, for example controllers/ctlog/create.go?

Overall, I like the pattern.

I think it's important that we get all of the initialization steps done correctly, of course. It might be nice to reference the scaffolding code in comments here, so that it's clear where the behavior is derived from. This also introduces a risk that scaffolding and the operator could diverge - something we need to be on top of.

This does seem to set us back a little bit with regard to BYO certs/keys/db/etc. doesn't it?

controllers/trillian/create.go
bouskaJ: There is already open issue for this https://github.com/securesign/secure-sign-operator/issues/18

controllers/trillian/create.go
bouskaJ: Thank you for spotting this! It seems I missed this job. Strange that it is still working :sweat_smile:. I will fix this tomorrow.

config/crd/bases/rhtas.redhat.com_securesigns.yaml
bouskaJ: It was prepared for bringing the locally running development again (see my latest commit https://github.com/securesign/secure-sign-operator/pull/19/commits/920bcbf0f8c1968c7bb4b0db25401652b9165b91). Now it is used.

controllers/common/utils/kubernetes/config_map.go
bouskaJ: It's always passed in the `labels` map that is parsed by https://github.com/securesign/secure-sign-operator/blob/labels/controllers/common/utils/kubernetes/common.go#L6 but maybe it make sense to have it explicitly defined. :thumbsup: WIll fix it.

controllers/ctlog/create.go
bouskaJ: There is possibility to set the secretName in the CRD (https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/fulcio_types.go#L13) the automatic key generation is optional and it was added in https://github.com/securesign/secure-sign-operator/pull/14 (we will need to tidy the API once it's fully defined to be more understandable)

controllers/trillian/create.go
bouskaJ: I did a little digging into the DB image. It work because the built-in image is already initialized by https://github.com/securesign/trillian/blob/redhat-v1.5/Dockerfile.database#L8. I added comment to https://github.com/securesign/secure-sign-operator/issues/18 that the `create-db` code would be required for customer provided database. I would keep it as part of the https://github.com/securesign/secure-sign-operator/issues/18

bouskaJ:
@lance I think we are good to go. I would left the custom DB support (and the related create-db job), API refactoring and better documentation for follow-up PRs.

lance (approved):
/lgtm

PR #25, opened 2024-01-10
feat: Add service monitors to rekor and fulcio

JasonPowr:
This pr aims to add the service monitors for rekor and fulcio.

# Comments
In order for me to create the service monitor I needed to import `github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1`. This pkg needs an updated version of sigs.k8s.io/controller-runtime to run and in this version MetricsBindAddress and Port where removed https://github.com/securesign/secure-sign-operator/blob/59a3a27241d3157c141fc5846e0741824f1b1884/main.go#L80 Ive commented them out for now but let me know if this is ok?

Source: https://github.com/operator-framework/operator-sdk/issues/6608

## Testing
1. Do usual Setup Cluster, Login, Keycloak, update yaml file with Keycloak URL.
2. Create a ns with the cluster monitoring label: `oc create ns openshift-trusted-artifact-signer && oc label namespace openshift-trusted-artifact-signer openshift.io/cluster-monitoring=true`
3. Apply yaml file with `enableMonitoring: true`.
4. Wait for a bit then go to Observe --> Targets, eventual you should see two targets, `fulcio-metrics` & `rekor-metrics`
5. These metrics can be queried by going to Observe --> Metrics


controllers/rekor/init.go
```diff
@@ -33,6 +35,65 @@ func (i initializeAction) CanHandle(Rekor *rhtasv1alpha1.Rekor) bool {
...
+			return instance, fmt.Errorf("could not create rekor roleBinding: %w", err)
+		}
+
+		serviceMonitor := utils2.CreateServiceMonitor(
+			instance.Namespace,
+			"rekor-metrics",
+			[]monitoringv1.Endpoint{
+				{
+					Interval: monitoringv1.Duration("30s"),
+					Port:     "rekor-server",
```
bouskaJ, line 81: We should avoid using the `hardcoded` names. I know, it's everywhere right now but we should fix it ASAP to avoid other replication. You can fix it or let it for another PR - as you wish.

controllers/common/utils/role.go
```diff
@@ -0,0 +1,16 @@
+package utils
+
+import (
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func CreateRole(namespace, name string, rules []rbacv1.PolicyRule) *rbacv1.Role {
```
bouskaJ, line 8: Can the Role have labels to signal that it is part of our APP?

controllers/common/utils/role_binding.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func CreateRoleBinding(namespace, name string, roleRef rbacv1.RoleRef, subjects []rbacv1.Subject) *rbacv1.RoleBinding {
+	return &rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
```
bouskaJ, line 12: Can the RoleBinding have labels to signal that it is part of our APP?

JasonPowr:
> I left a few comments but otherwise it looks good! 👍 Does the metrics endpoint make sense also for other components?

nb :) ill work on the changes, I don't believe any other components expose a metrics endpoint, ill double check and if they do Ill add because I agree that it would make sense for other components to have it.

JasonPowr:
I've had to bump the version of ci  to 1.21 due to some issues, let me know if this is an acceptable change

bouskaJ:
> I've had to bump the version of ci to 1.21 due to some issues, let me know if this is an acceptable change

I was having troubles with the `1.19` as well so from my perspective the change is welcomed but I know there were some problems with the RH builder image.

@cooktheryan WDYT?

controllers/fulcio/fulcio_controller.go
```diff
@@ -44,6 +44,12 @@ type FulcioReconciler struct {
 //+kubebuilder:rbac:groups=rhtas.redhat.com,resources=fulcios/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=rhtas.redhat.com,resources=fulcios/finalizers,verbs=update
 //+kubebuilder:rbac:groups=rhtas.redhat.com,resources=secrets,verbs=create;get;list;watch;update;patch;delete
+//+kubebuilder:rbac:groups=rhtas.redhat.com,resources=roles,verbs=create;get;list;watch;update;patch;delete
```
bouskaJ, line 47: I don't think that `rhtas.redhat.com` goup has resource of kind `roles` nor `rolebinding`

controllers/rekor/create.go
```diff
@@ -126,6 +132,79 @@ func (i createAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Rekor)
...
+			return instance, fmt.Errorf("could not create rekor roleBinding: %w", err)
+		}
+
+		serviceMonitorLabels := k8sutils.FilterCommonLabels(instance.Labels)
+		serviceMonitorLabels["app.kubernetes.io/component"] = ComponentName
+		serviceMonitorLabels["app.kubernetes.io/name"] = rekorServiceMonitorName
+
+		serviceMonitorMatchLabels := k8sutils.FilterCommonLabels(instance.Labels)
+		serviceMonitorMatchLabels["app.kubernetes.io/component"] = ComponentName
+		serviceMonitorMatchLabels["app.kubernetes.io/name"] = rekorDeploymentName
```
bouskaJ, line 186: Do we need to be specific with `app.kubernetes.io/name`? The https://github.com/securesign/secure-sign-operator/blob/main/controllers/common/utils/kubernetes/common.go#L16 passes the `instance` name so together with the `component` label it should be enough to get the proper deployment. Correct me, I may be wrong....

controllers/rekor/rekor_controller.go
```diff
@@ -46,6 +46,12 @@ type RekorReconciler struct {
 //+kubebuilder:rbac:groups=rhtas.redhat.com,resources=rekors/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=rhtas.redhat.com,resources=rekors/finalizers,verbs=update
 //+kubebuilder:rbac:groups=rhtas.redhat.com,resources=secrets,verbs=create;get;list;watch;update;patch;delete
+//+kubebuilder:rbac:groups=rhtas.redhat.com,resources=roles,verbs=create;get;list;watch;update;patch;delete
```
bouskaJ, line 49: The same as on the fulcio.

controllers/fulcio/fulcio_controller.go
JasonPowr: apologies copy/paste err

bouskaJ (approved):
/LGTM Let's wait for @cooktheryan with the Go version

JasonPowr:
> /LGTM Let's wait for @cooktheryan with the Go version

no, worries :) thats rebased there

cooktheryan:
@JasonPowr lets rebase. I think one concern i would have is if we don't get a RHEL go toolset builder but we have worked around that in the past and the migrated to the newer toolset once available. I am good with moving up though so once you rebase @JasonPowr lets move forward with merging 

LGTM

cooktheryan (approved):

PR #27, opened 2024-01-10
test connection pod

cooktheryan:
test connectivity to created resources


cooktheryan:
@sallyom can I get you to verify the output of the cosign job to ensure its correct?

cooktheryan:
got a verbal LGTM from sally going to write the additional test for create cert and then merge

PR #42, opened 2024-01-16
Add support for ingress

bouskaJ:


cooktheryan:
@bouskaJ nothing pops up as a major concern. Is it possible to include an OpenShift example under samples or within ci?

@sabre1041 @sallyom wanted to get a look here to see how we are handling the differences between OCP and not.

config/crd/bases/rhtas.redhat.com_fulcios.yaml
```diff
@@ -62,9 +62,30 @@ spec:
...
+                properties:
+                  enabled:
+                    description: If set to true, the Operator will create an Ingress
+                      or a Route pointing to Keycloak.
+                    type: boolean
+                  host:
+                    description: If set, the Operator will use value of host for Ingress
+                      host instead of default value keycloak.local. Using this setting
+                      in OpenShift environment will result an error. Only users with
+                      special permissions are allowed to modify the hostname.
```
sallyom, line 76: Is it documented somewhere why we default to `keycloak.local`? Should we default to `keycloak.local`? 

I think the 2nd sentence can be shortened, "Using this setting...." to `If deploying in OpenShift, this setting requires elevated privilege as regular users cannot modify the hostname.`

sabre1041:
@cooktheryan @bouskaJ is there a reason why we do not just focus on leveraging `Ingress` resources exclusively that are managed from an operator perspective. OpenShift will automatically upconvert an `Ingress` resource to a `Route` resource. Setting the annotation `route.openshift.io/termination`: `edge|passthrough|reencrypt` will enable the configuration. This would make it easier from a management perspective as there is only 1 resource to manage from an operator perspective instead of having a bunch of conditional logic based on platform type.

config/crd/bases/rhtas.redhat.com_fulcios.yaml
bouskaJ: @sallyom It's copy-paste error. I will fix that :thumbsup:

cooktheryan:
@bouskaJ any thoughts on converting completely to ingress based on @sabre1041 suggestion. This way we could potentially remove routes making this more friendly to non-openshift installs

cooktheryan:
@lance @sallyom @sabre1041 can I defer to one of y'all to review and push this through. I'm out todu

hack/cosign-testns.yaml
```diff
@@ -29,11 +29,11 @@ spec:
...
-              value: "http://fulcio-server.test.svc"
+              value: "http://fulcio.test.svc"
             - name: OIDC_ISSUER_URL
               value: "http://keycloak-internal.keycloak-system.svc/auth/realms/$(OIDC_AUTHENTICATION_REALM)"
             - name: REKOR_URL
-              value: "http://rekor-server.test.svc"
+              value: "http://rekor.test.svc"
             - name: TUF_URL
               value: "http://tuf.test.svc"
             command: ["/bin/sh", "-c"]
```
sallyom, line 39: we should update these env vars to use the `COSIGN_*` env vars, and then most of the flags in the `cosign sign/verify` commands are unnecessary - we can do this in a follow-up PR, though!

sallyom: (same with the cosign-createcertns.yaml)

sallyom:
amazing work on this operator @cooktheryan @bouskaJ 🤯 !! 
LGTM

sallyom (approved):
LGTM

sabre1041:
Maybe for a followup. But we should support user defined annotations on `externalAccess` configurations. Third party IngressControllers are driven by annotations on Ingress resources. Is there also a desire to specify the TLS termination type? maybe something that we can enable by just providing user defined annotations.

bouskaJ:
> Maybe for a followup. But we should support user defined annotations on `externalAccess` configurations. Third party IngressControllers are driven by annotations on Ingress resources. Is there also a desire to specify the TLS termination type? maybe something that we can enable by just providing user defined annotations.

@sabre1041 Feel free to open new issue. I removed the TLS termination flag because the beckend services are not TLS secured - it does not make sense to set `passthrough` or `reencrypt` termination IMHO but maybe I am wrong.

hack/cosign-testns.yaml
bouskaJ: @sallyom I am working on the go-lang based e2e test suite. My plan is to migrate those test jobs to go code. I can rework it.

PR #45, opened 2024-01-17
Add all options of Fulcio config

osmman:
Structure of Fulcio's config file https://github.com/sigstore/fulcio/blob/main/pkg/config/config.go#L50


cooktheryan:
LGTM

Issue #34, opened 2024-01-12
[SECURESIGN-578] CR Status handling


bouskaJ:
Currently, we use the `phase` field on each CR. In case of error, there is no message on why it happened. 
We should take advantage from using [conditions](https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#manage-cr-status-conditions)

Also status on the `securesign` resource is not implemented yet.

bouskaJ:
@osmman you mentioned that you played with `conditions`do you have any progress? May I assign this issue to you?

osmman:
> @osmman you mentioned that you played with `conditions`do you have any progress? May I assign this issue to you?

Yes I can take it. Currently I am playing with TUF controller and auto discovering of public keys and certs for it. Currently I am putting status of is and one condition per service (Fulcio, Rekor, CTlog, TSA)

PR #46, opened 2024-01-17
feat: Add rekor search ui

JasonPowr:
This pr aims to add the rekor search ui into the operator. Completes: https://github.com/securesign/secure-sign-operator/issues/39

## Testing
1. Usual, cluster, login, build, deploy
2. Create ns `oc create ns trusted-artifact-signer`
3. In your yaml file make sure  `rekor.rekorSearchUI.external = true` to expose the route
4. Apply the yaml file
5. Go to Networking -> Routes -> Rekor-Search-UI
6. When you visit the page you will need to manually input the rekor server url by clicking the gear in the top right
7. You should be able to view info then (sign and verify so there is something there to see ).


JasonPowr:
@bouskaJ, @cooktheryan  I've had to make some adjustments to how ingress is handled for rekor as with this pr there is more than one for the rekor component, let me know if this is all good.

https://github.com/securesign/secure-sign-operator/pull/46/files#diff-05cbca423c4122c2c65b3973f70f2208b5184559cb636d1e2af5f0087fca089c

https://github.com/securesign/secure-sign-operator/pull/46/files#diff-b53b22be10f54f54674230525810c35aaa06fd8fcf677d65084a29c658b2fe75

bouskaJ:
> @bouskaJ, @cooktheryan I've had to make some adjustments to how ingress is handled for rekor as with this pr there is more than one for the rekor component, let me know if this is all good.
> 
> https://github.com/securesign/secure-sign-operator/pull/46/files#diff-05cbca423c4122c2c65b3973f70f2208b5184559cb636d1e2af5f0087fca089c
> 
> https://github.com/securesign/secure-sign-operator/pull/46/files#diff-b53b22be10f54f54674230525810c35aaa06fd8fcf677d65084a29c658b2fe75

The idea behind the previous function was keeping the loosely coupled services and search by labels but it seems it was useless with our current deployment model. Your solution seems reasonable.

api/v1alpha1/rekor_types.go
```diff
@@ -27,6 +29,11 @@ type RekorCert struct {
 	SecretName string `json:"secretName"`
 }
 
+type RekorSearchUI struct {
+	// Define whether you want to export service or not
+	ExternalAccess ExternalAccess `json:"externalAccess,omitempty"`
```
bouskaJ, line 34: `ExternalAccess` does not make sense for me. I would not deploy the UI servise at all if it is not accessible from my browser. Maybe, we can disable/enable the whole `RekorSearch` deployment.

controllers/rekor/init.go
```diff
@@ -106,20 +105,17 @@ func (i waitAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Rekor) (
 
 	if instance.Spec.ExternalAccess.Enabled {
 		protocol := "http://"
-		ingressList := &v12.IngressList{}
-		err = i.Client.List(ctx, ingressList, client2.InNamespace(instance.Namespace), client2.MatchingLabels(labels), client2.Limit(1))
+		ingress := &v12.Ingress{}
```
bouskaJ, line 108: Update also `fulcio` and `tuf` components to follow the same approach.

bouskaJ (changes requested):

api/v1alpha1/rekor_types.go
JasonPowr: ExternalAccess still needs to be passed in so an ingress can be created here, but will set it to true by default

bouskaJ: We can create the UI server **always** with ingress because it does not make sense without it. If we don't want to install the UI server, we can disable it as a whole.

bouskaJ: What about
```
 if instance.Spec.RekorSearchUI.Enabled = true {
// create deployment

 rekorSearchUiIngress, err := k8sutils.CreateIngress(ctx, i.Client, *rekorSearchUiService, ExternalAccess{}, rekorSearchUiDeploymentName, rekorSearchUiLabels) 
```
We can consider to provide `RekorSearchUI.host` field configuration.

bouskaJ: I am thinking even about 
```
//Rekor Search UI
	RekorSearchUI ExternalAccess `json:"rekorSearchUI,omitempty"`
```

It's a little hack but maybe it make sense.

bouskaJ (commented):
Can you add `e2e` test? I think just simple http call is enough to verify that the UI is running.

Otherwise, LGTM. Good job. :+1:

bouskaJ:
rebase

JasonPowr:
rebased :)

api/v1alpha1/rekor_types.go
```diff
@@ -29,11 +31,18 @@ type RekorCert struct {
...
+
 // RekorStatus defines the observed state of Rekor
 type RekorStatus struct {
-	Url    string `json:"url,omitempty"`
-	Phase  Phase  `json:"phase,omitempty"`
-	TreeID *int64 `json:"treeID,omitempty"`
+	Url                string `json:"url,omitempty"`
+	Phase              Phase  `json:"phase,omitempty"`
+	TreeID             *int64 `json:"treeID,omitempty"`
+	RekorSearchUIPhase Phase  `json:"rekorSearchUIPhase,omitempty"`
```
bouskaJ, line 44: I don't think we need another `phase` field. The Serch UI deployment status can be kept in conditions https://github.com/securesign/secure-sign-operator/issues/34 but we can leave it here and refactor later... WDYT?

e2e/byodb_test.go
```diff
@@ -104,6 +108,25 @@ var _ = Describe("Securesign install with byodb", Ordered, func() {
 			tas.VerifyRekor(ctx, cli, namespace.Name, securesign.Name)
 			tas.VerifyCTLog(ctx, cli, namespace.Name, securesign.Name)
 			tas.VerifyTuf(ctx, cli, namespace.Name, securesign.Name)
+			tas.VerifyRekorSearchUI(ctx, cli, namespace.Name, securesign.Name)
+		})
+
+		It("Verify Rekor Search UI is accessible", func() {
```
bouskaJ, line 114: It's enough to verify it in single test workflow. I would keep this one focused on the DB.

api/v1alpha1/rekor_types.go
JasonPowr: Yea that sounds good to me :)

e2e/create_certs_test.go
```diff
@@ -47,6 +48,9 @@ var _ = Describe("Securesign install with certificate generation", Ordered, func
 						Create:     true,
 						SecretName: "rekor-secret",
 					},
+					RekorSearchUI: v1alpha1.RekorSearchUI{
```
bouskaJ, line 51: Maybe we can rename this test workflow as `common_install_test` or something like that and verify some kind of the default installation.

e2e/provided_certs_test.go
```diff
@@ -161,6 +165,25 @@ var _ = Describe("Securesign install with provided certs", Ordered, func() {
 			tas.VerifyCTLog(ctx, cli, namespace.Name, securesign.Name)
 			tas.VerifyTrillian(ctx, cli, namespace.Name, securesign.Name)
 			tas.VerifyTuf(ctx, cli, namespace.Name, securesign.Name)
+			tas.VerifyRekorSearchUI(ctx, cli, namespace.Name, securesign.Name)
+		})
+
+		It("Verify Rekor Search UI is accessible", func() {
```
bouskaJ, line 171: Let's keep this workflow for certificate validation only.

e2e/support/tas/rekor.go
```diff
@@ -44,3 +44,14 @@ func GetRekor(ctx context.Context, cli client.Client, ns string, name string) fu
...
+
+func VerifyRekorSearchUI(ctx context.Context, cli client.Client, namespace string, name string) {
+	Eventually(GetRekor(ctx, cli, namespace, name)).Should(
+		WithTransform(func(f *v1alpha1.Rekor) v1alpha1.Phase {
+			return f.Status.RekorSearchUIPhase
+		}, Equal(v1alpha1.PhaseReady)))
+
+	list := &v1.PodList{}
+	cli.List(ctx, list, client.InNamespace(namespace), client.MatchingLabels{kubernetes.NameLabel: rekor.RekorSearchUiDeploymentName})
+	Expect(list.Items).To(And(Not(BeEmpty()), HaveEach(WithTransform(func(p v1.Pod) v1.PodPhase { return p.Status.Phase }, Equal(v1.PodRunning)))))
```
bouskaJ, line 56: Can you add http call to the search UI to verify that it's responding, please?

JasonPowr: I may be misunderstanding, but I believe thats done here: https://github.com/securesign/secure-sign-operator/blob/ea8b1e7e7f5380f3dc97ba3e6c55bb27febf63df/e2e/create_certs_test.go#L181? 

Shall I move this logic to the VerifyRekorSearchUI function?

bouskaJ (approved):
Great work! Thank you.

PR #48, opened 2024-01-18
Create different trees for CTLog and Rekor services

osmman:
Trillian service should not be responsible for creation and keeping of Merkle tree ID. This responsible should be moved to services which owns tree like CTLog or Rekor.


bouskaJ (approved):

cooktheryan:
LGTM @osmman feel free to merge once you resolve conflict

PR #61, opened 2024-01-19
Rc1 images

cooktheryan:
Release candidate images for TP2


cooktheryan:
@bouskaJ should we merge this as a placeholder before we migrate over to the images from registry.redhat.io they are the release candidate images. Going to file an additional issue to identify the flags required to run the redis image that we create

cooktheryan:
@bouskaJ got the redis flags figured out. This would close #64

PR #71, opened 2024-01-22
pvc reuse if rekor was deleted and redeploy

### LINKED ISSUE #62: rekor PVC stays after deletion
We may want this to occur but the rekor pvc stays when the secure sign resources are deleted

cooktheryan:
Closes #62


cooktheryan:
I'm wondering if I should generate a function here for getting the pvc https://github.com/securesign/secure-sign-operator/blob/main/controllers/common/utils/kubernetes/pvc.go or if this strategy is ok

cooktheryan:
rewrote because my initial commit was not a clean experience

bouskaJ (approved):

bouskaJ:
@cooktheryan rebase please

PR #72, opened 2024-01-22
adding creation of service account to the deployments

Gregory-Pereira:
/cc @cooktheryan 
/cc @sallyom 

This is the service account per component feature.


Gregory-Pereira:
Moving to #76

PR #81, opened 2024-01-24
feat: add client server and Console CLI resources

JasonPowr:
This pr adds the client server and the console CLI resources, to the operator

## Note
One of the CLI binarys is not present in the console cli's download, there is a fix for this in sigstore-ocp, but the image is not produced yet.


cooktheryan:
thanks for including a test. I will keep the issue open until we get that final binary + image

api/v1alpha1/securesign_types.go
```diff
@@ -31,17 +31,27 @@ type SecuresignSpec struct {
...
 	Ctlog CTlogSpec `json:"ctlog,omitempty"`
+	//Enable the ClientServer to serve rekor-cli, gitsign and cosign
+	ClientServer ClientServer `json:"clientServer,omitempty"`
+}
+
+type ClientServer struct {
+	//Enable the ClientServer
+	Enabled bool `json:"enabled,omitempty"`
+	//Enable the OpenshiftCliDownload crd
+	EnableOpenshiftCliDownload bool `json:"enableOpenshiftCliDownload,omitempty"`
```
bouskaJ, line 42: @JasonPowr you can avoid new parameter. ConsoleCliDownload should be automatically created for openshift. Use https://github.com/securesign/secure-sign-operator/blob/main/controllers/common/utils/kubernetes/common.go#L64

controllers/securesign_controller.go
```diff
@@ -318,6 +335,126 @@ func (r *SecuresignReconciler) ensureRekor() func(context.Context, *rhtasv1alpha
 		return false, nil
 	}
 }
+
+func (r *SecuresignReconciler) ensureClientServer() func(context.Context, *rhtasv1alpha1.Securesign) (bool, error) {
```
bouskaJ, line 339: I don't think, this should be bind to `securesign` resource creation. The client server should be spin up automatically once the operator is deployed. Please move it to the `main.go`.

PR #85, opened 2024-01-24
adding sa per component

Gregory-Pereira:
/cc @cooktheryan

### DIFF
--- a/controllers/ctlog/utils/ctlog_deployment.go
+++ b/controllers/ctlog/utils/ctlog_deployment.go
@@ -9,9 +9,6 @@ import (
 )
 
 func CreateDeployment(namespace string, deploymentName string, configName string, labels map[string]string, serviceAccountName string) *appsv1.Deployment {
-	if serviceAccountName == "" {
-		serviceAccountName = constants.ServiceAccountName
-	}
 	replicas := int32(1)
 	// Define a new Deployment object
 	return &appsv1.Deployment{
--- a/controllers/ctlog/utils/ctlog_deployment.go
+++ b/controllers/ctlog/utils/ctlog_deployment.go
@@ -8,7 +8,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-func CreateDeployment(namespace string, deploymentName string, configName string, labels map[string]string) *appsv1.Deployment {
+func CreateDeployment(namespace string, deploymentName string, configName string, labels map[string]string, serviceAccountName string) *appsv1.Deployment {
+	if serviceAccountName == "" {
+		serviceAccountName = constants.ServiceAccountName
+	}
 	replicas := int32(1)
 	// Define a new Deployment object
 	return &appsv1.Deployment{
@@ -27,7 +30,7 @@ func CreateDeployment(namespace string, deploymentName string, configName string
 					Labels: labels,
 				},
 				Spec: corev1.PodSpec{
-					ServiceAccountName: constants.ServiceAccountName,
+					ServiceAccountName: serviceAccountName,
 					Containers: []corev1.Container{
 						{
 							Name:  "ctlog",
--- a/controllers/rekor/create.go
+++ b/controllers/rekor/create.go
@@ -3,11 +3,13 @@ package rekor
 import (
 	"context"
 	"fmt"
+
 	"github.com/securesign/operator/controllers/common"
 
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	rhtasv1alpha1 "github.com/securesign/operator/api/v1alpha1"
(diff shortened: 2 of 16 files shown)


controllers/securesign_controller.go
```diff
@@ -53,6 +53,8 @@ const (
 
 	gitsignConsoleCliName        = "gitsign"
 	gitsignConsoleCliDescription = "gitsign is a CLI tool that allows you to digitally sign and verify git commits."
+
+	OperatorServiceAccountName = "rhtas-operator-controller-manager"
```
cooktheryan, line 57: I think we can remove these

cooktheryan: Any of the service account, role and binding in this specific file creation for the manager is done by the deployment

controllers/securesign_controller.go
```diff
@@ -154,86 +151,6 @@ func (r *SecuresignReconciler) ensureSecureSign(
...
-				if err != nil {
-					return false, err
-				}
-			}
-			if err != nil {
-				return false, err
-			}
-		}
-
-		role := kubernetes.CreateRole(securesign.Namespace, constants.ServiceAccountName, labels(*securesign),
```
cooktheryan, line 185: we will need to create two new common/utils/kubernetes files one for role and one for binding. 

The first role will need to be assigned to fulcio tuf and rekor
https://github.com/sigstore/scaffolding/blob/main/config/fulcio/certs/101-binding.yaml
https://github.com/sigstore/scaffolding/blob/main/config/rekor/createsecret/101-binding.yaml
 https://github.com/sigstore/scaffolding/blob/main/config/tuf/createsecret/101-binding.yaml and have the secret resources access

The next will need to be for ctlog
https://github.com/sigstore/scaffolding/blob/main/config/ctlog/certs/101-binding.yaml which looks to have both configmap and secrets

cooktheryan: These actually dont need to be new files but we should follow the same procedures done by the monitoring enablement. For example, https://github.com/securesign/secure-sign-operator/blob/main/controllers/rekor/create.go#L157

config/manager/kustomization.yaml
```diff
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/grpereir/operator
-  newTag: 1.0.1
+  newTag: testing
```
bouskaJ, line 8: Can you avoid replacing the image and image tag? We update it almost in every PR. :-/

controllers/ctlog/utils/ctlog_deployment.go
```diff
@@ -8,7 +8,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-func CreateDeployment(namespace string, deploymentName string, configName string, labels map[string]string) *appsv1.Deployment {
+func CreateDeployment(namespace string, deploymentName string, configName string, labels map[string]string, serviceAccountName string) *appsv1.Deployment {
+	if serviceAccountName == "" {
+		serviceAccountName = constants.ServiceAccountName
```
bouskaJ, line 13: I think we can remove the shared account ( and also account creation from the `securesign_controller`)

controllers/rekor/create.go
```diff
@@ -24,6 +26,7 @@ const (
 	ComponentName               = "rekor"
 	rekorMonitoringRoleName     = "prometheus-k8s-rekor"
 	rekorServiceMonitorName     = "rekor-metrics"
+	RekorServiceAccountName     = "rekor-sa"
```
bouskaJ, line 29: I don't think we need to export `RekorServiceAccountName` -> change to lowercase `rekorServiceAccountName`

controllers/trillian/create.go
```diff
@@ -21,7 +22,10 @@ const (
 	logserverDeploymentName = "trillian-logserver"
 	logsignerDeploymentName = "trillian-logsigner"
 
-	ComponentName = "trillian"
+	dbServiceAccountName        = "trillian-db-sa"
+	logsignerServiceAccountName = "trillian-logsigner-sa"
```
bouskaJ, line 26: I think we can share one SA for all deployments.

controllers/rekor/create.go
Gregory-Pereira: Ill swap it back but I was under the assumption that the capital letters implied that they were constants

bouskaJ:
```
In Go, a name is exported if it begins with a capital letter. For example, Pizza is an exported name, as is Pi, which is exported from the math package.

pizza and pi do not start with a capital letter, so they are not exported.
```
See https://go.dev/tour/basics/3

bouskaJ (commented):
Few more comments but otherwise looks good :+1:

bouskaJ:
resolved by https://github.com/securesign/secure-sign-operator/pull/95

PR #90, opened 2024-01-25
Rbac error currently when following the operator log there are permissions issues

cooktheryan:
Err `cannot create resource "events" in API group "" in the namespace`


Gregory-Pereira (approved):
/lgtm
Great catch thanks

Issue #54, opened 2024-01-19
Create ServiceAccount for every service


osmman:
We should create service account for every TAS service to be deployable independently and enable to solve https://github.com/securesign/secure-sign-operator/issues/38 issue.

PR #95, opened 2024-01-26
Operator 2.0

bouskaJ:
**Operator controllers refactoring**
-  new approach to the controller's reconciliation loop. With one loop rotation, only single action is threat. This concept follows operator best practise (see https://www.redhat.com/en/blog/7-best-practices-for-writing-kubernetes-operators-an-sre-perspective), allow us to re-enable client caching again and should bring more stability.
- better readability
- controller unit tests
- closes https://github.com/securesign/secure-sign-operator/issues/54
- deployed stack is automatically restored (example: the `service` resource is redeployed after removal)
- progress on https://github.com/securesign/secure-sign-operator/issues/34 (but it needs to be finished)


bouskaJ:
@osmman WDYT?

controllers/tuf/init.go
```diff
@@ -4,63 +4,48 @@ import (
...
+	actions := []func(context.Context, *rhtasv1alpha1.Tuf) (bool, error){
+		i.deploymentRunning(),
+		i.updateURL(),
 	}
 
-	if instance.Spec.ExternalAccess.Enabled {
-		protocol := "http://"
-		ingress := &v12.Ingress{}
-		err = i.Client.Get(ctx, types.NamespacedName{Name: ComponentName, Namespace: instance.Namespace}, ingress)
+	for _, a := range actions {
```
osmman, line 40: I am thinking to extract these to separate action.Action implementations. For me it looks realy similiar to main reconcile loop. Example of sequences of actions.
```
	actions := []action.Action[rhtasv1alpha1.Tuf]{
		ChangePhase(PhaseNone, PhasePending),
		ResolvingSecrets(),
		ChangePhase(PhasePending, PhaseCreating),
		ReconcileSA(),
		ReconcileDeployment(),
		ReconcileService()
		ReconcileIngress(),
		ChangePhase(PhaseCreating, PhaseInitialize),
		DeploymentRunning(),
		UpdateURL(),
		ChangePhase(PhaseInitialize, PhaseReady),
	}
```

controllers/tuf/tuf_controller.go
```diff
@@ -121,22 +121,20 @@ func (r *TufReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 
 		if a.CanHandle(target) {
 			newTarget, err := a.Handle(ctx, target)
-			if err != nil {
-				if newTarget != nil {
-					_ = r.Status().Update(ctx, newTarget)
+			if newTarget != nil {
+				if equality.Semantic.DeepDerivative(target.Spec, newTarget.Spec) {
+					_ = r.Update(ctx, target.DeepCopy())
```
osmman, line 126: we should not ignore errors, minimaly we should return reconcilation results as failure

controllers/tuf/tuf_controller.go
```diff
@@ -121,22 +121,20 @@ func (r *TufReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 
 		if a.CanHandle(target) {
 			newTarget, err := a.Handle(ctx, target)
-			if err != nil {
-				if newTarget != nil {
-					_ = r.Status().Update(ctx, newTarget)
+			if newTarget != nil {
```
osmman, line 124: I think how to make for Action better interface to control reconcilation result. It is not exactly clear from Handle function how it will impact reconcile loop. Currently handle function can:

- run next action
- stop loop and return reconcile.Result
   - failure
   - update object
   - update status
   - requeue reconcile request

```

type Continue func() R
type Update func(T) R
type Failed func(error) R
type Requeue func(time.Duration) R

func (c tuf) Handle(ctx context.Context, instance *v1alpha1.Tuf) R {
if err != nil {
   retrun Failure(err)
}

if requeue {
   // waiting for some resource to be awalaible
   if !not {
      return Requeue(30 * time.Seconds)
   }
}

if change {
   instance.Spec.A = "B"
   return Update(instance)
}

return Continue()
} 
```

bouskaJ:
I can share my messy configuration with you if you want to execute the test.

osmman:
> @osmman I have a new purposal. Please review and comment, before I will work on other component. I needed to hack all other components in order to make it executable but the `tuf_controller_test` already passed.

it's looking great

controllers/tuf/pending/pending.go
```diff
@@ -39,49 +40,48 @@ func (i pendingAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Tuf)
...
+			updated, err := i.handleKey(ctx, instance, &instance.Spec.Keys[index])
+			if err != nil {
+				meta.SetStatusCondition(&instance.Status.Conditions, v1.Condition{
+					Type:    key.Name,
+					Status:  v1.ConditionFalse,
+					Reason:  "Failure",
+					Message: err.Error(),
+				})
+				// lets retry
+				return i.FailedWithStatusUpdate(ctx, err, instance)
```
osmman, line 66: originaly I used Requeue action with some delay because there is some limit how many time failed event can be retried untill it will be ignored (~5 times). This happens because it waits for other services to create their secrets and it took some time.

osmman: It is possible that current solution works because of meta.SetStatus... is called every time.

bouskaJ:
Ready! @cooktheryan @osmman I was able to pass all tests but please double check. It's hard to keep eye on everything....

cooktheryan:
everything looks great. The only question I have is would it be less resources to do a get rather than a list for the deployments and secrets. If it's not much different then that's fine. LGTM 

@osmman can you take a look as well since there is a lot changing

api/v1alpha1/rekor_types.go
```diff
@@ -16,7 +16,7 @@ type RekorSpec struct {
 	// Persistent volume claim name to bound with Rekor component
 	PvcName string `json:"pvcName,omitempty"`
 	//Enable Service monitors for rekor
-	Monitoring bool `json:"monitoring,omitempty"`
+	Monitoring MonitoringConfig `json:"monitoring,omitempty"`
```
osmman, line 19: [samples](https://github.com/securesign/secure-sign-operator/blob/e74c1ab29556696edc5809b0f71a77d6cf890ddf/config/samples/rhtas_v1alpha1_securesign.yaml#L13) will have to be modified to this change

cooktheryan:
I don't think my question is worth worrying about. With the labels it should be low resource usage

PR #105, opened 2024-01-29
Fix IT test after the ec removal

bouskaJ:
Fix IT test after the `ec` cli removal


bouskaJ:
@Gregory-Pereira @cooktheryan PTAL

bouskaJ:
The CI is failing with `Back-off pulling image` which is expected

cooktheryan (approved):

PR #112, opened 2024-01-31
trillian-db

cooktheryan:
Image needs to be tested by rebooting nodes after DB has been running. If passes we should bring the registry.redhat.io image into the operator


cooktheryan:
the updated image passes the reboot test

bouskaJ (approved):

PR #137, opened 2024-02-02
replicating fbc-v4.13 into other overlays

Gregory-Pereira:
/cc @lance @tommyd450 @cooktheryan


Gregory-Pereira:
Changes include:
- Updating versions of component images and swapping to what they will become at `registry.redhat.io`.
- creating FBC overlays for all OCP catalogs 4.11 - 4.16.
- removing `.tekton` files for the FBC component as I will be re-creating them after this merges in keeping with how `cnv` did theirs so we can have a reference and keep things consistent for the `konflux` folks

lance (approved):

PR #159, opened 2024-02-05
Db configuration

cooktheryan:
Allow for pvc to be retained or deleted when defining the deployment. Also allow for PVC size to be defined. I had to define the if statement in the way I did in case a user provided no input for the field. Open to suggestions on doing it cleaner though.

closes #112

### DIFF
--- a/api/v1alpha1/rekor_types.go
+++ b/api/v1alpha1/rekor_types.go
@@ -21,6 +21,10 @@ type RekorSpec struct {
 	RekorSearchUI RekorSearchUI `json:"rekorSearchUI,omitempty"`
 	// Signer configuration
 	Signer RekorSigner `json:"signer,omitempty"`
+	// Retain the PVC after Rekor is deleted
+	RetainPVC bool `json:"retainPVC,omitempty"`
+	// Rekor PVC Size in MB or GB. Example: 5Gi
+	PvcSize string `json:"pvcSize,omitempty"`
 }
 
 type RekorSigner struct {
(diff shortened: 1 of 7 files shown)


api/v1alpha1/rekor_types.go
```diff
@@ -21,6 +21,10 @@ type RekorSpec struct {
 	RekorSearchUI RekorSearchUI `json:"rekorSearchUI,omitempty"`
 	// Signer configuration
 	Signer RekorSigner `json:"signer,omitempty"`
+	// Retain the PVC after Rekor is deleted
```
bouskaJ, line 24: I would create structure like this:
```
persistentVolume:
  size:
  retian:
  name:

File: controllers/rekor/create.go [hunk 2]
```diff
@@ -76,12 +77,18 @@ func (i createAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Rekor)
 		} else {
 			// PVC does not exist, create a new one
 			i.Logger.V(1).Info("Creating new PVC")
-			rekorPvc := k8sutils.CreatePVC(instance.Namespace, "rekor-server", "5Gi")
-			if err := i.Client.Create(ctx, rekorPvc); err != nil {
+			if instance.Spec.PvcSize == "" {
+				instance.Spec.PvcSize = "5Gi"
```
bouskaJ, line 81: We can set default value in the API struct (see https://book.kubebuilder.io/reference/markers/crd-validation)

cooktheryan:
Going to hold this until the 2.0 change merges

PR #168, opened 2024-02-09
Fix keys autodiscovery

bouskaJ:


controllers/ctlog/actions/servrConfig.go
osmman, line 1: typo in file name `servr`

controllers/rekor/actions/server/resolve_pub_key.go
```diff
@@ -55,18 +53,19 @@ func (i resolvePubKeyAction) Handle(ctx context.Context, instance *rhtasv1alpha1
...
+	scr := k8sutils.CreateSecret(secretName, instance.Namespace,
 		map[string][]byte{
-			"public": key,
+			keyName: key,
 		}, labels)
-	if err = controllerutil.SetControllerReference(instance, secret, i.Client.Scheme()); err != nil {
+	if err = controllerutil.SetControllerReference(instance, scr, i.Client.Scheme()); err != nil {
 		return i.Failed(fmt.Errorf("could not set controller reference for Secret: %w", err))
 	}
-	if _, err = i.Ensure(ctx, secret); err != nil {
```
osmman, line 69: Rise event of created secret with public key

e2e/provided_certs_test.go
```diff
@@ -113,35 +117,6 @@ var _ = Describe("Securesign install with provided certs", Ordered, func() {
 					ExternalAccess: v1alpha1.ExternalAccess{
 						Enabled: true,
 					},
-					Keys: []v1alpha1.TufKey{
```
osmman, line 116: We should keep this use case, I think it is more connected to original scenario which is testing custom user certificates. It will be better to create a new scenario to test TUF's autodiscovery.

PR #170, opened 2024-02-09
Defining fbc ReleasePlans

tommyd450:
Defining the yaml we use for the ReleasePlans.
THe will be applied on the Konflux cluster after merging.


lance:
These are now being tracked in securesign/releases

PR #173, opened 2024-02-12
feat: Add Back fill Redis CronJob

JasonPowr:
This Pr aims to add the Back fill Redis CronJob to the stack.
Target: GA


JasonPowr:
> > @bouskaJ @cooktheryan @osmman
> > Looking for a bit of guidance on this one, the way the backfill redis job works is it takes a start index and an end index, loops through this and patches redis where it needs to. I'm wondering if there's a better way to do this then how I have it setup in this pr (why it's a draft) At the moment I have a init container to curl the Rekor API, get the current tree size -1 (index), write to a shared volume, then its read in in the main container, this works but its not going to scale well if it's needed.
> > Should the start and end index, be read in through the API, might be a better way of handling it? Although a bit more awkward for the end user
> > Any better way of handling it?
> 
> I am just wondering why you do the CURL in the separate container? It is not possible to do it in one container?

For jq, the image uses in the init container has both curl and jq

controllers/rekor/actions/backfillRedis/backfill_redis_cronjob.go
```diff
@@ -0,0 +1,132 @@
...
+									VolumeSource: corev1.VolumeSource{
+										EmptyDir: &corev1.EmptyDirVolumeSource{},
+									},
+								},
+							},
+							InitContainers: []corev1.Container{
+								{
+									Name:    "get-current-tree-size",
+									Image:   constants.BackfillRedisInitImage,
+									Command: []string{"sh", "-c", fmt.Sprintf("curl -sS http://%s/api/v1/log | jq '.treeSize -1' > %s", rekorServiceName, treeSizeFilePath)},
```
lance, line 91: Agree this will not be efficient as the tlog grows.

controllers/rekor/actions/backfillRedis/backfill_redis_cronjob.go
```diff
@@ -0,0 +1,132 @@
...
+										},
+									},
+								},
+							},
+							Containers: []corev1.Container{
+								{
+									Name:  name,
+									Image: constants.BackfillRedisImage,
+									Command: []string{
+										"sh", "-c", fmt.Sprintf("endIndex=$(cat %s) && exec backfill-redis --hostname=rekor-redis --port=6379 --rekor-address=http://%s --start=0 --end=$endIndex", treeSizeFilePath, rekorServiceName),
```
lance, line 116: There is a bit of a race condition here isn't there? I guess it's not necessarily a bad one, but the size of the log may change betwen the init container's work and here.

lance (commented):
Looking at the Rekor API, there is no way to get a count of entries directly from the API. This is a good upstream contribution we might consider.

controllers/rekor/actions/backfillRedis/backfill_redis_cronjob.go
JasonPowr: Didn't even think of that, good catch although jans comment about removing the init containers and using `sed` to get the treeSize instead of jq, should reduce that chance a good bit.

controllers/constants/images.go
```diff
@@ -10,9 +10,11 @@ const (
...
-	RekorServerImage   = "quay.io/redhat-user-workloads/rhtas-tenant/rekor-1-0-gamma/rekor-server-1-0-gamma@sha256:53b650ad487dce78025d1dbddc5f25116c132f4e78b7d6f8c1dd0638574f6db3"
-	RekorSearchUiImage = "quay.io/redhat-user-workloads/rhtas-tenant/rekor-1-0-gamma/rekor-search-ui-1-0-gamma@sha256:ea4344bc762809ca46ea0708de378d8592b97194a9c1e08eb9396294276818bf"
+	RekorRedisImage        = "quay.io/redhat-user-workloads/rhtas-tenant/trillian-1-0-gamma/redis-0-6@sha256:acf920baf6ee1715c0c9d7ddf69867d331c589d3afa680048c508943078d9585"
+	RekorServerImage       = "quay.io/redhat-user-workloads/rhtas-tenant/rekor-1-0-gamma/rekor-server-1-0-gamma@sha256:53b650ad487dce78025d1dbddc5f25116c132f4e78b7d6f8c1dd0638574f6db3"
+	RekorSearchUiImage     = "quay.io/redhat-user-workloads/rhtas-tenant/rekor-1-0-gamma/rekor-search-ui-1-0-gamma@sha256:ea4344bc762809ca46ea0708de378d8592b97194a9c1e08eb9396294276818bf"
+	BackfillRedisInitImage = "registry.redhat.io/openshift4/ose-tools-rhel8@sha256:ec47625c842cb1d042157fe32193f28fbb9ddc710eecf16c93d8fa228faa4c13"
```
bouskaJ, line 16: @JasonPowr you can remove this image

controllers/rekor/actions/backfillRedis/backfill_redis_cronjob.go
```diff
@@ -0,0 +1,92 @@
...
+
+func CreateBackfillRedisCronjob(namespace, name, rekorServiceName string, labels map[string]string) *batchv1.CronJob {
+	return &batchv1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    labels,
+		},
+		Spec: batchv1.CronJobSpec{
+			Schedule: "0 0 * * *",
```
bouskaJ, line 69: we can make this `schedule` part of the configuration WDYT?

controllers/constants/images.go
JasonPowr: good point, forgot about that, apologies

controllers/rekor/actions/backfillRedis/backfill_redis_cronjob.go
JasonPowr: yea that makes sense to me

api/v1alpha1/rekor_types.go
```diff
@@ -41,6 +43,13 @@ type RekorSearchUI struct {
 	Enabled bool `json:"enabled,omitempty"`
 }
 
+type BackFillRedis struct {
+	//Enable the BackFillRedis CronJob
+	Enabled bool `json:"enabled,omitempty"`
+	//Schedule for the BackFillRedis CronJob
+	Schedule string `json:"schedule,omitempty"`
```
bouskaJ, line 50: Set the default value https://book.kubebuilder.io/reference/markers/crd-validation

bouskaJ (approved):
One comment but otherwise it looks good. The curl API request is not the best solution but I was not able to find a better one...

api/v1alpha1/rekor_types.go
osmman: and we should add a validation restriction to allow only cron format.

Issue #171, opened 2024-02-09
Client server resources left uninstalled after operator deletion


bouskaJ:
Operator is not able to uninstall Client server resources on deletion
```
rhtas-operator-controller-manager-68f597c65f-ntm78 manager 2024-02-09T15:05:48Z DEBUG   events  rhtas-operator-controller-manager-68f597c65f-ntm78_ba7537bc-de69-4abd-8e45-b54bf23ed1e7 stopped leading {"type": "Normal", "object": {"kind":"Lease","namespace":"openshift-operators","name":"f81d37df.redhat.com","uid":"ca73f1bb-d05b-4563-b8de-6928c376324b","apiVersion":"coordination.k8s.io/v1","resourceVersion":"121349706"}, "reason": "LeaderElection"}
rhtas-operator-controller-manager-68f597c65f-ntm78 manager 2024-02-09T15:05:48Z ERROR   error received after stop sequence was engaged  {"error": "leader election lost"}
rhtas-operator-controller-manager-68f597c65f-ntm78 manager sigs.k8s.io/controller-runtime/pkg/manager.(*controllerManager).engageStopProcedure.func1
rhtas-operator-controller-manager-68f597c65f-ntm78 manager      /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/manager/internal.go:490
rhtas-operator-controller-manager-68f597c65f-ntm78 manager 2024-02-09T15:05:48Z ERROR   setup   unable to cleanup client-server resources       {"error": "deployments.apps \"cli-server\" is forbidden: User \"system:serviceaccount:openshift-operators:rhtas-operator-controller-manager\" cannot delete resource \"deployments\" in API group \"apps\" in the namespace \"trusted-artifact-signer\""}
rhtas-operator-controller-manager-68f597c65f-ntm78 manager main.main.func1
rhtas-operator-controller-manager-68f597c65f-ntm78 manager      /workspace/main.go:191
rhtas-operator-controller-manager-68f597c65f-ntm78 manager main.main
rhtas-operator-controller-manager-68f597c65f-ntm78 manager      /workspace/main.go:202
rhtas-operator-controller-manager-68f597c65f-ntm78 manager runtime.main
rhtas-operator-controller-manager-68f597c65f-ntm78 manager      /usr/local/go/src/runtime/proc.go:267
[...]

PR #174, opened 2024-02-12
Fix 171

bouskaJ:
Fixes https://github.com/securesign/secure-sign-operator/issues/171.

**Problem:**
Operator uninstall means delete of all created resources - even the `Role`. The `Role` does not exists at the operator cleanup time so resources can't be deleted.

**possible solutions**
1) move the client-server to the operator-namespace and bound all resources to the operator deployment
  :-1:  bring mess to the `openshift-operators` namespace
2) bound all CLI server resources to some cluster-wide resource
  :-1:  resource cleanup is not very clear, cluster wide resources stay present in the cluster even after the operator removal (so as the CLI server)
3) Manage CLI server by securesign controller
  :-1:  :-1:  I think this is not suitable for us - the cli-server should arise with the operator installation and disappear on the operator deletion, not with the `securesign` resource

I choose the solution nr. 2 the CLI server is bound to the `securesign` CRD.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,6 +10,7 @@ on:
 env:
   GO_VERSION: 1.21
   IMAGE_TAG_BASE: ${KIND_REGISTRY}/securesign
+  IMAGE_TAG: latest
 
 jobs:
   build-operator:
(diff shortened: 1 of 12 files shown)


bouskaJ:
@cooktheryan @osmman WDYT?

cooktheryan:
I'm good with this approach. @osmman any thoughts?

PR #184, opened 2024-02-14
Chore: timestamp --> trigger rebuild

Gregory-Pereira:
good old foobar


tommyd450:
Using this to test the EC failures.

tommyd450:
Have a sneaking suspicion what it may be.

tommyd450:
So the last promotion of this and the actual operator image was on Friday, sometimes during updates in Konflux pipelines it requires us to mark ec as optional so that it can update the references to the tasks in pipeline if that makes sense? Marking as optional and attempting to see if images are promoted.

Issue #79, opened 2024-01-24
[SECURESIGN-570] Rekor add options to set trillian_log_server 


osmman:
We should add options to configure external trillian log server. Currently deployed Rekor instance is configured to use only Trillain instance in same namespace.

It should help https://github.com/securesign/secure-sign-operator/issues/38 issue.

tommyd450:
Still figuring out operators so I will give this one a try. Unless someone is already working on it.

bouskaJ:
Migrated to Jira

PR #196, opened 2024-02-20
[SECURESIGN-570]  Rekor add options to set trillian_log_server

tommyd450:
#79 
First actual pass at a pr for this, creating the yaml with the "externalTrillian" field added as well as providing the treeID  of that trillian instance worked seemingly though I would love for someone else to take a brief look.

My process for testing this was installing the helm chart and pointing the the operators rekor-server at the trillian-system namespace and providing its TreeID.


controllers/rekor/utils/rekor_deployment.go
```diff
@@ -12,9 +12,16 @@ import (
...
+	trillianNamespace := instance.Namespace
+
+	if instance.Spec.ExternalTrillian != "" {
+		trillianNamespace = instance.Spec.ExternalTrillian
+	}
+
 	appArgs := []string{
 		"serve",
-		"--trillian_log_server.address=trillian-logserver." + instance.Namespace + ".svc",
+		"--trillian_log_server.address=trillian-logserver." + trillianNamespace + ".svc",
```
osmman, line 24: make --trillain_log_server.address fully configurable, do not limit it only for kubernetes internal network

osmman (commented):
I think it will be better to change the API to something like this:
```
spec:
   trillian:
      address: trillian.example.com
      port: 8091
```

current state has problem that Trillain instance has to be installed in same cluster as Rekor which is unnecessary limitation. It will be much more flexibel for example: 
- set ip address instead of dns record
- usage of API Gateway
- external DNS name

You can set address and port default values to make default deployment simple and configure these options if changes are required.
Default values:
- address = trillian-logserver.svc
- port = 8091

controllers/rekor/utils/rekor_deployment.go
tommyd450: Awesome thanks for the feedback.

osmman:
I think that you can skip helm chart for testing it and use only Operator. Trillian and Rekor should be possible to install separately via Operator CRD. 

Trillain in trillian-system namespace
```
kind: Trillian
metadata:
  name: example
  namespace: trillian-system
spec:
  database:
    create: true
```

Rekor in rekor-system namespace
```
kind: Rekor
metadata:
  name: example
  namespace: rekor-system
spec:
  trillain:
    address: trillain-server.trillain-system.svc
```

api/v1alpha1/rekor_types.go
```diff
@@ -25,6 +25,10 @@ type RekorSpec struct {
 	Pvc Pvc `json:"pvc,omitempty"`
 	// BackFillRedis CronJob Configuration
 	BackFillRedis BackFillRedis `json:"backFillRedis,omitempty"`
+	// Trillian External Log Server
+	ExternalTrillianUrl string `json:"externalTrillian,omitempty"`
```
osmman, line 29: I think that is better to structure API into hierarchical structure as I provided in this comment https://github.com/securesign/secure-sign-operator/pull/196#pullrequestreview-1890294188. I have these reasons:

- keep consistency across Operator APIs (like pvc,signer, backfillredis, ...) 
- `externalTrillian` do not fit well. `external` does not describe what its purpose is and can confuse user. Host or address should fit much better

api/v1alpha1/rekor_types.go
```diff
@@ -25,6 +25,10 @@ type RekorSpec struct {
 	Pvc Pvc `json:"pvc,omitempty"`
 	// BackFillRedis CronJob Configuration
 	BackFillRedis BackFillRedis `json:"backFillRedis,omitempty"`
+	// Trillian External Log Server
```
osmman, line 28: I would use the description from the log server [flags](https://github.com/sigstore/rekor/blob/main/cmd/rekor-server/app/root.go#L85-L86).
`Trillian log server address`

controllers/rekor/utils/rekor_deployment.go
```diff
@@ -12,10 +12,21 @@ import (
 
 func CreateRekorDeployment(instance *v1alpha1.Rekor, dpName string, sa string, labels map[string]string) *apps.Deployment {
 	env := make([]core.EnvVar, 0)
+
```
osmman, line 15: It is better to define default values into API. That way values should be predefined by API server and information about default values will be generated in OpenAPI. 

Currently we are using `//+kubebuilder:defaults:` annotation for it in the project

config/samples/rhtas_v1alpha1_securesign.yaml
osmman, line 1: These changes should be reverted from commit.

api/v1alpha1/rekor_types.go
tommyd450: Understood.

api/v1alpha1/rekor_types.go
tommyd450: Understood will make that change too.

controllers/rekor/utils/rekor_deployment.go
tommyd450: Ok I understand, I will make those changes.

api/v1alpha1/rekor_types.go
```diff
@@ -25,8 +25,12 @@ type RekorSpec struct {
 	Pvc Pvc `json:"pvc,omitempty"`
 	// BackFillRedis CronJob Configuration
 	BackFillRedis BackFillRedis `json:"backFillRedis,omitempty"`
-	// Trillian External Log Server
-	ExternalTrillian string `json:"externalTrillian,omitempty"`
+	// Trillian Log Server End point
+	//+kubebuilder:default:="trillian-logserver.trillian-system.svc"
```
osmman, line 29: I think that `trillian-logserver`will be better as default value. It should work with our default deployment of Securesign resource.

Problem is in `trillian-system.svc` which targets to specific namespace which we unable to predict.

api/v1alpha1/rekor_types.go
```diff
@@ -25,6 +25,12 @@ type RekorSpec struct {
 	Pvc Pvc `json:"pvc,omitempty"`
 	// BackFillRedis CronJob Configuration
 	BackFillRedis BackFillRedis `json:"backFillRedis,omitempty"`
+	// Trillian Log Server End point
+	//+kubebuilder:default:="trillian-logserver.trillian-system.svc"
+	TrillianAddress string `json:"trillianAddress,omitempty"`
+	// Trillian Log Server End point
+	//+kubebuilder:default:="8091"
+	TrillianPort string `json:"trillianPort,omitempty"`
```
osmman, line 33: port should be of integer typ, ideally it should be uint16.

notes.txt
osmman, line 1: I think that it will be better to move it into README file or tranform it into bash file and store it in hack folder

api/v1alpha1/rekor_types.go
osmman: It is possible that uint16 will not becompatible with OpenApi so other option is to use int32 and add validation rule for min max values.
```
//+kubebuilder:validation:Minimum:=1
//+kubebuilder:validation:Maximum:= 65535
//+kubebuilder:default:="8091"
TrillianPort int32 `json:"trillianPort,omitempty"`
```

controllers/rekor/actions/server/createTree.go
```diff
@@ -38,12 +39,20 @@ func (i createTrillianTreeAction) Handle(ctx context.Context, instance *rhtasv1a
...
 	var err error
+	var tree *trillian.Tree
+	var trillUrl string
 
-	trillUrl, err := k8sutils.GetInternalUrl(ctx, i.Client, instance.Namespace, trillian.LogserverDeploymentName)
-	if err != nil {
-		return i.Failed(err)
+	switch {
+	case instance.Spec.Trillian.Port == nil:
+		err = fmt.Errorf("%s: %w", i.Name(), utils.TrillianPortNotSpecified)
```
bouskaJ, line 47: You should throw the error. It could be swallowed by `tree, err = common.CreateTrillianTree(ctx, "rekor-tree", trillUrl, constants.CreateTreeDeadline)`.

osmman: this check is required for building of `trillUrl`. It will fails on fatal error before it will be passed to CreateTrillianTree function if it is nil. The reason is that `Port` has to be dereferenced and uset to crreate `trillUrl`

controllers/rekor/actions/server/deployment.go
```diff
@@ -38,7 +39,14 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Rekor)
 		updated bool
 	)
 	labels := constants.LabelsFor(actions.ServerComponentName, actions.ServerDeploymentName, instance.Name)
-	dp, err := utils.CreateRekorDeployment(instance, actions.ServerDeploymentName, actions.RBACName, labels)
+
+	insCopy := instance.DeepCopy()
+	if insCopy.Spec.Trillian.Address == "" {
+		insCopy.Spec.Trillian.Address = fmt.Sprintf("%s.%s.svc", actions2.LogserverDeploymentName, instance.Namespace)
```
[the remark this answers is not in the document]
osmman: I hope that in near feature we could solve these type of values via `Defaulter` and remove this ugly modification of CRD objects.

controllers/rekor/actions/pending.go
```diff
@@ -1,48 +0,0 @@
...
-func (i pendingAction) CanHandle(_ context.Context, instance *rhtasv1alpha1.Rekor) bool {
-	c := meta.FindStatusCondition(instance.Status.Conditions, constants.Ready)
-	return c.Reason == constants.Pending
-}
-
-func (i pendingAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Rekor) *action.Result {
-	var err error
-	_, err = utils.GetInternalUrl(ctx, i.Client, instance.Namespace, trillian.LogserverDeploymentName)
-	if err != nil {
-		meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
```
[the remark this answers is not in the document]
osmman: I removed this action because it is not needed and it will create problems for scenarios on which Trillian Log server will be deployed for example on different cluster to which the operator do not have access to it's API ([Scenario](https://issues.redhat.com/browse/SECURESIGN-573)).

Provided `TreeId` scenario, I think that this is simple because operator does not need any interaction with Trillian service. From here it is responsibility of how Rekor server is implemented to handle unavability of Trillian service. For example Trillian pod will be be rescheduled to new node (hw/network incident).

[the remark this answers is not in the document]
osmman: I tested and Rekor server will start without running Trillian log server and if there is some action to Trillian than it will correctly return error message like 
```
{"code":500,"message":"unexpected error communicating with transparency log"}
```

bouskaJ (approved):

osmman (approved):

PR #200, opened 2024-02-21
Restart on configuration change

bouskaJ:


cooktheryan:
did a quick test with this by modifying the OIDC configuration. It changed the configmap and notified fulcio to redeploy

cooktheryan:
Code wise nothing jumped out at me as incorrect. I did testing with two deployments and modified OIDC configuration no panics and logs were clean.

LGTM from a usage test

controllers/fulcio/actions/generate_cert.go
```diff
@@ -57,38 +69,41 @@ func (g generateCert) Handle(ctx context.Context, instance *v1alpha1.Fulcio) *ac
...
 
-	secret := k8sutils.CreateSecret(secretName, instance.Namespace, config.ToMap(), secretLabels)
-	if err = controllerutil.SetControllerReference(instance, secret, g.Client.Scheme()); err != nil {
+	newCert := k8sutils.CreateImmutableSecret(fmt.Sprintf("fulcio-cert-%s", instance.Name), instance.Namespace, cert.ToMap(), secretLabels)
+	if err = controllerutil.SetControllerReference(instance, newCert, g.Client.Scheme()); err != nil {
 		return g.Failed(fmt.Errorf("could not set controller reference for Secret: %w", err))
 	}
-	if updated, err = g.Ensure(ctx, secret); err != nil {
+	// ensure that only new key is exposed
+	if err = g.Client.DeleteAllOf(ctx, &v1.Secret{}, client.InNamespace(instance.Namespace), client.MatchingLabels(constants.LabelsFor(ComponentName, DeploymentName, instance.Name)), client.HasLabels{FulcioCALabel}); err != nil {
```
osmman, line 103: I think that old keys should be keept in cluster. The reason is that these CA key could be used to sign entry into transaction log, so it still have to be linked to CTlog configuration `rootCertificates`

controllers/fulcio/actions/servrConfig.go
```diff
@@ -30,44 +27,79 @@ func (i serverConfig) Name() string {
...
 	c := meta.FindStatusCondition(instance.Status.Conditions, constants.Ready)
-	return c.Reason == constants.Creating || c.Reason == constants.Ready
+	if c.Reason != constants.Creating && c.Reason != constants.Ready {
+		return false
+	}
+
+	if instance.Status.ServerConfigRef == nil {
+		return true
+	}
+	existing, err := kubernetes.GetConfigMap(ctx, i.Client, instance.Namespace, instance.Status.ServerConfigRef.Name)
```
osmman, line 39: Maybe better solution to sove this could be usage of single status field which will contains latest hash of spec content which has been reconciled. For example: 
```
spec:
   config:
     aaa: 
       IssuerURL: aaa
status:
   lastReconciled: b9c66b80ae026e9768222122e8063dd1b66f36ce03fdf2625b198e5e6d2669fa
```
if hash changed than execute reconciliation loop. Benefit will be more generic detection of changes in spec. Currently changes in monitoring or external access spec will not be reconciled. 

Other benefit is to remove kubernetes api requests from CanHandle method

controllers/fulcio/actions/generate_cert.go
bouskaJ: consider this PR as the baseline for the key rotation. I am sure that we will need something more sophisticated for key rotation support.

PR #213, opened 2024-02-26
modify the default value of database

### LINKED ISSUE #212: Modify the default boolean for Database creation
Database create is default true but should be default false

cooktheryan:
Closes #212


Gregory-Pereira (approved):
/lgtm

PR #225, opened 2024-02-27
include yaml in file and dont curl

cooktheryan:
Bringing the item creation into the shell script rather than pulling external


sallyom (approved):
LGTM

PR #227, opened 2024-02-28
Split Client server into two images

JasonPowr:
This pr makes the necessary changes to the client sever, after the split one contains cosign and gitsign the other contains rekor-cli and ec

### DIFF
--- a/controllers/constants/images.go
+++ b/controllers/constants/images.go
@@ -17,6 +17,9 @@ const (
 
 	TufImage = "quay.io/redhat-user-workloads/rhtas-tenant/scaffold/tuf-server@sha256:975bc755f80eefa615f796555f09997b4c7e200d19f1b50702ece63d9f942737"
 
-	CTLogImage        = "quay.io/redhat-user-workloads/rhtas-tenant/certificate-transparency-go/certificate-transparency-go@sha256:a851350d37cb2105c35c529a30f1b5432c85056e24c95838cfc39b0870187cc6"
-	ClientServerImage = "quay.io/redhat-user-workloads/rhtas-tenant/access-1-0-gamma/client-server-1-0-gamma@sha256:91caede7f666f380bd3e437444386a3818d89d50f28bb468c79294450c6bca9e"
+	CTLogImage = "quay.io/redhat-user-workloads/rhtas-tenant/certificate-transparency-go/certificate-transparency-go@sha256:a851350d37cb2105c35c529a30f1b5432c85056e24c95838cfc39b0870187cc6"
+
+	ClientServerImage    = "registry.access.redhat.com/ubi9/httpd-24@sha256:965f7b03ae8f45228bad765ce47bc8956711e854213df0e4cee8623d51317b0a"
+	ClientServerImage_cg = "quay.io/redhat-user-workloads/rhtas-tenant/cli/client-server-cg@sha256:5c7e83f13f9ba52c927bcb86abb5f9efa8ca4fe9ab7031709bde7f48829f6d0f"
+	ClientServerImage_re = "quay.io/redhat-user-workloads/rhtas-tenant/cli/client-server-re@sha256:d20c57b457650a025c225863462b2f9bf193ca35534dbf2cfcd9f6ab07a091b9"
 )
--- a/main.go
+++ b/main.go
@@ -62,6 +62,9 @@ const (
 	cliServerNs        = "trusted-artifact-signer"
 	cliServerName      = "cli-server"
 	cliServerComponent = "client-server"
+	sharedVolumeName   = "shared-data"
+	cliBinaryPath      = "/opt/app-root/src/clients/*"
+	cliWebServerPath   = "/var/www/html/clients/"
 
 	crdName = "securesigns.rhtas.redhat.com"
 )
@@ -284,15 +287,53 @@ func createClientserverDeployment(namespace string, labels map[string]string) *a
 					Labels: labels,
 				},
 				Spec: core.PodSpec{
+					Volumes: []core.Volume{
+						{
+							Name: sharedVolumeName,
+							VolumeSource: core.VolumeSource{
+								EmptyDir: &core.EmptyDirVolumeSource{},
+							},
+						},
+					},
+					InitContainers: []core.Container{
+						{
+							Name:    "init-shared-data-cg",
+							Image:   constants.ClientServerImage_cg,
+							Command: []string{"sh", "-c", fmt.Sprintf("cp -r %s %s", cliBinaryPath, cliWebServerPath)},
+							VolumeMounts: []core.VolumeMount{
+								{
+									Name:      sharedVolumeName,
+									MountPath: cliWebServerPath,
+								},
+							},
+						},
+						{
+							Name:    "init-shared-data-re",
+							Image:   constants.ClientServerImage_re,
+							Command: []string{"sh", "-c", fmt.Sprintf("cp -r %s %s", cliBinaryPath, cliWebServerPath)},
+							VolumeMounts: []core.VolumeMount{
+								{
+									Name:      sharedVolumeName,
+									MountPath: cliWebServerPath,
+								},
+							},
+						},
+					},
 					Containers: []core.Container{
 						{
 							Name:            cliServerName,
 							Image:           constants.ClientServerImage,
-							ImagePullPolicy: "IfNotPresent",
+							ImagePullPolicy: core.PullAlways,
 							Ports: []core.ContainerPort{
 								{
 									ContainerPort: 8080,
-									Protocol:      "TCP",
+									Protocol:      core.ProtocolTCP,
+								},
+							},
+							VolumeMounts: []core.VolumeMount{
+								{
+									Name:      sharedVolumeName,
+									MountPath: cliWebServerPath,
 								},
 							},
 						},


bouskaJ:
@JasonPowr I don't like the concept of having two pods at all TBH. I know you have reason to do that so I am thinking about having both container running in the single pod. That would need containers to serve on different port. Would it be possible to configure it using ENV variable or so?

JasonPowr:
> @JasonPowr I don't like the concept of having two pods at all TBH. I know you have reason to do that so I am thinking about having both container running in the single pod. That would need containers to serve on different port. Would it be possible to configure it using ENV variable or so?

I agree its not the nicest solution, I'm not sure if I can configure it using a Env var but ill look into it. The main reason for the separate pods is because Apache web sever automatically binds to port 8080 by default, and when running the two images in the same pod it results in a port binding error

cooktheryan:
@JasonPowr stepping back what is the why? To make the change

JasonPowr:
> @JasonPowr stepping back what is the why? To make the change

@cooktheryan I believe its because the client server image got too big during TP2 and was breaking builds: https://issues.redhat.com/browse/SECURESIGN-624

JasonPowr:
@bouskaJ So I don't think there is a way I can set it up using env vars, If we don't want it running two pods I believe we may have two options

We could do something like this:
```
Containers: []core.Container{
	{
		Name:            cliServerName,
		Image:           image,
		ImagePullPolicy: "IfNotPresent",
		Ports: []core.ContainerPort{
			{
				ContainerPort: 8080,
				Protocol:      "TCP",
			},
		},
		Command: []string{"/bin/sh", "-c"},
		Args: []string{
			"sed -i 's/Listen 0.0.0.0:8080/Listen 0.0.0.0:8081/' /etc/httpd/conf/httpd.conf && run-httpd",
		},
	},
},
```

Or we could move the sed command to the build process of the second client server image, wdyt?

lance:
> > @JasonPowr stepping back what is the why? To make the change
> 
> @cooktheryan I believe its because the client server image got too big during TP2 and was breaking builds: https://issues.redhat.com/browse/SECURESIGN-624

Yes - it's not the image itself, but the `source-image-build` task results are too large for konflux. This is because it was including all of the source for cosign, gitsign, rekor-cli and ec. Once we added ec to the mix it went over the task result size limit, and that was not something easily fixable by the konflux team iiuc.

lance:
> We could do something like this:
> 
> ```
> Containers: []core.Container{
> 	{
> 		Name:            cliServerName,
> 		Image:           image,
> 		ImagePullPolicy: "IfNotPresent",
> 		Ports: []core.ContainerPort{
> 			{
> 				ContainerPort: 8080,
> 				Protocol:      "TCP",
> 			},
> 		},
> 		Command: []string{"/bin/sh", "-c"},
> 		Args: []string{
> 			"sed -i 's/Listen 0.0.0.0:8080/Listen 0.0.0.0:8081/' /etc/httpd/conf/httpd.conf && run-httpd",
> 		},
> 	},
> },
> ```
> 
> Or we could move the sed command to the build process of the second client server image, wdyt?

IMO the container should default to 8080. The reason we are having to do all of these shenanigans is because we want both servers to run from the same pod. So I think having the `sed` command to hack the port here in the Container command args is the right place for it. Don't forget to set `ContainerPort` to 8081 on that Container spec too.

bouskaJ:
> > We could do something like this:
> > ```
> > Containers: []core.Container{
> > 	{
> > 		Name:            cliServerName,
> > 		Image:           image,
> > 		ImagePullPolicy: "IfNotPresent",
> > 		Ports: []core.ContainerPort{
> > 			{
> > 				ContainerPort: 8080,
> > 				Protocol:      "TCP",
> > 			},
> > 		},
> > 		Command: []string{"/bin/sh", "-c"},
> > 		Args: []string{
> > 			"sed -i 's/Listen 0.0.0.0:8080/Listen 0.0.0.0:8081/' /etc/httpd/conf/httpd.conf && run-httpd",
> > 		},
> > 	},
> > },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or we could move the sed command to the build process of the second client server image, wdyt?
> 
> IMO the container should default to 8080. The reason we are having to do all of these shenanigans is because we want both servers to run from the same pod. So I think having the `sed` command to hack the port here in the Container command args is the right place for it. Don't forget to set `ContainerPort` to 8081 on that Container spec too.

yes, I agree with @lance I would not hack the container image. I still think that having both in one pod (including the hack) is less messy than having two deployments but let's hear from others @osmman @cooktheryan WDYT?

osmman:
> > > We could do something like this:
> > > ```
> > > Containers: []core.Container{
> > > 	{
> > > 		Name:            cliServerName,
> > > 		Image:           image,
> > > 		ImagePullPolicy: "IfNotPresent",
> > > 		Ports: []core.ContainerPort{
> > > 			{
> > > 				ContainerPort: 8080,
> > > 				Protocol:      "TCP",
> > > 			},
> > > 		},
> > > 		Command: []string{"/bin/sh", "-c"},
> > > 		Args: []string{
> > > 			"sed -i 's/Listen 0.0.0.0:8080/Listen 0.0.0.0:8081/' /etc/httpd/conf/httpd.conf && run-httpd",
> > > 		},
> > > 	},
> > > },
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Or we could move the sed command to the build process of the second client server image, wdyt?
> > 
> > 
> > IMO the container should default to 8080. The reason we are having to do all of these shenanigans is because we want both servers to run from the same pod. So I think having the `sed` command to hack the port here in the Container command args is the right place for it. Don't forget to set `ContainerPort` to 8081 on that Container spec too.
> 
> yes, I agree with @lance I would not hack the container image. I still think that having both in one pod (including the hack) is less messy than having two deployments but let's hear from others @osmman @cooktheryan WDYT?

Did you think about usage of sidecars? Basically to have one containers with HTTP server and sidecars with binaries

bouskaJ:
> > > > We could do something like this:
> > > > ```
> > > > Containers: []core.Container{
> > > > 	{
> > > > 		Name:            cliServerName,
> > > > 		Image:           image,
> > > > 		ImagePullPolicy: "IfNotPresent",
> > > > 		Ports: []core.ContainerPort{
> > > > 			{
> > > > 				ContainerPort: 8080,
> > > > 				Protocol:      "TCP",
> > > > 			},
> > > > 		},
> > > > 		Command: []string{"/bin/sh", "-c"},
> > > > 		Args: []string{
> > > > 			"sed -i 's/Listen 0.0.0.0:8080/Listen 0.0.0.0:8081/' /etc/httpd/conf/httpd.conf && run-httpd",
> > > > 		},
> > > > 	},
> > > > },
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Or we could move the sed command to the build process of the second client server image, wdyt?
> > > 
> > > 
> > > IMO the container should default to 8080. The reason we are having to do all of these shenanigans is because we want both servers to run from the same pod. So I think having the `sed` command to hack the port here in the Container command args is the right place for it. Don't forget to set `ContainerPort` to 8081 on that Container spec too.
> > 
> > 
> > yes, I agree with @lance I would not hack the container image. I still think that having both in one pod (including the hack) is less messy than having two deployments but let's hear from others @osmman @cooktheryan WDYT?
> 
> Did you think about usage of sidecars? Basically to have one containers with HTTP server and sidecars with binaries

Good idea @osmman I like it! Here is a nice example with nginx server https://kubernetes.io/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume/

cooktheryan:
I would like to see if there is any way to fix this in Konflux. This seems wrong and something more folks are going to face in the future.

cooktheryan:
+1 @osmman idea that would be a very smooth way to solve this

JasonPowr:
@bouskaJ @osmman So to do this Ill need to make changes to the client-server images and remove the HTTP aspects of them, before I begin the process are you looking for something like this or am I misunderstanding?

```
Spec: core.PodSpec{
	Volumes: []core.Volume{
		{
			Name: "shared-data",
			VolumeSource: core.VolumeSource{
				EmptyDir: &core.EmptyDirVolumeSource{},
			},
		},
	},
	InitContainers: []core.Container{
		{
			Name:    "init-shared-data-cg",
			Image:   "quay.io/japower/client_server_test:latest",
			Command: []string{"sh", "-c", "cp -r /opt/app-root/src/clients/* /var/www/html/clients/"},
			VolumeMounts: []core.VolumeMount{
				{
					Name:      "shared-data",
					MountPath: "/var/www/html/clients/",
				},
			},
		},
		{
			Name:    "init-shared-data-re",
			Image:   "quay.io/japower/client_server_test_re:latest",
			Command: []string{"sh", "-c", "cp -r /opt/app-root/src/clients/* /var/www/html/clients/"},
			VolumeMounts: []core.VolumeMount{
				{
					Name:      "shared-data",
					MountPath: "/var/www/html/clients/",
				},
			},
		},
	},
	Containers: []core.Container{
		{
			Name:            cliServerName,
			Image:           "registry.access.redhat.com/ubi9/httpd-24@sha256:965f7b03ae8f45228bad765ce47bc8956711e854213df0e4cee8623d51317b0a",
			ImagePullPolicy: core.PullAlways,
			Ports: []core.ContainerPort{
				{
					ContainerPort: 8080,
					Protocol:      core.ProtocolTCP,
				},
			},
			VolumeMounts: []core.VolumeMount{
				{
					Name:      "shared-data",
					MountPath: "/var/www/html/clients/",
				},
			},
		},
	},
},
```

cooktheryan:
this is a very smooth implementation of this idea +1

osmman (approved):

PR #235, opened 2024-03-01
run make bundle

JasonPowr:
In order to generate the correct catalog.json, files I need a bundle with the correct version for GA this prjust runs make bundle with the correct version


JasonPowr:
@bouskaJ @Gregory-Pereira @cooktheryan @lance @osmman Can I get a lgtm for someone pls ? :)

PR #250, opened 2024-03-06
Update Konflux CEL expressions 

JasonPowr:
This pr makes changes to the tekton cel expressions, this makes sure that the fbc images, the bundle and the operator can build independently. 

## Explanation (Copied from slack)
"I think having the operator and the bundle in the same application in Konflux causes some issues.When we generate an image for the operator-controller in Konflux, we then go back and update the bundle. When we update the bundle it generates a new image for the operator-controller, this means that when we release we use a snapshot that has an up to date controller image, but the image being used in the bundle isn’t the one in the snapshot, it's the one generated previously. If we were to separate them into different applications they could be built independently, wdyt? Or probably a better solution would be to add the cel expression to the operator pipelines?"

### DIFF
--- a/.tekton/rhtas-operator-pull-request.yaml
+++ b/.tekton/rhtas-operator-pull-request.yaml
@@ -7,7 +7,8 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main"
+    pipelinesascode.tekton.dev/on-cel-expression: | 
+      event == "pull_request" && target_branch == "main" && (".tekton/rhtas-operator-pull-request.yaml".pathChanged() || "api/***".pathChanged() || "config/***".pathChanged() || "controllers/***".pathChanged() || "Dockerfile.rhtas-operator.rh".pathChanged() || "go.mod".pathChanged() || "main.go".pathChanged() || "Makefile".pathChanged() || "go.sum".pathChanged() )
   creationTimestamp: null
   labels:
     appstudio.openshift.io/application: operator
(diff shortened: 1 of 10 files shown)


.tekton/rhtas-operator-pull-request.yaml
```diff
@@ -7,7 +7,8 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main"
+    pipelinesascode.tekton.dev/on-cel-expression: | 
+      event == "pull_request" && target_branch == "main" && (".tekton/rhtas-operator-pull-request.yaml".pathChanged() || "api/***".pathChanged() || "config/***".pathChanged() || "controllers/***".pathChanged() || "Dockerfile.rhtas-operator.rh".pathChanged() || "go.mod".pathChanged() || "main.go".pathChanged() )
```
lance, line 11: I don't love this, since it won't catch new files or directories at the root. But I'm not sure what is a better way to handle it. You might add `go.sum` and `Makefile` to this list.

.tekton/rhtas-operator-push.yaml
```diff
@@ -6,7 +6,8 @@ metadata:
     build.appstudio.redhat.com/commit_sha: '{{revision}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "push" && target_branch == "main"
+    pipelinesascode.tekton.dev/on-cel-expression: | 
+      event == "push" && target_branch == "main" && (".tekton/rhtas-operator-push.yaml".pathChanged() || "api/***".pathChanged() || "config/***".pathChanged() || "controllers/***".pathChanged() || "Dockerfile.rhtas-operator.rh".pathChanged() || "go.mod".pathChanged() || "go.sum".pathChanged() || "main.go".pathChanged() )
```
lance, line 10: As above with `go.sum` and `Makefile`.

lance (changes requested):
Can you add `go.sum` and `Makefile` to the operator pipelines?

.tekton/rhtas-operator-pull-request.yaml
JasonPowr: I get what your saying, but at the same time if a new directory or file is introduced that affects the build of the operator, there would  be changes in the directories listed in the expression to reflect whatever changes are needed. 

I've added `go.sum` and `Makefile` :)

PR #253, opened 2024-03-06
chore: Update to latest images 

JasonPowr:
Updating the images to the latest snapshots.

### DIFF
--- a/controllers/constants/images.go
+++ b/controllers/constants/images.go
@@ -1,25 +1,25 @@
 package constants
 
 const (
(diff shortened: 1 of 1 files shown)


osmman:
It looks like there is some problem with certificate-transparency-go image.

JasonPowr:
> It looks like there is some problem with certificate-transparency-go image. Ctlog pod is stuck in pending state.

Thanks Tomas i'll investigate this :)

JasonPowr:
I've deployed this branch to a Rosa cluster and it seems to be deploying fine, ct-log comes up and is running and ready, im also able to sign and verify without issue

JasonPowr:
passing now

PR #254, opened 2024-03-06
chore: bump fulcio, trillian and ct

lance:
Use the latest versions of these dependencies to eliminate failures in snyk.


osmman:
Hi @lance, unfortunately current Github actions does not work agains fork repositories and it will fail because of missing secret credentials for podman.

PR #256, opened 2024-03-08
[SECURESIGN-655] Enable OIDCs with private or self-signed certificates

### LINKED ISSUE #206: Fulcio CA certificate issue
Currently, if a certificate is private or self signed fulcio will require the following. `oc set env -n fulcio-system deployment/fulcio-server SSL_CERT_DIR=/var/run/fulcio` 

We need to try to see if the new Fulcio code fixes this or if we will need to code in our own fix

osmman:
Resolve #206


controllers/fulcio/utils/fulcio_deployment.go
```diff
@@ -56,14 +60,41 @@ func CreateDeployment(instance *v1alpha1.Fulcio, deploymentName string, sa strin
 		args = append(args, "--fileca-key-passwd", "$(PASSWORD)")
 	}
 
+	oidcInfo := make([]corev1.VolumeProjection, 0)
+	// Integration with https://kubernetes.default.svc" OIDC issuer and ctlog service
+	oidcInfo = append(oidcInfo, corev1.VolumeProjection{
```
bouskaJ, line 65: We can give it better name

controllers/fulcio/utils/fulcio_deployment.go
```diff
@@ -41,6 +40,11 @@ func CreateDeployment(instance *v1alpha1.Fulcio, deploymentName string, sa strin
 		fmt.Sprintf("--ct-log-url=http://ctlog.%s.svc/trusted-artifact-signer", instance.Namespace)}
 
 	env := make([]corev1.EnvVar, 0)
+	env = append(env, corev1.EnvVar{
+		Name:  "SSL_CERT_DIR",
```
bouskaJ, line 44: :thinking: Can we have use-case where user want to remove default truststore and trust only provided certs? If we want to support that we can say _ConfigMap with <s>additional</s> bundle of trusted CA_ and mount it directly to the default ssl cert directory.

osmman: I do not know if some one really want to remove all system trusted CA and trust only provided. It will defiantly add much more complexity to our deployment and format of config map because it will have to be specifically structured (key name) to be correctly consumed.

bouskaJ: It was just idea. We can improve it in future if needed.

Issue #38, opened 2024-01-15
[SECURESIGN-571] Test deploying components independently


cooktheryan:
With the architecture we have chosen we should be able to independently deploy and test the individual components. It would be important to figure out what is the bare minimum to deploy each components. Such as Rekor requiring TUF(it probably doesn't but I need an example)

osmman:
I have a PoC which enables TUF service to be independently deployed. https://github.com/securesign/secure-sign-operator/pull/35

bouskaJ:
Migrated to Jira

PR #260, opened 2024-03-12
Implementation of Segment Backup Job and Cronjob

Gregory-Pereira:


api/v1alpha1/securesign_types.go
```diff
@@ -28,6 +28,9 @@ type SecuresignSpec struct {
 	Rekor    RekorSpec    `json:"rekor,omitempty"`
 	Fulcio   FulcioSpec   `json:"fulcio,omitempty"`
 	Trillian TrillianSpec `json:"trillian,omitempty"`
+	// SegmentBackup CronJob Configuration
+	//+kubebuilder:default:={enabled: true, schedule: "0 0 * * *"}
+	SegmentBackupCronJob SegmentBackupCronJob `json:"SegmentBackupCronJob,omitempty"`
```
osmman, line 33: I will rather choose a different name. I think that simple `metrics` should be better

osmman: At the moment, I thought of another way to set it up. We can use labels/annotations to manage metric feature, because this way it will only work for Securesign CRD and will cause problem when we will implement https://issues.redhat.com/browse/SECURESIGN-571.
Example to disable metrics:
```
kind: Securesign
metadata:
  labels:
    rhtas.redhat.com/metrics: false
    app.kubernetes.io/name: securesign-rotate
  name: securesign-rotate
spec:
  ctlog:
  rekor:
  trillian:
  fulcio:
  tuf:
```

I think that CronJob's schedule should not be configurable by user.
What do you think? @Gregory-Pereira @lance @cooktheryan

controllers/securesign/actions/constants.go
```diff
@@ -1,9 +1,15 @@
...
+	FulcioCondition          = "FulcioAvailable"
+	RekorCondition           = "RekorAvailable"
+	TrillianCondition        = "TrillianAvailable"
+	CTlogCondition           = "CTlogAvailable"
+	SegmentBackupCronJobName = "segment-backup-nightly-metrics"
+	SegmentBackupJobName     = "segment-backup-installation"
+	SegmentRBACName          = "rhtas-segment-backup-job"
+	SecuresignCondition      = "SecuresignAvailable"
+	AnalyiticsCronSchedule   = " 0 0 * * *"
+	OperatorNamespace        = "openshift-rhtas-operator"
```
bouskaJ, line 14: I would be careful defining operator namespace like that. OLM installation would install operator to the `openshift-operator` namespace.

controllers/securesign/actions/segment_backup_cronjob.go
```diff
@@ -0,0 +1,176 @@
...
+
+type segmentBackupCronJob struct {
+	action.BaseAction
+}
+
+func (i segmentBackupCronJob) Name() string {
+	return "segment-backup-nightly-metrics"
+}
+func (i segmentBackupCronJob) CanHandle(_ context.Context, instance *rhtasv1alpha1.Securesign) bool {
+	labels := instance.GetLabels()
```
bouskaJ, line 39: Using annotation instead of the label would be more suitable in this case IMHO (see https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)

controllers/securesign/actions/segment_backup_cronjob.go
```diff
@@ -0,0 +1,176 @@
...
+	})
+
+	if err = ctrl.SetControllerReference(instance, role, i.Client.Scheme()); err != nil {
+		return i.Failed(fmt.Errorf("could not set controll reference for role: %w", err))
+	}
+	_, err = i.Ensure(ctx, role)
+
+	rb := kubernetes.CreateClusterRoleBinding(instance.Namespace, SegmentRBACName, labels, rbacv1.RoleRef{
+		APIGroup: v1.SchemeGroupVersion.Group,
+		Kind:     "Role",
```
bouskaJ, line 99:
```suggestion
		Kind:     "ClulsterRole",
```

controllers/securesign/actions/segment_backup_job.go
```diff
@@ -0,0 +1,115 @@
...
+	})
+
+	if err = ctrl.SetControllerReference(instance, role, i.Client.Scheme()); err != nil {
+		return i.Failed(fmt.Errorf("could not set controll reference for role: %w", err))
+	}
+	_, err = i.Ensure(ctx, role)
+
+	rb := kubernetes.CreateClusterRoleBinding(instance.Namespace, SegmentRBACName, labels, rbacv1.RoleRef{
+		APIGroup: v1.SchemeGroupVersion.Group,
+		Kind:     "Role",
```
bouskaJ, line 89:
```suggestion
		Kind:     "ClusterRole",
```

controllers/securesign/securesign_controller.go
```diff
@@ -101,7 +102,8 @@ func (r *SecuresignReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		actions.NewRekorAction(),
 		actions.NewCtlogAction(),
 		actions.NewTufAction(),
-
+		actions.NewSegmentBackupCronJobAction(),
```
bouskaJ, line 105: We should care every component separately and create action(s) for each one. The `securesign` resource is only an umbrella and it can be ommited(see https://github.com/securesign/secure-sign-operator/issues/38).

controllers/common/utils/kubernetes/job.go
```diff
@@ -0,0 +1,42 @@
...
+			ActiveDeadlineSeconds: &activeDeadlineSeconds,
+			BackoffLimit:          &backoffLimit,
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					ServiceAccountName: serviceAccountName,
+					RestartPolicy:      "OnFailure",
+					Containers: []corev1.Container{
+						{
+							Name:    name,
+							Image:   image,
```
bouskaJ, line 28: Util functions from `common` package should be generic and reusable. Replace `segment-job` specifics with parameters ( `Command` and ` Env` fields)

controllers/securesign/actions/segment_backup_cronjob.go
```diff
@@ -0,0 +1,176 @@
...
+			Resources: []string{"configmaps", "secrets"},
+			Verbs:     []string{"get", "list"},
+		},
+		{
+			APIGroups: []string{"route.openshift.io"},
+			Resources: []string{"routes"},
+			Verbs:     []string{"get", "list"},
+		},
+		{
+			APIGroups: []string{"operator.openshift.io/v1"},
```
bouskaJ, line 86:
```suggestion
			APIGroups: []string{"operator.openshift.io"},
```

controllers/securesign/actions/segment_backup_cronjob.go
```diff
@@ -0,0 +1,176 @@
...
+			Resources: []string{"console"},
```
bouskaJ, line 87:
```suggestion
			Resources: []string{"consoles"},
```

controllers/securesign/actions/segment_backup_job.go
```diff
@@ -0,0 +1,115 @@
...
+			Resources: []string{"configmaps", "secrets"},
+			Verbs:     []string{"get", "list"},
+		},
+		{
+			APIGroups: []string{"route.openshift.io"},
+			Resources: []string{"routes"},
+			Verbs:     []string{"get", "list"},
+		},
+		{
+			APIGroups: []string{"operator.openshift.io/v1"},
```
bouskaJ, line 76:
```suggestion
			APIGroups: []string{"operator.openshift.io"},
```

controllers/securesign/actions/segment_backup_job.go
```diff
@@ -0,0 +1,115 @@
...
+			Resources: []string{"console"},
```
bouskaJ, line 77:
```suggestion
			Resources: []string{"consoles"},
```

controllers/securesign/securesign_controller.go
Gregory-Pereira: Does this mean you suggest moving my implementation of this to the initial install and not the creation of the `securesign`?

bouskaJ: There are two ways of RHTAS installation
1) create securesign resource (your implementation will work)
2) create fulcio, ctlog, rekor, trillian, tuf resources (and do not create the securesign) - in this case, your code will not be invoked

I am suggesting to move your implementation to all component controllers (fulcio ctlog, rekor, trillian, tuf)

osmman: I think that implementation for second option can be postponed to feature release. @Gregory-Pereira can you create issue and add it into [SECURESIGN-610](https://issues.redhat.com/browse/SECURESIGN-610) epic.

osmman:
MR requires rebase

bundle/manifests/rhtas.redhat.com_securesigns.yaml
```diff
@@ -337,7 +337,7 @@ spec:
                     description: Enable Service monitors for fulcio
                     properties:
                       enabled:
-                        default: false
+                        default: true
```
osmman, line 340: Default values is false, if you need it to change it than it has to be done in different 
 places:
 - https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/common.go#L20 
 - https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/fulcio_types.go#L20
 - https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/rekor_types.go#L19

controllers/securesign/actions/segment_backup_cronjob.go
Gregory-Pereira: I agree that this is a better implementation but due to time constraints I think we should move this to  release `1.0.1`

config/samples/rhtas_v1alpha1_securesign.yaml
```diff
@@ -6,12 +6,13 @@ metadata:
     app.kubernetes.io/instance: securesign-sample
     app.kubernetes.io/part-of: trusted-artifact-signer
   name: securesign-sample
+  namespace: rhtas-operator
```
osmman, line 9: Please do not set namespace in samples. 
```suggestion
```

JasonPowr:
@Gregory-Pereira When I deployed this it was giving out about the pyYaml dependency and the pod was failing my commit fixed this by updating to the newest image, I also removed the `namespace: rhtas-operator` CC: @osmman. When I deploy now the job completes, but I'm seeing this in the log 

```
route "thanos-querier" is up and running in namespace openshift-monitoring.
console-openshift-console.apps.rosa.a2yk8-vmqnb-jck.vxfp.p3.openshiftapps.com
rosa.a2yk8-vmqnb-jck.vxfp.p3.openshiftapps.com
mapping values are not allowed here
  in "config.yaml", line 1, column 68
```

I'm not at all familiar with this, so I dunno if this is supposed to happen or not, doesn't seem like it

Gregory-Pereira:
> apping values are not allowed here
>   in "config.yaml", line 1, column 68

Could you share what type of cluster you are on? This script dumps the `cluster-monitoring-config` to the config.yaml file and reads it in python. I think the issue is that it is stored as unformatted yaml and when reading it in something is formatted incorrectly. Could you maybe dump that file? `oc get configmap cluster-monitoring-config -n openshift-monitoring`

JasonPowr:
> > apping values are not allowed here
> > in "config.yaml", line 1, column 68
> 
> Could you share what type of cluster you are on? This script dumps the `cluster-monitoring-config` to the config.yaml file and reads it in python. I think the issue is that it is stored as unformatted yaml and when reading it in something is formatted incorrectly. Could you maybe dump that file? `oc get configmap cluster-monitoring-config -n openshift-monitoring`

I was using a rosa cluster created using cluster bot, version 4.14

PR #261, opened 2024-03-12
Fix validation error

bouskaJ:


controllers/tuf/actions/generate_cert_test.go
osmman, line 1: it would be nice to extend tests to cover CanHandle method and scenarios which should not trigger action

osmman: I will create a new PR to solve that

PR #266, opened 2024-03-15
feat(config): add config action for rekor-ui

kahboom:
This PR adds reference to a ConfigMap with the required environment variable in order for the `rekor-search-ui` to be able to issue requests to the Rekor server.

PR #268, opened 2024-03-15
feat(config): add env var for rekor endpoint to search-ui deployment

kahboom:
This PR adds the required environment variable (`NEXT_PUBLIC_REKOR_DEFAULT_DOMAIN`) in order for the `rekor-search-ui` to be able to issue requests to the Rekor server.

We've since decided to define the variable directly, rather than through a ConfigMap.

Original PR: https://github.com/securesign/secure-sign-operator/pull/266


controllers/rekor/utils/rekor_search_ui_deployment.go
```diff
@@ -7,13 +7,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func CreateRekorSearchUiDeployment(namespace, dpName string, sa string, labels map[string]string) *apps.Deployment {
+func CreateRekorSearchUiDeployment(instance *v1alpha1.Rekor, dpName string, sa string, labels map[string]string) *apps.Deployment {
```
kahboom, line 10: I've updated the signature here but surely this means I need to make a change elsewhere. 😂  First day of golang for me.

osmman: Yes, the change has to be done here https://github.com/securesign/secure-sign-operator/blob/ed26066dfc7959eeb2dfe224944f5638c2b5d549/controllers/rekor/actions/ui/deployment.go#L41

kahboom:
I've added some instructions for testing the PR, hopefully it helps as I didn't previously share the container image for `rekor-search-ui` I had been using (sorry!): https://gist.github.com/kahboom/3294d64b5186b3c79b68bd78923ebec5#file-sigstore-ocp-md

osmman:
Please, can you squash commits into single commit

controllers/rekor/utils/rekor_search_ui_deployment.go
```diff
@@ -29,7 +30,13 @@ func CreateRekorSearchUiDeployment(namespace, dpName string, sa string, labels m
 					ServiceAccountName: sa,
 					Containers: []core.Container{
 						{
-							Name:  dpName,
+							Name: dpName,
+							Env: []core.EnvVar{
+								{
+									Name:  "NEXT_PUBLIC_REKOR_DEFAULT_DOMAIN",
+									Value: instance.Status.Url,
```
osmman, line 37: I found in failed e2e log that this value is empty when Recor Search UI is created

https://github.com/securesign/secure-sign-operator/actions/runs/8330192759/job/22795693014?pr=268#step:12:4712

osmman: Unfortunatly, we are generating this values in this [action](https://github.com/securesign/secure-sign-operator/blob/main/controllers/rekor/rekor_controller.go#L113) which is executed after ui deployment [action](https://github.com/securesign/secure-sign-operator/blob/main/controllers/rekor/rekor_controller.go#L104).

JasonPowr: I suppose there's two options around this, we could move the deployment of the rekor-search-ui until after the `server.NewInitializeAction(),` or we could create a new action to initialise the Env var after its set?
@osmman @kahboom wdyt?

JasonPowr: I forgot we already have an initialise action for rekor-search-ui:
```
	env := []core.EnvVar{
		{
			Name:  "NEXT_PUBLIC_REKOR_DEFAULT_DOMAIN",
			Value: instance.Status.Url,
		},
	}
	deployment := &apps.Deployment{}
	err = i.Client.Get(ctx, types.NamespacedName{Name: actions.SearchUiDeploymentName, Namespace: instance.Namespace}, deployment)
	if err != nil {
		return i.FailedWithStatusUpdate(ctx, err, instance)
	}
	deployment.Spec.Template.Spec.Containers[0].Env = append(deployment.Spec.Template.Spec.Containers[0].Env, env...)

```
Something like this should work when placed in the Handle function of the controllers/rekor/actions/ui/initialize.go file

osmman: Hi I think thatsimples solution will be to create a new action which will configure that value after `server.NewIngressAction(),` action. I created [this commit](https://github.com/securesign/secure-sign-operator/commit/e21075e7ad995fb9f41e5e0bb69fc9664a792228) which you can cherry pick into your PR.

osmman: @bouskaJ what is your opinion?

bouskaJ: @JasonPowr I would tend to use option nr. 1 (`we could move the deployment of the rekor-search-ui until after the server.NewInitializeAction()`) I would rather avoid nr 2 because non-working pod would be created and then redeployed (after the Deployment resource modification). The whole deployment would not be transparent.

We can use rebase on @osmman commit but still, the `ui.*` actions needs to be called after the `server.NewStatusUrlAction()`.

JasonPowr:
@bouskaJ @osmman can you re-review ?

osmman (approved):

PR #270, opened 2024-03-19
run-make-bundle

JasonPowr:
run-make-bundle


JasonPowr:
I'm gonna override the openshift CI bits for this pr, there seems to be an outage, and this pr doesn't changed much and all other tests are passing

PR #274, opened 2024-03-20
Apply securityContext also for OpenShift env

bouskaJ:


bouskaJ:
@cooktheryan was there any particular reason why you skipped the security context definition on OCP cluster?

osmman (approved):

osmman:
@bouskaJ it looks that there is some problem to

PR #280, opened 2024-03-21
[SECURESIGN-681] Move TreeID to status field

bouskaJ:


api/v1alpha1/ctlog_types.go
```diff
@@ -43,6 +43,7 @@ type CTlogStatus struct {
 	PrivateKeyPasswordRef *SecretKeySelector    `json:"privateKeyPasswordRef,omitempty"`
 	PublicKeyRef          *SecretKeySelector    `json:"publicKeyRef,omitempty"`
 	RootCertificates      []SecretKeySelector   `json:"rootCertificates,omitempty"`
+	TreeID                *int64                `json:"treeID,omitempty"`
```
osmman, line 46: Please add doc

bouskaJ: Do we need DOC also for status fields? There is no doc in any component for status fields.

osmman: I would like to start document these status field too because list of them start to grow and these docs can be useful for users to debug cluster (show tips in OCP UI).

Example:
```
// The current ID of a Trillian tree which is used to store log data
```

PR #281, opened 2024-03-22
[SECURESIGN-683] Fix autodiscovery of Fulcio Root CA in CTlog

osmman:


cooktheryan:
@osmman testing is going to be broken right now due to a quay and registry.redhat outage

JasonPowr (approved):
/lgtm

PR #292, opened 2024-03-28
[SECURESIGN-704] Create ClusterRoleBinding per instance

bouskaJ:


Gregory-Pereira (approved):
/lgtm

I have tested this with the new `graceful-failures` SBJ image and it works. You the man Jan!

PR #301, opened 2024-04-02
fix: Use roles instead of cluster roles for SBJ

Gregory-Pereira:
/cc @securesign/qe @lance

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -161,6 +161,14 @@ rules:
 - apiGroups:
   - monitoring.coreos.com
   resources:
+  - prometheuses/api
+  verbs:
+  - create
+  - get
+  - update
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
   - servicemonitors
   verbs:
   - create
@@ -233,6 +241,7 @@ rules:
   verbs:
   - create
   - delete
+  - deletecollection
   - get
   - list
   - patch
@@ -245,6 +254,7 @@ rules:
   verbs:
   - create
   - delete
+  - deletecollection
   - get
   - list
   - patch
@@ -257,6 +267,7 @@ rules:
   verbs:
   - create
   - delete
+  - deletecollection
   - get
   - list
   - patch
--- a/config/samples/rhtas_v1alpha1_securesign.yaml
+++ b/config/samples/rhtas_v1alpha1_securesign.yaml
@@ -5,14 +5,15 @@ metadata:
     app.kubernetes.io/name: securesign-sample
     app.kubernetes.io/instance: securesign-sample
     app.kubernetes.io/part-of: trusted-artifact-signer
+  annotations:
+    rhtas.redhat.com/metrics: "true"
   name: securesign-sample
-  namespace: rhtas-operator
 spec:
   rekor:
     externalAccess:
       enabled: true
     monitoring:
-      enabled: false
+      enabled: true
   trillian:
     database:
       create: true
@@ -30,7 +31,7 @@ spec:
       organizationEmail: jdoe@redhat.com
       commonName: fulcio.hostname
     monitoring:
-      enabled: false
+      enabled: true
   tuf:
     externalAccess:
       enabled: true
--- a/controllers/securesign/actions/rbac.go
+++ b/controllers/securesign/actions/rbac.go
@@ -13,10 +13,14 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-const namespacedNamePattern = SegmentRBACName + "-%s"
+const (
+	namespacedNamePattern  = SegmentRBACName + "-%s"
+	clusterWideNamePattern = SegmentRBACName + "-%s" + "-%s"
+	OpenshiftMonitoringNS  = "openshift-monitoring"
+)
 
 func NewRBACAction() action.Action[rhtasv1alpha1.Securesign] {
 	return &rbacAction{}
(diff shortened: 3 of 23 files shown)


lance:
@Gregory-Pereira I am seeing this in the e2e test logs
```
error: the namespace from the provided object "rhtas-operator" does not match the namespace "securesign". You must pass '--namespace=rhtas-operator' to perform this operation.
```

lance:
The konflux build failed on `build-source-image` with this error, which I suspect may be transient.

```
Copying blob sha256:5038e19ffe9670630d6f85c37de580326cf0a715f5bac4dfbceee8ccb3236c12
time="2024-04-02T20:12:53Z" level=fatal msg="reading blob sha256:aea816dcd21789d1ec292c9318fb768591fd6e5a00baa492582a39d332b306c2: fetching blob: received unexpected HTTP status: 502 Bad Gateway"
2024-04-02 20:12:53,178:source-build:ERROR:failed to build source image
```

Gregory-Pereira:
These errors I think are due to the wrong default toggle in the `api/v1/common/monitoring.go`. Fixing

Gregory-Pereira:
Owner references have to be dropped:
```log
ERROR ensure RBAC for segment job error during action execution {"controller": "securesign", "controllerGroup": "rhtas.redhat.com", "controllerKind": "Securesign", "Securesign": {"name":"securesign-sample-testing","namespace":"testing"}, "namespace": "testing", "name": "securesign-sample-testing", "reconcileID": "d4855b74-a45c-460e-9650-bab52d833aa5", "error": "could not set controll reference for role: cross-namespace owner references are disallowed, owner's namespace testing, obj's namespace openshift-monitoring"}
```

controllers/securesign/actions/rbac.go
```diff
@@ -57,41 +58,103 @@ func (i rbacAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Securesi
...
+		Kind:     "Role",
+		Name:     fmt.Sprintf(namespacedResourcedNamePattern, tmpNamespace, "configmaps"),
+	},
+		[]rbacv1.Subject{
+			{Kind: "ServiceAccount", Name: SegmentRBACName, Namespace: instance.Namespace},
+		})
+
+	i.Ensure(ctx, rolebindingOpenshiftMontiroingConfigmaps)
+
+	roleOpenshiftMontiroingRoutes := kubernetes.CreateRole("openshift-monitoring", fmt.Sprintf(namespacedResourcedNamePattern, tmpNamespace, "routes"), labels, []rbacv1.PolicyRule{
```
bouskaJ, line 88: I think we can create one role per namespace and group `routes` and `configmaps` together.

controllers/securesign/actions/rbac.go
```diff
@@ -57,41 +58,103 @@ func (i rbacAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Securesi
...
 			Resources: []string{"consoles"},
 			Verbs:     []string{"get", "list"},
 		},
 	})
-	_, err = i.Ensure(ctx, role)
 
-	rb := kubernetes.CreateClusterRoleBinding(fmt.Sprintf(namespacedNamePattern, instance.Namespace), labels, rbacv1.RoleRef{
+	i.Ensure(ctx, clusterRoleOpenshiftConsole)
+
+	rolebindingOpenshiftConsole := kubernetes.CreateRoleBinding("openshift-console", fmt.Sprintf(namespacedNamePattern, tmpNamespace), labels, rbacv1.RoleRef{
```
bouskaJ, line 123: `Console` is non-namespaced resource. I guess, you would need to create `ClusterRoleBinding` to make it working.

bouskaJ (commented):
You can remove resources without ownership in the finalizer section (see https://github.com/securesign/secure-sign-operator/blob/main/controllers/securesign/securesign_controller.go#L97)

JasonPowr:
@Gregory-Pereira Just a thought feel free to ignore if I'm wrong (I probably am), Is there a reason we need to check these config maps to see if telemetry is enabled, Can we just add a field to the API, maybe `securesign.telemetry: bool` can also do a check to see if monitoring is enabled, if it is then make the calls to segment? It would at least remove the calls to `cluster-monitoring-config` and the console resource, for `enable_user_workload_monitoring_if_does_not_exist` it could just be a INFO:msg mentioning that User workload monitoring is not enabled, please enable it, with a link to docs?

controllers/securesign/actions/rbac.go
```diff
@@ -57,41 +58,103 @@ func (i rbacAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Securesi
...
+	rolebindingOpenshiftMontiroingRoutes := kubernetes.CreateRoleBinding("openshift-monitoring", fmt.Sprintf(namespacedResourcedNamePattern, tmpNamespace, "routes"), labels, rbacv1.RoleRef{
+		APIGroup: v1.SchemeGroupVersion.Group,
+		Kind:     "Role",
+		Name:     fmt.Sprintf(namespacedResourcedNamePattern, tmpNamespace, "routes"),
+	},
+		[]rbacv1.Subject{
+			{Kind: "ServiceAccount", Name: SegmentRBACName, Namespace: instance.Namespace},
+		})
+
+	i.Ensure(ctx, rolebindingOpenshiftMontiroingRoutes)
```
cooktheryan, line 107: small nit on spelling

controllers/securesign/actions/rbac.go
Gregory-Pereira: I did try this however when I went to go about testing I found that I couldnt see the the rbac related to routes from `kubectl auth can-i --list`. However this could have been to failing to delete the existing roles, not sure

Gregory-Pereira:
> @Gregory-Pereira Just a thought feel free to ignore if I'm wrong (I probably am), Is there a reason we need to check these config maps to see if telemetry is enabled, Can we just add a field to the API, maybe `securesign.telemetry: bool` can also do a check to see if monitoring is enabled, if it is then make the calls to segment? It would at least remove the calls to `cluster-monitoring-config` and the console resource, for `enable_user_workload_monitoring_if_does_not_exist` it could just be a INFO:msg mentioning that User workload monitoring is not enabled, please enable it, with a link to docs?

I am not sure that would work, these checks to see how / if monitoring were disabled were explicitly shared with me at product management (see jira:  https://issues.redhat.com/browse/CONSOLE-3944 ). I think we are required to check in these pre-established ways to see if monitoring is disabled.

JasonPowr:
> > @Gregory-Pereira Just a thought feel free to ignore if I'm wrong (I probably am), Is there a reason we need to check these config maps to see if telemetry is enabled, Can we just add a field to the API, maybe `securesign.telemetry: bool` can also do a check to see if monitoring is enabled, if it is then make the calls to segment? It would at least remove the calls to `cluster-monitoring-config` and the console resource, for `enable_user_workload_monitoring_if_does_not_exist` it could just be a INFO:msg mentioning that User workload monitoring is not enabled, please enable it, with a link to docs?
> 
> I am not sure that would work, these checks to see how / if monitoring were disabled were explicitly shared with me at product management (see jira: https://issues.redhat.com/browse/CONSOLE-3944 ). I think we are required to check in these pre-established ways to see if monitoring is disabled.

Ahhhhhhh Ok yea, I figured something like this applied, just wanted to make sure, thanks.

kahboom (commented):
/lgtm

config/rbac/role.yaml
```diff
@@ -257,6 +259,7 @@ rules:
   verbs:
   - create
   - delete
+  - deletecollection
```
Gregory-Pereira, line 262: Why are we adding delete collection to the RBAC? AFAIK we dont have reconciliation yet to clean it up if they opt out so these RBAC pieces never get deleted.

JasonPowr: I believe it is needed because we are using r.Client.DeleteAllOf(), to delete a collection of roles and rolebindings, I am going off of this pr made by jan: https://github.com/securesign/secure-sign-operator/pull/292/files#diff-b29242814604426591a2abc523f7ed4feb8bda3e57c9285b6d0422fe7d5d7245


I have added logic to delete the created resources:  https://github.com/securesign/secure-sign-operator/pull/301/files#diff-b29242814604426591a2abc523f7ed4feb8bda3e57c9285b6d0422fe7d5d7245

osmman (commented):
I think that we should use only one method to decide if analytics should be enabled. Right now CanHandle and Handle method uses different method (annotation/spec api). 

I do not know what was the reason to introduce `spec.Analytics` into API, if original purpose was to disable analytics by default than it could be removed and we can depend only on annotation.

controllers/securesign/actions/rbac.go
```diff
@@ -47,87 +54,208 @@ func (i rbacAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Securesi
...
-	if err != nil {
+	openshiftMonitoringSBJRole := kubernetes.CreateRole(
+		OpenshiftMonitoringNS,
+		fmt.Sprintf(namespacedNamePattern, instance.Namespace),
+		labels,
+		[]rbacv1.PolicyRule{
+			{
+				APIGroups: []string{""},
+				Resources: []string{"configmaps"},
+				Verbs:     []string{"get", "list", "patch"},
```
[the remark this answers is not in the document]
JasonPowr: I get where your coming from, Looking at SBJ: https://github.com/securesign/segment-backup-job/blob/main/src/script.py#L89, we do need to patch it to enable user-workload-monitoring if it is disabled, I can Create a whitelist of sort of resource names so it only access specific ones like its done here: https://github.com/securesign/secure-sign-operator/blob/8e7d6939ab89bb8f323862dd7c659bda0d9b1ba7/controllers/securesign/actions/rbac.go#L215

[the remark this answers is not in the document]
osmman: I think that we should not modify any global openshift configuration and keep that responsibility to user. Because user can intentionally disable user workload monitoring and we will override it without notice. If I understand it correctly only daily statistic are impacted by it, new installation messages will work if user workload monitoring is disabled.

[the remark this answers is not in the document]
JasonPowr: I get what you mean that makes sense, If removed we would be able to get installation metrics, but not nightly the job would fail if it was disabled. However I think the usage metrics eg: https://github.com/securesign/segment-backup-job/blob/main/src/script.py#L163

Are required, meaning monitoring needs to be enabled and so does user-work-load-monitoring. This may need to be brought to the group. As I don't have the answer 

@osmman Have you got any ideas on how this could be handled better?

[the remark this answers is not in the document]
JasonPowr: @osmman If it was Documented that having this job enabled, would enable user-work-load-monitoring, do you think that would suffice?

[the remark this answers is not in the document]
JasonPowr: Or maybe, We don't enable it, and just check it and just leave the job fail if disabled? It may be better than overriding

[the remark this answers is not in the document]
osmman: If option is not enabled then let job fails with warning. Solution should be to document this requirement in our documentation as part of monitoring doc as we have it for HELM deployment. 
https://github.com/securesign/sigstore-ocp/blob/main/docs/enable-grafana-monitoring.md?plain=1#L7-L25

JasonPowr:
@osmman I've removed the patch from the config map, will update the image with the proper one on monday, if you want to test you can use this one:  `quay.io/redhat-user-workloads/rhtas-tenant/segment-backup-job/segment-backup-job:on-pr-5df9ba8ea1c4a42cb85ff12afc4102ff83825e17`

config/samples/rhtas_v1alpha1_securesign.yaml
```diff
@@ -5,14 +5,15 @@ metadata:
     app.kubernetes.io/name: securesign-sample
     app.kubernetes.io/instance: securesign-sample
     app.kubernetes.io/part-of: trusted-artifact-signer
+  annotations:
+    rhtas.redhat.com/metrics: "false"
```
osmman, line 9: I would leave metrics enabled by default

Gregory-Pereira: Im not sure we care, you can see what domain name they use segment via the metrics we collect. Also, we can always write a python code block to do this, I wrote one in bash before we converted it over

controllers/securesign/actions/rbac.go
```diff
@@ -42,92 +47,183 @@ func (i rbacAction) CanHandle(_ context.Context, instance *rhtasv1alpha1.Secures
...
-		Name:     SegmentRBACName,
-	},
+	openshiftMonitoringClusterRoleBinding := kubernetes.CreateClusterRoleBinding(
+		fmt.Sprintf(clusterWideNamePattern, instance.Namespace, "clusterMonitoringRoleBinding"),
+		labels,
+		rbacv1.RoleRef{
+			APIGroup: v1.SchemeGroupVersion.Group,
+			Kind:     "ClusterRole",
+			Name:     "cluster-monitoring-view",
+		},
```
Gregory-Pereira, line 149:
```suggestion
```

controllers/securesign/actions/rbac.go
```diff
@@ -42,92 +47,183 @@ func (i rbacAction) CanHandle(_ context.Context, instance *rhtasv1alpha1.Secures
...
 		[]rbacv1.Subject{
 			{Kind: "ServiceAccount", Name: SegmentRBACName, Namespace: instance.Namespace},
 		})
```
Gregory-Pereira, line 152:
```suggestion
```

controllers/securesign/actions/rbac.go
```diff
@@ -42,92 +47,183 @@ func (i rbacAction) CanHandle(_ context.Context, instance *rhtasv1alpha1.Secures
...
+			Message: err.Error(),
+		})
+		meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+			Type:    constants.Ready,
+			Status:  metav1.ConditionFalse,
+			Reason:  constants.Failure,
+			Message: err.Error(),
+		})
+		return i.FailedWithStatusUpdate(ctx, fmt.Errorf("could not create monitoring ClusterRoleBinding for SBJ: %w", err), instance)
+	}
```
Gregory-Pereira, line 167: same as above, not sure what this CR + CRB give us? Pretty sure this worked before without any reference to NSes

controllers/securesign/actions/rbac.go
```diff
@@ -42,92 +47,183 @@ func (i rbacAction) CanHandle(_ context.Context, instance *rhtasv1alpha1.Secures
...
+				APIGroups:     []string{"operator.openshift.io"},
+				Resources:     []string{"consoles"},
+				Verbs:         []string{"get", "list"},
+				ResourceNames: []string{"cluster"},
+			},
+			{
+				APIGroups:     []string{"route.openshift.io"},
+				Resources:     []string{"routes"},
+				Verbs:         []string{"get", "list"},
+				ResourceNames: []string{"console"},
```
Gregory-Pereira, line 183: A bit confused what we have done with this so let me know if I am just wrong but from what I remember ...

These belong in 2 different namespaces. technically the `consoles` should be a clusterrole, but the resource lives in the `openshift-console` ns. As for the route, the only route we call from an API rather than the `requests` module is in the `openshift-monitoring` ns, see [unchanged chunk in your segment PR diff](https://github.com/securesign/segment-backup-job/pull/82/files#diff-adc9da422588504d59d5408ca0c4b49e36e5f82628301848a703612e45e959b2R119)

controllers/securesign/actions/segment_backup_cronjob.go
```diff
@@ -42,11 +42,6 @@ func (i segmentBackupCronJob) CanHandle(_ context.Context, instance *rhtasv1alph
 }
 
 func (i segmentBackupCronJob) Handle(ctx context.Context, instance *rhtasv1alpha1.Securesign) *action.Result {
-
-	if !instance.Spec.Analytics {
```
Gregory-Pereira, line 46: im sure you had a good reason for this, could you tell me what the correct way we handle disabling metrics to look like in the controller, or do we just not need to touch the action at all?

controllers/securesign/actions/rbac.go
```diff
@@ -42,92 +47,183 @@ func (i rbacAction) CanHandle(_ context.Context, instance *rhtasv1alpha1.Secures
...
-		Kind:     "ClusterRole",
-		Name:     SegmentRBACName,
-	},
+	openshiftMonitoringClusterRoleBinding := kubernetes.CreateClusterRoleBinding(
+		fmt.Sprintf(clusterWideNamePattern, instance.Namespace, "clusterMonitoringRoleBinding"),
+		labels,
+		rbacv1.RoleRef{
+			APIGroup: v1.SchemeGroupVersion.Group,
+			Kind:     "ClusterRole",
+			Name:     "cluster-monitoring-view",
```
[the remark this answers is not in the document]
JasonPowr: From the openshift docs: https://docs.openshift.com/container-platform/4.13/observability/monitoring/enabling-monitoring-for-user-defined-projects.html#accessing-metrics-from-outside-cluster_enabling-monitoring-for-user-defined-projects

If you bind the service account to the "cluster-monitoring-view" cluster role, you give it permissions to the Thanos Querier API

controllers/securesign/actions/rbac.go
JasonPowr: As above

controllers/securesign/actions/rbac.go
JasonPowr: As above

controllers/securesign/actions/rbac.go
JasonPowr: You also get the route here: https://github.com/securesign/segment-backup-job/blob/6a5b7cf72b25ebcea470e2e20e58f99780ad9a7f/src/script.py#L116, to get the cluster domain. 

The route for `openshift-monitoring` is still defined here

controllers/securesign/actions/segment_backup_cronjob.go
JasonPowr: So I believe the analytics part of the API was added so we could force disable SBJ because of the secrets issue, Now because thats fixed it can be removed the correct way to disable the SBJ now is through the annotations like so:

```
kind: Securesign
metadata:
  labels:
    app.kubernetes.io/name: securesign-sample
    app.kubernetes.io/instance: securesign-sample
    app.kubernetes.io/part-of: trusted-artifact-signer
  annotations:
    rhtas.redhat.com/metrics: "false"
  name: securesign-sample
```

osmman: https://github.com/securesign/secure-sign-operator/pull/301#pullrequestreview-2008878297

config/samples/rhtas_v1alpha1_securesign.yaml
JasonPowr: I've enabled it by default

osmman:
I tested it and it works without problem for me

JasonPowr:
@osmman Sorry, when you say it worked, are you referring to the pr as a whole?

osmman:
> @osmman Sorry, when you say it worked, are you referring to the pr as a whole? I'm just not sure if you are referring to [#301 (comment)](https://github.com/securesign/secure-sign-operator/pull/301#discussion_r1575041746) or not?

As a whole PR

controllers/securesign/actions/rbac.go
```diff
@@ -42,92 +47,183 @@ func (i rbacAction) CanHandle(_ context.Context, instance *rhtasv1alpha1.Secures
...
 	}
-
-	if err = ctrl.SetControllerReference(instance, sa, i.Client.Scheme()); err != nil {
-		return i.Failed(fmt.Errorf("could not set controll reference for SA: %w", err))
+	if err = controllerutil.SetControllerReference(instance, serviceAccount, i.Client.Scheme()); err != nil {
+		return i.Failed(fmt.Errorf("could not set controller reference for serviceAccount: %w", err))
+	}
+	if _, err = i.Ensure(ctx, serviceAccount); err != nil {
+		meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+			Type:    actions.ServerCondition,
```
osmman, line 67: Please create a new type of status condition and do not reuse [rekor's](https://github.com/securesign/secure-sign-operator/pull/301/files#diff-09d5020e13e6362acb4f2180ea842bb605870333483fd87a67a67d3caf117741R12) `ServerAvalaible`. It could be for example `AnalyticsAvalaible`.

PR #304, opened 2024-04-03
Stabilize e2e

bouskaJ:


bouskaJ:
Getting 
```------------------------------
• [FAILED] [0.384 seconds]
CliServer is running when operator is installed  [It] is up exposed
/home/runner/work/secure-sign-operator/secure-sign-operator/e2e/cli_server_test.go:24

  [FAILED] Client for ec on /clients/windows/%!s(MISSING)-amd64.gz not found
  Expected
      <int>: 404
  to equal
      <int>: 200
  In [It] at: /home/runner/work/secure-sign-operator/secure-sign-operator/e2e/cli_server_test.go:50 @ 04/05/24 09:41:48.[23](https://github.com/securesign/secure-sign-operator/actions/runs/8567623082/job/23479950132#step:12:24)5
------------------------------
•••••••

Summarizing 1 Failure:
  [FAIL] CliServer is running when operator is installed  [It] is up exposed
  /home/runner/work/secure-sign-operator/secure-sign-operator/e2e/cli_server_test.go:50
```
from CI that is expected (see https://issues.redhat.com/browse/SECURESIGN-737)

osmman:
@bouskaJ Please divide this PR into smaller and directly targeted ones. For example, changes for Rekor should be in a separate PR with explanation what it try to fix.

controllers/rekor/actions/pending.go
```diff
@@ -43,11 +43,6 @@ func (i pendingAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Rekor
 		return i.Requeue()
 	}
 
-	meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
```
osmman, line 46: Why are you moving transition of pending state into generate signer action? If you want to change it that it will be better to keep it consistent with other implementation of transition actions like `to_create_phase`, `to_create_initialize`, `to_create_pending`

osmman: after a second look, I understood the reason for moving generate signer action to pending state (return back to original place). I just think that we should extract the state transitions into a separate actions and make it consistent across operator.

PR #333, opened 2024-04-19
Rises a memory limit to 256Mi on manager's container

osmman:
Fixes [SECURESIGN-833](https://issues.redhat.com//browse/SECURESIGN-833)


osmman:
> There are more references, please include: https://github.com/securesign/secure-sign-operator/blob/main/config/default/manager_auth_proxy_patch.yaml#L46-L50

I don't think there is a need to adjust this limit. It is resource limit for `kube-rbac-proxy` container and our problem has been caused in `manager` container

lance (approved):
/lgtm

PR #334, opened 2024-04-23
Add pprof-address option to expose pprof server

osmman:
Add `--pprof-address` option to expose the pprof server. Default is empty string which disables the pprof server.

Fixes unused value ffrom `--metrics-bind-address` option.

/cc @lance @JasonPowr


JasonPowr (approved):
/lgtm

PR #340, opened 2024-04-24
[Securesign 835] | Create Tree Error Looping indefinetley.

tommyd450:
This pr is a WIP as I believe this is the solution and I  also need to update tests still.

Where the error itself originates is within the trillian repo [here](https://github.com/securesign/trillian/blob/303a8bdbe307e5dbc703221cc782395ee72ee986/client/admin.go#L49) there isnt anything wrong on the trillian end but rather ours, that snippet of code will run forever without a specified deadline.
I have added some deadline flags to ctlog and rekor with a default set for 5 minutes.

### DIFF
--- a/controllers/common/create_tree.go
+++ b/controllers/common/create_tree.go
@@ -17,6 +17,7 @@ import (
 
 // reference code https://github.com/sigstore/scaffolding/blob/main/cmd/trillian/createtree/main.go
 func CreateTrillianTree(ctx context.Context, displayName string, trillianURL string) (*trillian.Tree, error) {
+	var err error
 	inContainer, err := kubernetes.ContainerMode()
 	if err == nil {
 		if !inContainer {
@@ -36,7 +37,6 @@ func CreateTrillianTree(ctx context.Context, displayName string, trillianURL str
 	} else {
 		klog.Info("Can't recognise operator mode - expecting in-container run")
 	}
-
 	req, err := newRequest(displayName)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


api/v1alpha1/ctlog_types.go
```diff
@@ -37,6 +37,9 @@ type CTlogSpec struct {
 
 	//Enable Service monitors for ctlog
 	Monitoring MonitoringConfig `json:"monitoring,omitempty"`
+
+	//+kubebuilder:default:= 300000
+	Deadline int64 `json:"deadline,omitempty"`
```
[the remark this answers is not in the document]
tommyd450: @osmman Im not sure what you mean for this? The deadline amount we can specify in flags is seemingly related to operator life cycle?

controllers/common/create_tree.go
```diff
@@ -54,9 +54,11 @@ func CreateTrillianTree(ctx context.Context, displayName string, trillianURL str
 	adminClient := trillian.NewTrillianAdminClient(conn)
 	logClient := trillian.NewTrillianLogClient(conn)
 
-	tree, err := client.CreateAndInitTree(ctx, req, adminClient, logClient)
+	deadline := time.Now().Add(time.Duration(120000) * time.Millisecond)
```
osmman, line 57: It can be simplified to
```suggestion
	deadline := time.Now().Add(2 * time.Minute)
```

controllers/common/create_tree.go
```diff
@@ -54,12 +53,14 @@ func CreateTrillianTree(ctx context.Context, displayName string, trillianURL str
 	adminClient := trillian.NewTrillianAdminClient(conn)
 	logClient := trillian.NewTrillianLogClient(conn)
 
-	tree, err := client.CreateAndInitTree(ctx, req, adminClient, logClient)
-
+	timeout := time.Duration(120 * time.Second)
+	ctx2, cancel := context.WithTimeout(ctx, timeout)
```
osmman, line 57:
```suggestion
	ctx2, cancel := context.WithTimeout(ctx, timeout)
	defer cancel()
```
https://pkg.go.dev/context#WithTimeout

PR #344, opened 2024-04-25
feat: Add image EnVar/Flags 

JasonPowr:
Adds the ability to specify an image when deploying the operator can be done either by flag or env var

### DIFF
--- a/main.go
+++ b/main.go
@@ -25,6 +25,7 @@ import (
 	consolev1 "github.com/openshift/api/console/v1"
 	v1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
+	"github.com/securesign/operator/controllers/common/utils"
 	"github.com/securesign/operator/controllers/common/utils/kubernetes"
 	"github.com/securesign/operator/controllers/constants"
 	"github.com/securesign/operator/controllers/ctlog"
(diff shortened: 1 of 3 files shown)


osmman:
Hi @JasonPowr, I think about more generic solution to manage payload images by operator. We can make images configurable via ENV and manager flags which will give us more flexibility for example to run tests with upstream images or provide hotfixes for components without rebuilding and redistributing of operator. You can find inspiration in [opentelemetry operator](https://github.com/open-telemetry/opentelemetry-operator/blob/main/main.go#L149-L158).

JasonPowr:
> Hi @JasonPowr, I think about more generic solution to manage payload images by operator. We can make images configurable via ENV and manager flags which will give us more flexibility for example to run tests with upstream images or provide hotfixes for components without rebuilding and redistributing of operator. You can find inspiration in [opentelemetry operator](https://github.com/open-telemetry/opentelemetry-operator/blob/main/main.go#L149-L158).

@osmman Ahhh Ok I see what you mean, will work on it

JasonPowr:
@osmman PTAL :)

main.go
```diff
@@ -104,6 +105,20 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	utils.StringFlagOrEnv(&constants.TrillianLogSignerImage, "trillian-log-signer-image", "TRILLIAN-LOG-SIGNER-IMAGE", constants.TrillianLogSignerImage, "The image used for trillian log signer.")
```
osmman, line 108: please use `_` character instead of `-` when separating words in ENV variable name
```suggestion
	utils.StringFlagOrEnv(&constants.TrillianLogSignerImage, "trillian-log-signer-image", "TRILLIAN_LOG_SIGNER_IMAGE", constants.TrillianLogSignerImage, "The image used for trillian log signer.")
```

main.go
```diff
@@ -104,6 +105,20 @@ func main() {
...
+	utils.StringFlagOrEnv(&constants.RekorSearchUiImage, "rekor-search-ui-image", "REKOR_SEARCH_UI_IMAGE", constants.RekorSearchUiImage, "The image used for rekor search ui.")
+	utils.StringFlagOrEnv(&constants.BackfillRedisImage, "backfill-redis-image", "BACKFILL_REDIS_IMAGE", constants.BackfillRedisImage, "The image used for backfill redis.")
+	utils.StringFlagOrEnv(&constants.TufImage, "tuf-image", "TUF_IMAGE", constants.TufImage, "The image used for TUF.")
+	utils.StringFlagOrEnv(&constants.CTLogImage, "ctlog-image", "CTLOG_IMAGE", constants.CTLogImage, "The image used for ctlog.")
+	utils.StringFlagOrEnv(&constants.ClientServerImage_cg, "client-server-cg-image", "CLIENT_SERVER_CG_IMAGE", constants.ClientServerImage_cg, "The image used to serve cosign and gitsign.")
+	utils.StringFlagOrEnv(&constants.ClientServerImage_re, "client-server-re-image", "CLIENT_SERVER_RE_IMAGE", constants.ClientServerImage_re, "The image used to serve rekor-cli and the ec binary.")
+	utils.StringFlagOrEnv(&constants.SegmentBackupImage, "segment-backup-job-image", "SEGMENT_BACKUP_JOB_IMAGE", constants.SegmentBackupImage, "The image used for the segment backup job")
+
```
osmman, line 121: It is missing an option to set these images: ClientServerImage, TrillianNetcatImage

osmman:
It is working well. I tested it by setting custom rekor image via Subscription and operator correctly uses expected image.
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: rhtas-operator-v1-0-1-sub
spec:
  channel: operator-sdk-run-bundle
  config:
    env:
      - name: REKOR_SERVER_IMAGE
        value: 'gcr.io/projectsigstore/rekor-server:v1.3.6'
  installPlanApproval: Manual
  name: rhtas-operator
  source: rhtas-operator-catalog
  sourceNamespace: rhtas
  startingCSV: rhtas-operator.v1.0.1  
```

main.go
JasonPowr: I just didn't include them as they are not images we release, so I thought they shouldn't need to be changed, adding them now :)

PR #346, opened 2024-04-26
Add annotation to pause resource reconciliation

osmman:
Add a new `rhtas.redhat.com/pause-reconciliation` annotation to pause resource reconciliation. Annotation can be used on any Securesign's custom resource and resources which are owned by them like Deploymen, Service, Ingress

Example custom resource with a paused reconciliation condition
```
kind: Tuf
metadata:
  annotations:
    rhtas.redhat.com/pause-reconciliation: "true"
```

### DIFF
--- a/controllers/annotations/annotations.go
+++ b/controllers/annotations/annotations.go
@@ -1,0 +1,6 @@
+package annotations
+
+const (
+	// PausedReconciliation Annotation used to pause resource reconciliation
+	PausedReconciliation = "rhtas.redhat.com/pause-reconciliation"
+)
--- a/controllers/ctlog/ctlog_controller.go
+++ b/controllers/ctlog/ctlog_controller.go
@@ -18,6 +18,8 @@ package ctlog
 
 import (
 	"context"
+	olpredicate "github.com/operator-framework/operator-lib/predicate"
+	"github.com/securesign/operator/controllers/annotations"
 
 	"github.com/securesign/operator/controllers/ctlog/actions"
 	actions2 "github.com/securesign/operator/controllers/fulcio/actions"
@@ -115,6 +117,12 @@ func (r *CTlogReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *CTlogReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Filter out with the pause annotation.
+	pause, err := olpredicate.NewPause(annotations.PausedReconciliation)
+	if err != nil {
+		return err
+	}
+
 	secretPredicate, err := predicate.LabelSelectorPredicate(metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{
 		{
 			Key:      actions2.FulcioCALabel,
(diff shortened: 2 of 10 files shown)


osmman:
target for 1.1.0 release
/hold

controllers/ctlog/ctlog_controller.go
```diff
@@ -126,10 +134,11 @@ func (r *CTlogReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 
 	return ctrl.NewControllerManagedBy(mgr).
+		WithEventFilter(pause).
 		For(&rhtasv1alpha1.CTlog{}).
 		Owns(&v1.Deployment{}).
 		Owns(&v12.Service{}).
-		Watches(&v12.Secret{}, handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, object client.Object) []reconcile.Request {
+		WatchesMetadata(&v12.Secret{}, handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, object client.Object) []reconcile.Request {
```
bouskaJ, line 141: Reading https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/builder#Builder.WatchesMetadata I am not sure about this. We never use ` &metav1.PartialObjectMetadata{}` construct. Can you provide more info, why you are changing this?

controllers/annotations/annotations.go
```diff
@@ -0,0 +1,6 @@
+package annotations
+
+const (
+	// PausedReconciliation Annotation used to pause resource reconciliation
+	PausedReconciliation = "rhtas.redhat.com/pause-reconciliation"
```
bouskaJ, line 5: Can you move https://github.com/securesign/secure-sign-operator/blob/main/controllers/securesign/actions/segment_backup_cronjob.go#L34 to have everything defined on the same place?

osmman: I prefer to do that in new PR because it is lot of changes which are not connected to this  feature.

controllers/ctlog/ctlog_controller.go
osmman: It is to [reduce amount of data](https://issues.redhat.com/browse/SECURESIGN-833) stored in cache memory. We do not need to fetch secret's data and store them in local cache. We only need metadata to make a decision to enqueue new event.

bouskaJ: I did not study the concept deeply but from the documentation
```
You'll need to pass metav1.PartialObjectMetadata to the client when fetching objects in your reconciler, otherwise you'll end up with a duplicate structured or unstructured cache.
```
I would be careful with it - we don't want to have `duplicate structured or unstructured cache` but if you are 100% sure what you are doing, we can merge it.

osmman: I prefer to store small amount of duplicit data of our managed resources rather then pass thought all secrets in cluster through our local cache.

osmman: Simple reproducer it to run debugger on this line and compare what is stored in `o` 
https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/predicate/predicate.go#L409

bouskaJ (approved):

PR #349, opened 2024-04-26
enable-monitoring-of-operator-in-ocp

JasonPowr:
This should enable monitoring of operator metrics, as long as the namespace has the label 
`openshift.io/cluster-monitoring: "true"`


Gregory-Pereira:
This does help contextualize the solution. I appreciate you sharing this because it gives me a place to start from. You are aware of this based on our call but for provenance of the issue, I want to share background in this thread. The issue is that even redhat certified operators are considered `user-workloads` and that we cannot go around telling cluster monitoring to watch user workloads because it has a chance (however small) of breaking things, and this goes against the support policy. It is for this reason we need to sit down with Openshift Monitoring and find a way forward

JasonPowr:
Example:
Screenshot from 2024-04-26 16-16-46

PR #350, opened 2024-04-26
fix(rekor): enable rekor search ui by default

kahboom:
This PR updates the operator to enable the Rekor Search UI by default.

Changes:
- enable default to true
- swap enabled flag to a pointer rather than a boolean, regenerated types
- updates relevant tests to use the updated type
- update tests that compare rekor instance objects to check specific properties instead

NOTE: I'm experiencing some failures locally for CTLog and the following for Rekor, so leaving this as a draft for now:

```
Summarizing 2 Failures:
  [FAIL] Rekor controller Rekor controller test [It] should successfully reconcile a custom resource for Rekor
  /secure-sign-operator/controllers/rekor/rekor_controller_test.go:219
  [FAIL] Rekor hot update test Rekor update test [It] should successfully reconcile a custom resource for Rekor
  /secure-sign-operator/controllers/rekor/rekor_hot_update_test.go:146

Ran 2 of 2 Specs in 133.229 seconds
FAIL! -- 0 Passed | 2 Failed | 0 Pending | 0 Skipped
```

I've run out of time today so I'll continue to look at it next week. cc @lance @osmman @JasonPowr


controllers/common/utils/optional.go
```diff
@@ -1,5 +1,9 @@
 package utils
 
+func IsEnabled(flag *bool) bool {
```
kahboom, line 3: This function is redundant as it does the same as `OptionalBool` but I felt the name wouldn't have been very clear to future devs and I didn't want to rename `OptionalBool`, however, I'm happy to make any suggested changes here!

controllers/rekor/actions/initialize.go
```diff
@@ -24,15 +25,17 @@ func (i initializeAction) Name() string {
...
+	if c == nil {
+		return false
+	}
+	return (c.Reason == constants.Creating || c.Reason == constants.Ready) && utils.IsEnabled(instance.Spec.RekorSearchUI.Enabled)
 }
 
 func (i initializeAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Rekor) *action.Result {
 
-	if (!instance.Spec.RekorSearchUI.Enabled || meta.IsStatusConditionTrue(instance.Status.Conditions, UICondition)) &&
+	if (instance.Spec.RekorSearchUI.Enabled == nil || !*instance.Spec.RekorSearchUI.Enabled || meta.IsStatusConditionTrue(instance.Status.Conditions, UICondition)) &&
```
lance, line 36: Could you use the `IsEnabled()` function here?

JasonPowr (commented):
Some of the e2e tests still need to be updated, for example here: https://github.com/securesign/secure-sign-operator/blob/main/e2e/common_install_test.go#L62

and here: https://github.com/securesign/secure-sign-operator/blob/main/e2e/config_update_test.go#L66

api/v1alpha1/rekor_types_test.go
```diff
@@ -188,7 +188,9 @@ var _ = Describe("Rekor", func() {
 					Expect(k8sClient.Create(context.Background(), &rekorInstance)).To(Succeed())
 					fetched := &Rekor{}
 					Expect(k8sClient.Get(context.Background(), getKey(&rekorInstance), fetched)).To(Succeed())
-					Expect(fetched.Spec).To(Equal(expectedRekorInstance.Spec))
+					Expect(fetched.Spec.Pvc.Name).To(Equal(expectedRekorInstance.Spec.Pvc.Name))
+					Expect(fetched.Spec.Pvc.Size).To(Equal(expectedRekorInstance.Spec.Pvc.Size))
+					Expect(fetched.Spec.SearchUI.Enabled).To(BeTrue())
```
osmman, line 193:
```suggestion
					Expect(fetched.Spec.RekorSearchUI.Enabled).To(BeTrue())
```

api/v1alpha1/rekor_types_test.go
```diff
@@ -270,7 +272,9 @@ var _ = Describe("Rekor", func() {
 					Expect(k8sClient.Create(context.Background(), &rekorInstance)).To(Succeed())
 					fetchedRekor := &Rekor{}
 					Expect(k8sClient.Get(context.Background(), getKey(&rekorInstance), fetchedRekor)).To(Succeed())
-					Expect(fetchedRekor.Spec).To(Equal(expectedRekorInstance.Spec))
+					Expect(fetchedRekor.Spec.Pvc.Name).To(Equal(expectedRekorInstance.Spec.Pvc.Name))
+					Expect(fetchedRekor.Spec.Pvc.Size).To(Equal(expectedRekorInstance.Spec.Pvc.Size))
+					Expect(fetchedRekor.Spec.SearchUI.Enabled).To(BeTrue())
```
osmman, line 277:
```suggestion
					Expect(fetchedRekor.Spec.RekorSearchUI.Enabled).To(BeTrue())
```

api/v1alpha1/rekor_types_test.go
```diff
@@ -188,7 +188,9 @@ var _ = Describe("Rekor", func() {
 					Expect(k8sClient.Create(context.Background(), &rekorInstance)).To(Succeed())
 					fetched := &Rekor{}
 					Expect(k8sClient.Get(context.Background(), getKey(&rekorInstance), fetched)).To(Succeed())
-					Expect(fetched.Spec).To(Equal(expectedRekorInstance.Spec))
+					Expect(fetched.Spec.Pvc.Name).To(Equal(expectedRekorInstance.Spec.Pvc.Name))
+					Expect(fetched.Spec.Pvc.Size).To(Equal(expectedRekorInstance.Spec.Pvc.Size))
+					Expect(fetched.Spec.RekorSearchUI.Enabled).To(BeTrue())
```
lance, line 193: Here you need to dereference the boolean pointer. My mistake in the earlier suggestion.
```suggestion
					Expect(*fetched.Spec.RekorSearchUI.Enabled).To(BeTrue())
```

api/v1alpha1/rekor_types_test.go
```diff
@@ -270,7 +272,9 @@ var _ = Describe("Rekor", func() {
 					Expect(k8sClient.Create(context.Background(), &rekorInstance)).To(Succeed())
 					fetchedRekor := &Rekor{}
 					Expect(k8sClient.Get(context.Background(), getKey(&rekorInstance), fetchedRekor)).To(Succeed())
-					Expect(fetchedRekor.Spec).To(Equal(expectedRekorInstance.Spec))
+					Expect(fetchedRekor.Spec.Pvc.Name).To(Equal(expectedRekorInstance.Spec.Pvc.Name))
+					Expect(fetchedRekor.Spec.Pvc.Size).To(Equal(expectedRekorInstance.Spec.Pvc.Size))
+					Expect(fetchedRekor.Spec.RekorSearchUI.Enabled).To(BeTrue())
```
lance, line 277: Same here - again my earlier example omitted the deferencing.
```suggestion
					Expect(*fetchedRekor.Spec.RekorSearchUI.Enabled).To(BeTrue())
```

api/v1alpha1/rekor_types_test.go
JasonPowr, line 193: I don't think these need to be here, similar to my previous comments the check here is to match a rekor object to a rekor object.

Best to revert these changes

api/v1alpha1/rekor_types_test.go
JasonPowr, line 277: As above

controllers/rekor/actions/initialize.go
JasonPowr, line 36:
```suggestion
if (!utils.IsEnabled(instance.Spec.RekorSearchUI.Enabled) || meta.IsStatusConditionTrue(instance.Status.Conditions, UICondition)) &&
```

api/v1alpha1/rekor_types_test.go
lance: But isn't the check meant to ensure that the fetched data is the same as what is expected? If that's the case, you can't compare `fetched` to `expectedRekorInstance` directly because the boolean pointer. You have to compare each field, to determine equivalence don't you?

JasonPowr: Ahhh Ok apologies

kahboom: Thanks for the feedback. Yep @lance this was one of the goals was to avoid comparing directly (even if you could tbh, I prefer not to as it makes tests more fragile, where it makes sense of course).

PR #351, opened 2024-04-26
Update Rekor controller behavior to preserve generated keys

osmman:
Update Rekor's behavior of managing generated secrets with keys.

### Signer KMS: secret type
The generated signer key will now be persisted in the following scenarios:
- When the Rekor object is deleted.
- During key rotation with a new signer key.

### Signer Public Key: autodiscovery
A secret containing the public key will be generated from Rekor's API endpoint /api/v1/log/publicKey. This update is necessary because Rekor can be configured to use a signer key from a Cloud KMS, which means the operator might not have access to the key and thus cannot create a public key from it. The generated public key secret will be persistent and will not be automatically deleted by the operator.

The operator will ensure that only the newest secret, named `rekor-public-<name>-<id>`, contains the rhtas.redhat.com/rekor.pub label to maintain the autodiscovery feature for other components.

Related to: [SECURESIGN-568](https://issues.redhat.com/browse/SECURESIGN-568)


internal/controller/rekor/actions/server/createTree.go
```diff
@@ -69,7 +69,7 @@ func (i createTrillianTreeAction) Handle(ctx context.Context, instance *rhtasv1a
 		})
 		return i.FailedWithStatusUpdate(ctx, fmt.Errorf("could not create trillian tree: %w", err), instance)
 	}
-	i.Recorder.Event(instance, v1.EventTypeNormal, "TreeID", "New Trillian tree created")
+	i.Recorder.Eventf(instance, v1.EventTypeNormal, "TrillianTreeCreated", "New Trillian tree created: %d", tree.TreeId)
```
bouskaJ, line 72: Unify with https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/ctlog/actions/createTree.go#L62

bouskaJ (commented):
Looks good. It would be nice to unify modified actions with other components. I suppose you want to open new PR for each one. If it is that case, we can merge it.

internal/controller/rekor/actions/server/createTree.go
osmman: I created new PR it https://github.com/securesign/secure-sign-operator/pull/506

PR #358, opened 2024-05-03
Securesign 861

bouskaJ:


.github/workflows/upgrade.yml
```diff
@@ -0,0 +1,190 @@
...
+          config: ./ci/config.yaml
+
+      - name: Configure cluster
+        run: |
+          kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml
+          kubectl wait --namespace ingress-nginx --for=condition=ready pod --selector=app.kubernetes.io/component=controller --timeout=90s
+
+          #install OLM
+          kubectl create -f https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.25.0/crds.yaml
+          # wait for a while to be sure CRDs are installed
```
lance, line 141: Wouldn't it be better to use `kubectl wait` or something here?

.github/workflows/upgrade.yml
```diff
@@ -0,0 +1,190 @@
...
+          sleep 1
+          kubectl create -f https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.25.0/olm.yaml
+
+          kubectl create --kustomize https://github.com/securesign/sigstore-ocp/keycloak/operator/overlay/kind
+          until [ ! -z "$(kubectl get pod -l name=keycloak-operator -n keycloak-system 2>/dev/null)" ]
+          do
+            echo "Waiting for keycloak operator. Pods in keycloak-system namespace:"
+            kubectl get pods -n keycloak-system
+            sleep 10
```
lance, line 150: Same here and below on line 156 - why not use `kubectl wait`?

.github/workflows/upgrade.yml
bouskaJ: no, `kubectl wait` is not able to wait for resources that are not yet created. We do not need to wait for specific event from the cluster, the sleep is there just to wait for cluster to process the command.

.github/workflows/upgrade.yml
bouskaJ: `kubectl wait` is not able to check non-existing resources so it will fail immediately.

.github/workflows/upgrade.yml
```diff
@@ -0,0 +1,216 @@
...
+
+env:
+  GO_VERSION: 1.21
+
+jobs:
+  upgrade:
+    name: Upgrade operator test
+    runs-on: ubuntu-20.04
+    env:
+      IMG: "ttl.sh/securesign/secure-sign-operator:1h"
```
osmman, line 18: we should add some salt to image name, because there can be collision if the workflow will be triggered multiple time at same time

e2e/upgrade_test.go
```diff
@@ -0,0 +1,361 @@
...
+		rfulcio                                *tasv1alpha.Fulcio
+		rrekor                                 *tasv1alpha.Rekor
+		rtuf                                   *tasv1alpha.Tuf
+		oidcToken                              string
+	)
+
+	AfterEach(func() {
+		if CurrentSpecReport().Failed() {
+			if val, present := os.LookupEnv("CI"); present && val == "true" {
+				if val, present := os.LookupEnv("CI"); present && val == "true" {
```
osmman, line 60: there is duplicate if statement

PR #359, opened 2024-05-06
olm(channel): change from rhtas-1-0 to stable-1-0

lance:
This commit modifies the channels we advertise for the operator, removing the `rhtas-` prefix in favor of a `stable-` prefix, indicating the major/minor version for that particular stable line. This channel naming is based on the olm documentation best practices found at: https://olm.operatorframework.io/docs/best-practices/channel-naming/#example-3-recommended-option-is-the-most-common-scenarios

I did not include a `fast-1-0` or `candidate-1-0`, however the current configuration does not prevent these channels from being used in the future. At this point, I don't think we have a strong enough cadence to be using `fast` and `candidate`, however we may want to consider that for the July release, or at least the EOY release.

If we want to follow the recommended semver naming conventions, then we should consider the July release to be v1.1.0, given that it _should_ introduce new functionality with SECURESIGN-720 and possibly others Jiras as well (key rotation, independently deployable services, etc).

If this PR is merged, the user would see only two channels when installing `stable` and `stable-1-0`. They would (for now) contain the exact same versions. But ultimately, as newer minor versions are released, they will _all_ appear in `stable`, but `stable-1-0` will only contain 1.0.0, 1.0.1, and any other 1.0.* releases we produce.

Do not merge this commit without discussion. I think we really want to get it right and not have to make further changes.


bundle/manifests/rhtas-operator.clusterserviceversion.yaml
```diff
@@ -193,7 +193,7 @@ metadata:
     features.operators.openshift.io/token-auth-aws: "false"
     features.operators.openshift.io/token-auth-azure: "false"
     features.operators.openshift.io/token-auth-gcp: "false"
-    operators.operatorframework.io/builder: operator-sdk-v1.32.0
+    operators.operatorframework.io/builder: operator-sdk-v1.34.1
```
Gregory-Pereira, line 196: Since we are bumping this dont we also need to bump the [operator sdk verison in the makefile](https://github.com/securesign/secure-sign-operator/pull/359/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R56)

Gregory-Pereira (commented):
/lgtm just 1 question

bundle/manifests/rhtas-operator.clusterserviceversion.yaml
lance: Probably - I just ran `make bundle` 🤷

lance: I've updated the makefile and also noticed an old `alpha` annotation hanging around that I changed to `stable`

Gregory-Pereira (approved):
/lgtm

JasonPowr (approved):
/lgtm

The CI is currently failing because this is opened of a fork, and not a direct branch, just FYI  

https://github.com/securesign/secure-sign-operator/actions/runs/8974908261/job/24648503582?pr=359#step:4:1

bundle/metadata/annotations.yaml
```diff
@@ -4,7 +4,7 @@ annotations:
   operators.operatorframework.io.bundle.manifests.v1: manifests/
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: rhtas-operator
-  operators.operatorframework.io.bundle.channels.v1: alpha
+  operators.operatorframework.io.bundle.channels.v1: stable
   operators.operatorframework.io.metrics.builder: operator-sdk-v1.32.0
```
osmman, line 8:
```suggestion
  operators.operatorframework.io.metrics.builder: operator-sdk-v1.34.1
```

osmman (changes requested):
I agree with new format of channel naming.

I found few places with old version v1.32.0 of operator sdk and alpha channel. Can you please to update them too.

- https://github.com/lance/secure-sign-operator/blob/channel-conventions/bundle.Dockerfile#L8-L9
- https://github.com/securesign/secure-sign-operator/blob/5a523fd3c5351a9e625fab3b8008a29249ea5d65/bundle/metadata/annotations.yaml#L8
- https://github.com/securesign/secure-sign-operator/blob/5a523fd3c5351a9e625fab3b8008a29249ea5d65/fbc/generate-fbc.sh#L75-L76
- https://github.com/search?q=repo%3Asecuresign%2Fsecure-sign-operator+%22quay.io%2Foperator-framework%2Fscorecard-test%3Av1.32.0%22&type=code
- https://github.com/securesign/secure-sign-operator/blob/5a523fd3c5351a9e625fab3b8008a29249ea5d65/bundle/manifests/rhtas-operator.clusterserviceversion.yaml#L879
- https://github.com/securesign/secure-sign-operator/blob/5a523fd3c5351a9e625fab3b8008a29249ea5d65/config/manifests/bases/rhtas-operator.clusterserviceversion.yaml#L104

fbc/generate-fbc.sh
```diff
@@ -73,7 +73,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=rhtas-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
```
osmman, line 75:
```suggestion
LABEL operators.operatorframework.io.bundle.channels.v1=stable
```

lance:
> /lgtm
> 
> The CI is currently failing because this is opened of a fork, and not a direct branch, just FYI
> 
> https://github.com/securesign/secure-sign-operator/actions/runs/8974908261/job/24648503582?pr=359#step:4:1

Always forget about that...

fbc/generate-fbc.sh
```diff
@@ -147,14 +147,14 @@ case $cmd in
       exit 1
     fi
     setBrew "${frag}" "$3"
-    "${OPM_CMD}" alpha render-template basic "${frag}"/graph.yaml > "${frag}"/catalog/rhtas-operator/catalog.json
+    "${OPM_CMD}" stable render-template basic "${frag}"/graph.yaml > "${frag}"/catalog/rhtas-operator/catalog.json
```
JasonPowr, line 150: This change is wrong, alpha in this case refers to experimental features for opm.

fbc/generate-fbc.sh
```diff
@@ -147,14 +147,14 @@ case $cmd in
...
     unsetBrew "${frag}" "$3"
   ;;
   "--render-all")
     for f in ./"v4."*; do
       frag=${f#./}
       setBrew "${frag}" "$2"
-      "${OPM_CMD}" alpha render-template basic "${frag}"/graph.yaml > "${frag}"/catalog/rhtas-operator/catalog.json
+      "${OPM_CMD}" stable render-template basic "${frag}"/graph.yaml > "${frag}"/catalog/rhtas-operator/catalog.json
```
JasonPowr, line 157: Same as above

JasonPowr (commented):
Seems fine, just too things I noticed

JasonPowr (approved):
/lgtm

PR #361, opened 2024-05-07
chore(bundle): make bundle with new image

lance:
Adds latest operator snapshot sha to bundle


lance:
Build looks good - self-merging

PR #365, opened 2024-05-10
fix: Set the common name of the Fulcio cert automatically 

JasonPowr:
Jira: https://issues.redhat.com/browse/SECURESIGN-851

If commonName is empty in the CRD definition, One will be calculated based on the ns and name of service

### DIFF
--- a/api/v1alpha1/fulcio_types.go
+++ b/api/v1alpha1/fulcio_types.go
@@ -24,7 +24,6 @@ type FulcioSpec struct {
 }
 
 // FulcioCert defines fields for system-generated certificate
-// +kubebuilder:validation:XValidation:rule=(has(self.caRef) || self.commonName != ""),message=commonName cannot be empty
 // +kubebuilder:validation:XValidation:rule=(has(self.caRef) || self.organizationName != ""),message=organizationName cannot be empty
 // +kubebuilder:validation:XValidation:rule=(!has(self.caRef) || has(self.privateKeyRef)),message=privateKeyRef cannot be empty
 type FulcioCert struct {
@@ -40,6 +39,8 @@ type FulcioCert struct {
 	CARef *SecretKeySelector `json:"caRef,omitempty"`
 
 	//+optional
+	// CommonName specifies the common name for the Fulcio certificate.
+	// If not provided, the common name will default to the host name.
 	CommonName string `json:"commonName,omitempty"`
 	//+optional
 	OrganizationName string `json:"organizationName,omitempty"`
--- a/controllers/fulcio/utils/fulcio_secrets.go
+++ b/controllers/fulcio/utils/fulcio_secrets.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"bytes"
+	"context"
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/rand"
@@ -14,6 +15,8 @@ import (
 	"time"
 
 	rhtasv1alpha1 "github.com/securesign/operator/api/v1alpha1"
+	"github.com/securesign/operator/controllers/common/utils/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type FulcioCertConfig struct {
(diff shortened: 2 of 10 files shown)


controllers/fulcio/utils/fulcio_secrets.go
```diff
@@ -161,3 +165,16 @@ func GenerateSerialNumber() (*big.Int, error) {
 	}
 	return serial, nil
 }
+
+func getCommonName(client client.Client, instance *rhtasv1alpha1.Fulcio, deploymentName string) string {
+	ctx := context.Background()
```
bouskaJ, line 170: I believe, we should pass `context` from the root function rather than create new one. With new context the application can't cancel operation on graceful shutdown.

api/v1alpha1/fulcio_types.go
```diff
@@ -39,8 +38,6 @@ type FulcioCert struct {
 	//+optional
 	CARef *SecretKeySelector `json:"caRef,omitempty"`
 
-	//+optional
-	CommonName string `json:"commonName,omitempty"`
```
osmman, line 43: Do not remove this option. It will broke compatibility with old 1.0.x channel and it will remove feature to customize common name. 

We cannot fully depends on host from Ingress because the users can have their own specific configuration and they need to customize it.

osmman (changes requested):

controllers/fulcio/utils/fulcio_secrets.go
```diff
@@ -161,3 +165,16 @@ func GenerateSerialNumber() (*big.Int, error) {
...
+	if instance.Spec.ExternalAccess.Enabled {
+		ingress := &v12.Ingress{}
+		err := client.Get(ctx, types.NamespacedName{Name: deploymentName, Namespace: instance.Namespace}, ingress)
```
osmman, line 173: Ingress object do not exists in phase which this action is executed.
handle cert action is executed during  `Pending` phase, Ingress action is executed during `Creating` phase

api/v1alpha1/fulcio_types.go
JasonPowr:
```
Do not remove this option. It will broke compatibility with old 1.0.x channel and it will remove feature to customize common name.
```

Thanks for this, something I always forget about need to make a note for myself of it for future work

controllers/fulcio/utils/fulcio_secrets.go
JasonPowr: I have moved the ingress & service creation to the pending phase to facilitate this: https://github.com/securesign/secure-sign-operator/pull/365/files#diff-444aef5e7f264fafa6b15d801f33ad3982d46a27f77badc3ff210a84c137aeeaR89-R90

osmman: It is not good to move it from creating phase. Rather then dig information from ingress object you can reuse function which calculate host name

controllers/fulcio/utils/fulcio_secrets.go
```diff
@@ -103,6 +106,16 @@ func CreateFulcioCA(config *FulcioCertConfig, instance *rhtasv1alpha1.Fulcio) ([
 	notBefore := time.Now()
 	notAfter := notBefore.Add(365 * 24 * 10 * time.Hour)
 
+	if instance.Spec.Certificate.CommonName == "" {
+		if instance.Spec.ExternalAccess.Enabled {
+			if instance.Spec.Certificate.CommonName, err = kubernetes.CalculateHostname(ctx, client, deploymentName, instance.Namespace); err != nil {
```
osmman, line 111: `instance.Spec.ExternalAccess.Host` should be used if it is set

api/v1alpha1/fulcio_types.go
```diff
@@ -24,7 +24,6 @@ type FulcioSpec struct {
 }
 
 // FulcioCert defines fields for system-generated certificate
-// +kubebuilder:validation:XValidation:rule=(has(self.caRef) || self.commonName != ""),message=commonName cannot be empty
```
osmman, line 27: We should extend doc to inform that `commonName` will be calculated based on host name if it is not set

osmman:
PR targeted for 1.1.0 release
/hold

controllers/fulcio/utils/fulcio_secrets.go
JasonPowr: If we do this, we are going to have to make `instance.Spec.ExternalAccess.Host` required by default to make sure a value gets set, is this ok?

At this stage `instance.Spec.ExternalAccess.Host` is equal to whatever is defined in the crd definition as it doesn't get set until later in the loop.

osmman:
```
if CommonName empty {
   if ExternalAccess.Enable {
     if ExternalAccess.Host != "" {
        return ExternalAccess.Host
     } else {
        return CalculateHostname()
     }
   } else {
     return "fulcio.svc.local"
   }
}
```

osmman (approved):

JasonPowr:
Fix for upgrade test: https://github.com/securesign/secure-sign-operator/pull/389

PR #372, opened 2024-05-15
Remove analytics field from Securesign sample

osmman:


JasonPowr:
Found one more example to it here: https://github.com/securesign/secure-sign-operator/blob/main/README.md?plain=1#L136

Only the readme but should probably be removed :)

osmman:
> Found one more example to it here: https://github.com/securesign/secure-sign-operator/blob/main/README.md?plain=1#L136
> 
> Only the readme but should probably be removed :)

README fixed

PR #373, opened 2024-05-15
update-images

JasonPowr:


lance (approved):
Haven't double checked the shas but lgtm

PR #378, opened 2024-05-16
Disable kind registry to free node memory

bouskaJ:


Gregory-Pereira (approved):
Can confirm root cause of the issue is that it lacks fs storage space, this should help address that.

PR #390, opened 2024-05-27
[SECURESIGN-346] TAS acceptance tests

petrpinkas:


bouskaJ:
@petrpinkas  I am thinking about the placement of this test suite. I would prefer to move it outside the operator repository and create new one. My reasons:
- **dependency management** - we may need various dependencies not related to the operator. There is no `test` scope in Go so all dependencies would be compiled into the operator. It could bring problems with CVEs and so...
- the acceptance test suite will be tightly coupled with RH build system and RH specific tests. This repository could be provided upstream eventually, so I would prefer to have this RH specific tests in different repository.

@osmman, @JasonPowr WDYT?

petrpinkas:
I agree @bouskaJ. I was also thinking about new repository, because neither operator neither e2e tests seems like a right place. It got me started. If there are no other suggestions, I'll move it to separate repository tomorrow.

osmman:
Closing PR because it's content will be moved to different repository.

PR #393, opened 2024-05-29
Securesign-1001 | Create Tree Deadline not being enough

tommyd450:
Ive added configurable fields for the deadline amount. And increased the default time for the createTree job to run across 20 minutes as opposed to 2 minutes. I would appreiate if someone could run this on a cluster.

### DIFF
--- a/api/v1alpha1/ctlog_types.go
+++ b/api/v1alpha1/ctlog_types.go
@@ -37,6 +37,9 @@ type CTlogSpec struct {
 
 	//Enable Service monitors for ctlog
 	Monitoring MonitoringConfig `json:"monitoring,omitempty"`
+
+	//+kubebuilder:default:=1200
+	CreateTreeDeadline int64 `json:"createTreeDeadline,omitempty"`
 }
 
 // CTlogStatus defines the observed state of CTlog component
--- a/api/v1alpha1/ctlog_types_test.go
+++ b/api/v1alpha1/ctlog_types_test.go
@@ -86,6 +86,7 @@ var _ = Describe("CTlog", func() {
 							Namespace: "default",
 						},
 					}
+					expectedCTlogInstance.Spec.CreateTreeDeadline = 1200
 
 					Expect(k8sClient.Create(context.Background(), &ctlogInstance)).To(Succeed())
 					fetched := &CTlog{}
@@ -97,7 +98,6 @@ var _ = Describe("CTlog", func() {
 			When("CR is fully populated", func() {
 				It("outputs the CR", func() {
 					tree := int64(1269875)
-
 					ctlogInstance = CTlog{
 						ObjectMeta: metav1.ObjectMeta{
 							Name:      "ctlog-full-manifest",
@@ -145,18 +145,20 @@ var _ = Describe("CTlog", func() {
 
 				It("sets spec.pvc.storage if spec.pvc is partially set", func() {
 					tree := int64(1269875)
+					deadline := int64(1200)
 					ctlogInstance = CTlog{
 						ObjectMeta: metav1.ObjectMeta{
 							Name:      "ctlog-storage",
 							Namespace: "default",
 						},
 						Spec: CTlogSpec{
-							TreeID: &tree,
+							TreeID:             &tree,
+							CreateTreeDeadline: deadline,
 						},
 					}
 
 					expectedCTlogInstance.Spec.TreeID = &tree
-
+					expectedCTlogInstance.Spec.CreateTreeDeadline = deadline
 					Expect(k8sClient.Create(context.Background(), &ctlogInstance)).To(Succeed())
 					fetchedCTlog := &CTlog{}
 					Expect(k8sClient.Get(context.Background(), getKey(&ctlogInstance), fetchedCTlog)).To(Succeed())
--- a/api/v1alpha1/rekor_types.go
+++ b/api/v1alpha1/rekor_types.go
@@ -28,6 +28,8 @@ type RekorSpec struct {
 	// BackFillRedis CronJob Configuration
 	//+kubebuilder:default:={enabled: true, schedule: "0 0 * * *"}
 	BackFillRedis BackFillRedis `json:"backFillRedis,omitempty"`
+	//+kubebuilder:default:=1200
+	CreateTreeDeadline int64 `json:"createTreeDeadline,omitempty"`
 }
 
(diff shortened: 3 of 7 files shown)


tommyd450:
Failing to build will investigate if the same holds locally.

osmman (changes requested):
Hi, I suggest moving the deadline time setting option from the [CRDs](https://github.com/securesign/secure-sign-operator/pull/393/files#diff-8846307f2f2b6df3b1e115ff1aa3e6fb3eabbc1d81140dad1a92893b2f6c862eR42) to the [manager's flags](https://github.com/securesign/secure-sign-operator/blob/main/main.go#L102-L122). This adjustment is beneficial because the "create tree deadline" setting influences the behavior of manager operations and is not related to the definition of deployed services.

The new flag can use `create-tree-deadline` as a name.

osmman:
> @osmman I believe I have done what you ask however I am unsure about how I am [accessing the flag](https://github.com/securesign/secure-sign-operator/blob/c808037feb77d16b511ef240795b1507db36f823/controllers/rekor/actions/server/createTree.go#L47C78-L47C147) is ideal is there a better way to do so?

Maybe it will be better to make [createTreeDeadline](https://github.com/securesign/secure-sign-operator/pull/393/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R101) global variable and accessing it in [actions](https://github.com/securesign/secure-sign-operator/pull/393/files#diff-1fcd0d3324b98abc40b3cd145e5144ff82c7dd3ca58de0a9602f61e96bb72403R52) simply by:
```
common.CreateTrillianTree(..., main.CreateTreeDeadline)
```

main.go
```diff
@@ -98,13 +98,15 @@ func main() {
...
 		enableLeaderElection bool
+		createTreeDeadline   int64
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&pprofAddr, "pprof-address", "", "The address to expose the pprof server. Default is empty string which disables the pprof server.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.Int64Var(&createTreeDeadline, "create-tree-deadline", 1200, "The time allowance for the create tree job to run before failing.")
```
simonbaird, line 109: Perhaps mention that the value is seconds to make that clear.

Something like this:
```suggestion
	flag.Int64Var(&createTreeDeadline, "create-tree-deadline", 1200, "The time allowance (in seconds) for the create tree job to run before failing.")
```

controllers/constants/config.go
```diff
@@ -0,0 +1,7 @@
+package constants
+
+import "flag"
+
+var (
+	CreateTreeDeadline = flag.Lookup("create-tree-deadline").Value.(flag.Getter).Get().(int64)
```
osmman, line 6: You do not need to use Lookup function of `flag` package. The value will be automatically updated by `flag` when it is parsed.
```suggestion
	CreateTreeDeadline int64 = 1200
```

main.go
```diff
@@ -98,13 +98,15 @@ func main() {
 		probeAddr            string
 		pprofAddr            string
 		enableLeaderElection bool
+		createTreeDeadline   int64
```
osmman, line 102:
```suggestion
```

controllers/constants/config.go
JasonPowr, line 6: You should set a default value here

main.go
```diff
@@ -98,13 +99,15 @@ func main() {
...
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&pprofAddr, "pprof-address", "", "The address to expose the pprof server. Default is empty string which disables the pprof server.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.Int64Var(&createTreeDeadline, "create-tree-deadline", 1200, "The time allowance (in seconds) for the create tree job to run before failing.")
```
osmman, line 110:
```suggestion
	flag.Int64Var(&constants.CreateTreeDeadline, "create-tree-deadline", constants.CreateTreeDeadline, "The time allowance (in seconds) for the create tree job to run before failing.")
```

main.go
```diff
@@ -127,6 +130,8 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+
+	CreateTreeDeadline = createTreeDeadline
```
JasonPowr, line 134:
```suggestion
```

main.go
```diff
@@ -98,13 +99,15 @@ func main() {
 		probeAddr            string
 		pprofAddr            string
 		enableLeaderElection bool
+		createTreeDeadline   int64
```
JasonPowr, line 102:
```suggestion
```

main.go
JasonPowr, line 110:
```suggestion
	flag.Int64Var(&config.CreateTreeeDeadline, "create-tree-deadline", 1200, "The time allowance (in seconds) for the create tree job to run before failing.")
```

similar to how its done for the images

JasonPowr (commented):
Just some small things I noticed :)

main.go
osmman, line 134:
```suggestion
```

main.go
```diff
@@ -74,6 +74,7 @@ const (
 var (
 	scheme   = runtime.NewScheme()
 	setupLog = ctrl.Log.WithName("setup")
+	CreateTreeDeadline int64
```
osmman, line 77:
```suggestion
```

controllers/constants/config.go
osmman: This will removed coupling to `flag` package in shared libraries whil will make easier development of unit test.

osmman (approved):

osmman (approved):

PR #414, opened 2024-06-04
feat: provide host name option for CLI and search ui

JasonPowr:
jira: https://issues.redhat.com/browse/SECURESIGN-665

### DIFF
--- a/api/v1alpha1/rekor_types.go
+++ b/api/v1alpha1/rekor_types.go
@@ -51,6 +51,8 @@ type RekorSearchUI struct {
 	//+kubebuilder:validation:XValidation:rule=(self || !oldSelf),message=Feature cannot be disabled
 	//+kubebuilder:default:=true
 	Enabled *bool `json:"enabled"`
+	// Set hostname for your Ingress/Route.
+	Host string `json:"host,omitempty"`
 }
 
 type BackFillRedis struct {
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,17 +1,14 @@
-name: Validate SecureSign 
+name: Validate SecureSign
 on:
   workflow_dispatch:
   push:
-    branches: ["main", "release*"]
-    tags: ["*"]
+    branches: [ "main", "release*" ]
+    tags: [ "*" ]
   pull_request:
-    branches: ["main", "release*"]
+    branches: [ "main", "release*" ]
 
 env:
   GO_VERSION: 1.21
-  AWS_REGION: us-east-2
-  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
   IMG: ttl.sh/securesign/secure-sign-operator-${{github.run_number}}:1h
   BUNDLE_IMG: ttl.sh/securesign/bundle-secure-sign-${{github.run_number}}:1h
   CATALOG_IMG: ttl.sh/securesign/catalog-${{github.run_number}}:1h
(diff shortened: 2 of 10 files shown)


api/v1alpha1/rekor_types.go
```diff
@@ -48,6 +48,8 @@ type RekorSearchUI struct {
 	//+kubebuilder:validation:XValidation:rule=(self || !oldSelf),message=Feature cannot be disabled
 	//+kubebuilder:default:=true
 	Enabled *bool `json:"enabled"`
+	// Set hostname for your Ingress/Route.
+	Host string `json:"host,omitempty"`
```
osmman, line 52: I think it would be better to use the `ExternalAccess` strut to maintain consistency across APIs. It will add option to enable/dissable creation of Ingress.

JasonPowr: Do you think changing the API like this will break Automatic upgrades? That's the main reason I chose to extend the API vs changing it

osmman: I think that upgraded stack will survive but it will end up in different state. These scenarios came to my mind:

### Rekor with disabled external access and with enabled UI
Initial state of CRD
```
spec:
  externalAccess:
    enabled: false
  rekorSearchUI:
    enabled: true  
```

Ingress is created only for **RekorSearchUI**. This can be bug because [ui.ingressAction](https://github.com/securesign/secure-sign-operator/blob/main/controllers/rekor/actions/ui/ingress.go#L37-L39) do not test if Rekor has enabled **extenalAccess** and it will create ingress object every time when UI is enabled.

Existing Ingress objects: RekorSearchUI

After upgrade:
```
spec:
  externalAccess:
    enabled: false
  rekorSearchUI:
    enabled: true
    externalAccess:
      enabled: false  
```

Existing Ingress objects: RekorSearchUI

It ended up in an unwanted state, where the definition in the CDR says that the ingress object should not exist but the reality is different because the Operator does not currently include functionality to delete the object.
 
### Rekor with enabled external access and enabled UI
Initial state of CRD
```
spec:
  externalAccess:
    enabled: true
  rekorSearchUI:
    enabled: true  
```

Existing Ingress objects: Rekor, RekorSearchUI

After upgrade:
```
spec:
  externalAccess:
    enabled: false
  rekorSearchUI:
    enabled: true
    externalAccess:
      enabled: false  
```

Existing Ingress objects: Rekor, RekorSearchUI

In this case, a default value for externalAccess (false) breaks the integrity of the real state of ingress objects and what is defined in the CRD.

JasonPowr: @osmman Seem's like a tricky one alright, What do you think is the best way forward, for me although the API would not be uniform it seems like my solution causes less issues? I would prefer to have it uniform, but in this case it may not work.

osmman: In order to solve the problems for the unified solution, we will need to implement additional logic during the upgrade to correctly set values and reconcile objects. I suppose it would be some extra upgrade actions which will be executed before main reconciliation loop.

osmman: @bouskaJ @lance What is your opinion which way we should go?

bouskaJ: We should avoid changing the API by any means so we have two options:
- extend the API with `host` (this is what @JasonPowr suggested)
- add `externalAccess` but still can't remove `rekorSearchUI.enabled`. I think there is no need for another action (a simple `if` statement in the controller logic can solve it)

I do not have any strong opinion on this. We can go either way and fix the API design in the new version. The question that I am trying to solve in my mind is whether the "code hack" is worth unified API.

bouskaJ: I just checked the code base. It seems there is different purpose of the `rekorSearchUI.enabled` - it disable the whole SearchUI deployment (https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/rekor/actions/ui/deployment.go#L36) and not just route as `externalAccess.enable` do so I would prefer adding the `host` field.

PR #443, opened 2024-06-14
[SECURESIGN-1044] Error handler

bouskaJ:
This is my proposal on how to solve https://issues.redhat.com/browse/SECURESIGN-1001 or similar issues.

### DIFF
--- a/api/v1alpha1/ctlog_types.go
+++ b/api/v1alpha1/ctlog_types.go
@@ -49,6 +49,9 @@ type CTlogStatus struct {
 	RootCertificates      []SecretKeySelector   `json:"rootCertificates,omitempty"`
 	// The ID of a Trillian tree that stores the log data.
 	TreeID *int64 `json:"treeID,omitempty"`
+	// Number of component recovery attempts.
+	//+kubebuilder:default:=0
+	RecoveryAttempts int64 `json:"recoveryAttempts,omitempty"`
 	// +listType=map
 	// +listMapKey=type
 	// +patchStrategy=merge
@@ -60,6 +63,7 @@ type CTlogStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`,description="The component status"
+//+kubebuilder:printcolumn:name="Recovery Attempts",type=string,JSONPath=`.status.recoveryAttempts`,description="The component recovery attempts"
 
 // CTlog is the Schema for the ctlogs API
 type CTlog struct {
--- a/api/v1alpha1/rekor_types.go
+++ b/api/v1alpha1/rekor_types.go
@@ -79,6 +79,9 @@ type RekorStatus struct {
 	RekorSearchUIUrl string                `json:"rekorSearchUIUrl,omitempty"`
 	// The ID of a Trillian tree that stores the log data.
 	TreeID *int64 `json:"treeID,omitempty"`
+	// Number of component recovery attempts.
+	//+kubebuilder:default:=0
+	RecoveryAttempts int64 `json:"recoveryAttempts,omitempty"`
 	// +listType=map
 	// +listMapKey=type
 	// +patchStrategy=merge
@@ -91,6 +94,7 @@ type RekorStatus struct {
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`,description="The component status"
 //+kubebuilder:printcolumn:name="URL",type=string,JSONPath=`.status.url`,description="The component url"
+//+kubebuilder:printcolumn:name="Recovery Attempts",type=string,JSONPath=`.status.recoveryAttempts`,description="The component recovery attempts"
 
 // Rekor is the Schema for the rekors API
 type Rekor struct {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -94,6 +94,7 @@ func main() {
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
 	flag.Int64Var(&constants.CreateTreeDeadline, "create-tree-deadline", constants.CreateTreeDeadline, "The time allowance (in seconds) for the create tree job to run before failing.")
 	utils.BoolFlagOrEnv(&constants.Openshift, "openshift", "OPENSHIFT", false, "Enable to ensures the operator applies OpenShift specific configurations.")
+	flag.Int64Var(&constants.AllowedRecoveryAttempts, "recovery-attempts", constants.AllowedRecoveryAttempts, "Specifies the maximum number of recovery attempts allowed before an operation is considered failed.")
 	utils.StringFlagOrEnv(&constants.TrillianLogSignerImage, "trillian-log-signer-image", "TRILLIAN_LOG_SIGNER_IMAGE", constants.TrillianLogSignerImage, "The image used for trillian log signer.")
 	utils.StringFlagOrEnv(&constants.TrillianServerImage, "trillian-log-server-image", "TRILLIAN_LOG_SERVER_IMAGE", constants.TrillianServerImage, "The image used for trillian log server.")
 	utils.StringFlagOrEnv(&constants.TrillianDbImage, "trillian-db-image", "TRILLIAN_DB_IMAGE", constants.TrillianDbImage, "The image used for trillian's database.")
--- a/internal/controller/common/action/base_action.go
+++ b/internal/controller/common/action/base_action.go
@@ -8,11 +8,14 @@ import (
 	"strings"
 	"time"
 
-	"github.com/securesign/operator/internal/controller/annotations"
-
 	"github.com/go-logr/logr"
+	"github.com/securesign/operator/internal/apis"
+	"github.com/securesign/operator/internal/controller/annotations"
+	"github.com/securesign/operator/internal/controller/constants"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	client2 "sigs.k8s.io/controller-runtime/pkg/client"
@@ -49,21 +52,52 @@ func (action *BaseAction) StatusUpdate(ctx context.Context, obj client2.Object) 
 		if strings.Contains(err.Error(), OptimisticLockErrorMsg) {
 			return &Result{Result: reconcile.Result{RequeueAfter: 1 * time.Second}, Err: nil}
 		}
-		return action.Failed(err)
+		return action.Error(err)
 	}
 	// Requeue will be caused by update
 	return &Result{Result: reconcile.Result{Requeue: false}}
 }
 
-func (action *BaseAction) Failed(err error) *Result {
+func (action *BaseAction) Error(err error) *Result {
 	action.Logger.Error(err, "error during action execution")
 	return &Result{
-		Result: reconcile.Result{RequeueAfter: time.Duration(5) * time.Second},
-		Err:    err,
+		Err: err,
 	}
 }
 
-func (action *BaseAction) FailedWithStatusUpdate(ctx context.Context, err error, instance client2.Object) *Result {
+// ErrorWithStatusUpdate - Set `Error` status on deployment and execute error-recovery loop in 10 second
+func (action *BaseAction) ErrorWithStatusUpdate(ctx context.Context, err error, instance apis.ConditionsAwareObject) *Result {
+	action.Recorder.Event(instance, v1.EventTypeWarning, constants.Error, err.Error())
+
+	instance.SetCondition(metav1.Condition{
+		Type:    constants.Ready,
+		Status:  metav1.ConditionFalse,
+		Reason:  constants.Error,
+		Message: err.Error(),
+	})
+
+	if e := action.Client.Status().Update(ctx, instance); e != nil {
+		if strings.Contains(err.Error(), OptimisticLockErrorMsg) {
+			return &Result{Result: reconcile.Result{RequeueAfter: 1 * time.Second}, Err: err}
+		}
+		err = errors.Join(e, err)
+	}
+	// Requeue is disabled for Error objects
+	// wait for 10 seconds and invoke error-handler
+	return &Result{Result: reconcile.Result{RequeueAfter: 10 * time.Second}}
+}
+
+// FailWithStatusUpdate - Throw deployment to the Failure state with no error-recovery attempts
+func (action *BaseAction) FailWithStatusUpdate(ctx context.Context, err error, instance apis.ConditionsAwareObject) *Result {
+	action.Recorder.Event(instance, v1.EventTypeWarning, constants.Failure, err.Error())
+
+	instance.SetCondition(metav1.Condition{
+		Type:    constants.Ready,
+		Status:  metav1.ConditionFalse,
+		Reason:  constants.Failure,
+		Message: err.Error(),
+	})
+
 	if e := action.Client.Status().Update(ctx, instance); e != nil {
 		if strings.Contains(err.Error(), OptimisticLockErrorMsg) {
 			return &Result{Result: reconcile.Result{RequeueAfter: 1 * time.Second}, Err: err}
@@ -71,7 +105,7 @@ func (action *BaseAction) FailedWithStatusUpdate(ctx context.Context, err error,
 		err = errors.Join(e, err)
 	}
 	// Requeue will be caused by update
-	return &Result{Result: reconcile.Result{Requeue: false}, Err: err}
+	return &Result{Result: reconcile.Result{Requeue: false}}
 }
 
 func (action *BaseAction) Return() *Result {
--- a/internal/controller/common/action/transitions/restart_on_error.go
+++ b/internal/controller/common/action/transitions/restart_on_error.go
@@ -1,0 +1,84 @@
+package transitions
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	"github.com/securesign/operator/internal/apis"
+	"github.com/securesign/operator/internal/controller/common/action"
+	"github.com/securesign/operator/internal/controller/constants"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func NewRestartOnErrorAction[T apis.ConditionsAwareObject]() action.Action[T] {
+	return &restartAction[T]{}
+}
+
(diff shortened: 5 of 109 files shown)


bouskaJ:
/hold

we need to implement similar functionality for CTlog before decreasing `CreateTreeDeadline` timeout.

bouskaJ:
@JasonPowr @tommyd450 @osmman WDYT?

internal/controller/common/action/base_action.go
```diff
@@ -70,8 +70,9 @@ func (action *BaseAction) FailedWithStatusUpdate(ctx context.Context, err error,
 		}
 		err = errors.Join(e, err)
 	}
-	// Requeue will be caused by update
-	return &Result{Result: reconcile.Result{Requeue: false}, Err: err}
+	// Requeue is disabled for Failed objects
+	// wait for 30 second and invoke error-handler
+	return &Result{Result: reconcile.Result{RequeueAfter: 30 * time.Second}}
```
osmman, line 75: Keep original functionality and rather create a new method `RequeueWithStatusUpdate`

osmman (changes requested):

internal/controller/rekor/actions/error.go
```diff
@@ -0,0 +1,62 @@
...
+
+type handleErrorAction struct {
+	action.BaseAction
+}
+
+func (i handleErrorAction) Name() string {
+	return "error handler"
+}
+
+func (i handleErrorAction) CanHandle(_ context.Context, instance *rhtasv1alpha1.Rekor) bool {
```
osmman, line 26: We will need to add some type of brake which will limit number of retries to not ends with infinite loop.

internal/controller/common/action/base_action.go
bouskaJ: Calling the original function is not safe. After setting the `Failure` status, the only correct way to invoke the ErrorHandler is `RequeueAfter`. Calling the original function will result in new intermediate loop invocation. The ErrorHandler will try to restart and we will end-up in infinite loop **without any timeout**.

internal/controller/rekor/actions/error.go
bouskaJ: I kept the way that is implemented on Pod lifecycle (`Always`,`OnFailure` actions restarts the pod infinite number of times). I do agree, we need to implement some sort of strategy to stop it. This is only PoC.

internal/controller/rekor/actions/server/generate_signer.go
```diff
@@ -240,3 +225,48 @@ func (g generateSigner) CreateRekorKey(instance *v1alpha1.Rekor) (*RekorCertConf
 		RekorPubKey: pemPubKey.Bytes(),
 	}, nil
 }
+
+func (g generateSigner) CanHandleError(_ context.Context, instance *rhtasv1alpha1.Rekor) bool {
+	return !meta.IsStatusConditionTrue(instance.GetConditions(), actions.SignerCondition) &&
+		(instance.Status.Signer.KeyRef != nil || instance.Status.Signer.PasswordRef != nil)
+}
+
+func (g generateSigner) HandleError(ctx context.Context, instance *rhtasv1alpha1.Rekor) *action.Result {
```
osmman, line 234: I would recommend against attempting error recovery for similar events. It is better to adhere to a fail-fast approach in this context. In other words, if something goes wrong, stop the reconciliation process immediately.

api/v1alpha1/ctlog_types.go
```diff
@@ -49,6 +49,9 @@ type CTlogStatus struct {
 	RootCertificates      []SecretKeySelector   `json:"rootCertificates,omitempty"`
 	// The ID of a Trillian tree that stores the log data.
 	TreeID *int64 `json:"treeID,omitempty"`
+	// Number of component restarts.
+	//+kubebuilder:default:=0
+	Restarts int `json:"restarts,omitempty"`
```
osmman, line 54: I am not sure that `Restarts` is a good name. It is to vague and generic. Something like:
- RetryCount
- RecoveryRetryCount
- FailureRecoveryAttempts
- RecoveryAttempts
- RetryAttempts

osmman (changes requested):
**Missing logic to reset Restart counter:**
Currently, there is no mechanism to reset the retry counter after a successful recovery from failure. As a result, the number of attempts is limited to a fixed value for the entire lifecycle of the object.

**Unclear need for Restarting phase:**
The introduction of a new Restarting phase is confusing, and its purpose is not well-defined. Documentation is needed to explain the meaning of all phases and how they relate to each other.

**Error handling order:**
Errors are not being handled by the designated error handlers. Instead, the first action in the sequence always consumes the error, which can lead to repeated deletion and recreation of objects without reaching the appropriate error handler.

**Lack of Fail-Fast mechanism:**
There is no way to configure the operator to interrupt reconciliation without attempting to recover from failures (fail-fast). Most failures should simply cause the reconciliation to stop, rather than trying to recover. Recover functionality should be reserved for only a small subset of actions that can genuinely recover from their failed state.

internal/controller/constants/constants.go
```diff
@@ -9,4 +9,5 @@ const (
 	Creating   = "Creating"
 	Initialize = "Initialize"
 	Failure    = "Failure"
+	Restarting = "Restarting"
```
osmman, line 12:
```go
// The operator has encountered a recoverable error and is attempting to recover. This state includes a counter (RecoveryAttempts) to track the number of recovery attempts.
Recovering  = "Recovering"
// The operator has encountered an unrecoverable error or exceeded the maximum number of recovery attempts. Reconciliation stops.
Failure          = "Failure"
// An error has occurred, but it’s not yet classified as recoverable or irrecoverable. This state can transition to either Recovering or Failed based on error type and policy.
Error             = "Error"
```

internal/controller/rekor/actions/server/generate_signer.go
bouskaJ: I created special return type `i.FailedWithStatusUpdate`. I agree that we can throw it for example on wrong user input but it does make sense retry operator-generated keys.

api/v1alpha1/rekor_types.go
```diff
@@ -79,6 +79,9 @@ type RekorStatus struct {
 	RekorSearchUIUrl string                `json:"rekorSearchUIUrl,omitempty"`
 	// The ID of a Trillian tree that stores the log data.
 	TreeID *int64 `json:"treeID,omitempty"`
+	// Number of component restart attempts.
```
osmman, line 82:
```suggestion
	// Number of component recovery attempts.
```

cmd/main.go
```diff
@@ -94,6 +94,7 @@ func main() {
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
 	flag.Int64Var(&constants.CreateTreeDeadline, "create-tree-deadline", constants.CreateTreeDeadline, "The time allowance (in seconds) for the create tree job to run before failing.")
 	utils.BoolFlagOrEnv(&constants.Openshift, "openshift", "OPENSHIFT", false, "Enable to ensures the operator applies OpenShift specific configurations.")
+	flag.Int64Var(&constants.AllowedRestarts, "allowed-restarts", constants.AllowedRestarts, "Allowed restart threshold.")
```
osmman, line 97:
```suggestion
	flag.Int64Var(&constants.AllowedRecoveryAttempts, "allowed-recovery-attempts", constants.AllowedRecoveryAttempts, "Specifies the maximum number of recovery attempts allowed before an operation is considered failed.")
```

internal/controller/common/action/transitions/restart_on_error.go
```diff
@@ -0,0 +1,84 @@
...
+	restarts++
+	err = i.setRestartCount(instance, restarts)
+	if err != nil {
+		return i.Error(err)
+	}
+	if restarts < constants.AllowedRestarts {
+		instance.SetCondition(metav1.Condition{Type: constants.Ready,
+			Status: metav1.ConditionFalse, Reason: constants.Pending})
+	} else {
+		return i.FailWithStatusUpdate(ctx, fmt.Errorf("restart threshold reached"), instance)
```
osmman, line 61:
```suggestion
		return i.FailWithStatusUpdate(ctx, fmt.Errorf("recovery threshold reached"), instance)
```

internal/controller/common/action/transitions/restart_on_error.go
```diff
@@ -0,0 +1,84 @@
...
+	"reflect"
+
+	"github.com/securesign/operator/internal/apis"
+	"github.com/securesign/operator/internal/controller/common/action"
+	"github.com/securesign/operator/internal/controller/constants"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func NewRestartOnErrorAction[T apis.ConditionsAwareObject]() action.Action[T] {
```
osmman, line 15: Could you please add tests for the **RestartOnError** action? It's crucial to have this action thoroughly covered by tests since it will be heavily utilized. Ensuring test coverage will help prevent potential breakages in future changes.

internal/controller/common/utils/kubernetes/deployment.go
```diff
@@ -10,17 +10,27 @@ import (
 
 func DeploymentIsRunning(ctx context.Context, cli client.Client, namespace string, labels map[string]string) (bool, error) {
```
osmman, line 11: I think that we should depend Deployment's `Available` condition. It will work much better in feature where we will need to depend on minimum available replicas for HA.

```yaml
kind: Deployment
spec:
  replicas: 5
  strategy:
    rollingUpdate:
      maxUnavailable: 0 # 🔰
status:
  conditions:
    - lastTransitionTime: "2020-07-07T14:04:27Z"
      lastUpdateTime: "2020-07-07T14:04:27Z"
      message: Deployment does not have minimum availability.
      reason: MinimumReplicasUnavailable
      status: "False"
      type: Available
    - lastTransitionTime: "2020-07-07T14:04:27Z"
      lastUpdateTime: "2020-07-07T14:04:27Z"
      message: 'pods "test-7df57bd99d-5qw47" is forbidden: exceeded quota: mem-cpu-demo, requested: requests.memory=50Mi, used: requests.memory=200Mi, limited: requests.memory=200Mi'
      reason: FailedCreate
      status: "True"
      type: ReplicaFailure
    - lastTransitionTime: "2020-07-07T14:04:27Z"
      lastUpdateTime: "2020-07-07T14:04:38Z"
      message: ReplicaSet "test-7df57bd99d" is progressing.
      reason: ReplicaSetUpdated
      status: "True"
      type: Progressing
  replicas: 4
  availableReplicas: 4
  readyReplicas: 4
  unavailableReplicas: 1
  updatedReplicas: 4
```

internal/controller/common/action/base_action.go
```diff
@@ -49,29 +53,61 @@ func (action *BaseAction) StatusUpdate(ctx context.Context, obj client2.Object)
...
 	return &Result{
 		Result: reconcile.Result{RequeueAfter: time.Duration(5) * time.Second},
 		Err:    err,
 	}
 }
 
-func (action *BaseAction) FailedWithStatusUpdate(ctx context.Context, err error, instance client2.Object) *Result {
+// ErrorWithStatusUpdate - Set `Error` status on deployment and execute error-recovery loop in 1 minute
+func (action *BaseAction) ErrorWithStatusUpdate(ctx context.Context, err error, instance apis.ConditionsAwareObject) *Result {
+	action.Recorder.Event(instance, v1.EventTypeWarning, constants.Error, err.Error())
```
osmman, line 72: lets add information which action produced error. We have two options modify message body to contains `action.name()`: 
```
action.Recorder.Eventf(instance, v1.EventTypeWarning, constants.Error, "%s: %s", action.name(), err.Error())
```

or produce annotated event:
```
action.Recorder.AnnotatedEventf(instance, map[string]string{"action":action.name()},v1.EventTypeWarning, constants.Error, "%s: %s", action.name(), err.Error())
```

internal/controller/common/action/base_action.go
```diff
@@ -49,29 +53,61 @@ func (action *BaseAction) StatusUpdate(ctx context.Context, obj client2.Object)
...
+
```
osmman, line 73: lets automaticaly add condition for failed action, that could be used for CanHandleError to to identify it's responsibility for recovery.

```
instance.SetCondition(metav1.Condition{Type: action.name(), Reason: constant.Error})
```

internal/controller/common/action/base_action.go
bouskaJ: `action.Name()` is not available on `BaseAction` it would need more code alignment. I would not like to mixed with this already complex PR.

internal/controller/common/action/base_action.go
```diff
@@ -49,29 +52,61 @@ func (action *BaseAction) StatusUpdate(ctx context.Context, obj client2.Object)
...
+	if e := action.Client.Status().Update(ctx, instance); e != nil {
+		if strings.Contains(err.Error(), OptimisticLockErrorMsg) {
+			return &Result{Result: reconcile.Result{RequeueAfter: 1 * time.Second}, Err: err}
+		}
+		err = errors.Join(e, err)
+	}
+	// Requeue is disabled for Error objects
+	// wait for 1 minute and invoke error-handler
+	return &Result{Result: reconcile.Result{RequeueAfter: 1 * time.Minute}}
```
osmman, line 88: Currently our tests take 10x more time to be executed and I think that it has been caused by this 1 minute requeue. We should use smaller value for default behavior and if some action relay need some time than we should define it in that action. 5-10 second could work

bouskaJ: I modified some timeouts. We should be fine now.

internal/controller/common/action/base_action.go
```diff
@@ -49,29 +52,61 @@ func (action *BaseAction) StatusUpdate(ctx context.Context, obj client2.Object)
...
+	})
+
+	if e := action.Client.Status().Update(ctx, instance); e != nil {
+		if strings.Contains(err.Error(), OptimisticLockErrorMsg) {
+			return &Result{Result: reconcile.Result{RequeueAfter: 1 * time.Second}, Err: err}
+		}
+		err = errors.Join(e, err)
+	}
+	// Requeue is disabled for Error objects
+	// wait for 1 minute and invoke error-handler
```
osmman, line 87:
```suggestion
	// wait for 10 second and invoke error-handler
```

internal/controller/rekor/actions/server/generate_signer.go
```diff
@@ -124,36 +118,33 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Rekor) *a
 	}
 	secret := k8sutils.CreateImmutableSecret(fmt.Sprintf(secretNameFormat, instance.Name), instance.Namespace,
 		data, labels)
+	if err = controllerutil.SetControllerReference(instance, secret, g.Client.Scheme()); err != nil {
```
osmman, line 121: do not modify behavior of this action, secret should not be deleted

internal/controller/rekor/actions/server/monitoring.go
```diff
@@ -131,15 +124,32 @@ func (i monitoringAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Re
...
+		return i.ErrorWithStatusUpdate(ctx, fmt.Errorf("could not create serviceMonitor: %w", err), instance)
 	}
 
 	// monitors & RBAC are not watched - do not need to re-enqueue
 	return i.Continue()
 }
+
+func (i monitoringAction) CanHandleError(ctx context.Context, instance *rhtasv1alpha1.Rekor) bool {
+	err := i.Client.Get(ctx, types.NamespacedName{Name: actions.ServerDeploymentName, Namespace: instance.Namespace}, &monitoringv1.ServiceMonitor{})
+	return !meta.IsStatusConditionTrue(instance.GetConditions(), actions.ServerCondition) && instance.Spec.Monitoring.Enabled && (err == nil || !errors.IsNotFound(err))
```
osmman, line 136:
```suggestion
	return meta.IsStatusConditionFalse(instance.GetConditions(), actions.ServerCondition) && instance.Spec.Monitoring.Enabled && (err == nil || !errors.IsNotFound(err))
```

internal/controller/rekor/actions/ui/svc.go
```diff
@@ -76,3 +73,26 @@ func (i createServiceAction) Handle(ctx context.Context, instance *rhtasv1alpha1
 		return i.Continue()
 	}
 }
+
+func (i createServiceAction) CanHandleError(ctx context.Context, instance *rhtasv1alpha1.Rekor) bool {
+	err := i.Client.Get(ctx, types.NamespacedName{Name: actions.SearchUiDeploymentName, Namespace: instance.Namespace}, &v1.Service{})
+	return *instance.Spec.RekorSearchUI.Enabled && !meta.IsStatusConditionTrue(instance.GetConditions(), actions.UICondition) && (err == nil || !errors.IsNotFound(err))
```
osmman, line 79: nil in `instance.Spec.RekorSearchUI.Enabled` will cause panic error

internal/controller/rekor/rekor_controller.go
```diff
@@ -100,6 +105,9 @@ func (r *RekorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 			if *rekor.Spec.RekorSearchUI.Enabled {
```
osmman, line 105: https://github.com/securesign/secure-sign-operator/pull/443/files#r1691636450

internal/controller/rekor/rekor_controller.go
```diff
@@ -100,6 +105,9 @@ func (r *RekorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
 				components = append(components, actions2.UICondition)
 			}
+			if *rekor.Spec.BackFillRedis.Enabled {
```
osmman, line 108: https://github.com/securesign/secure-sign-operator/pull/443/files#r1691636450

internal/controller/rekor/actions/server/monitoring.go
bouskaJ: I use ` !meta.IsStatusConditionTrue` on purpose. It will run on `ConditionUnknown` too.

internal/controller/common/action/base_action.go
```diff
@@ -49,29 +52,61 @@ func (action *BaseAction) StatusUpdate(ctx context.Context, obj client2.Object)
...
 		}
-		return action.Failed(err)
+		return action.Error(err)
 	}
 	// Requeue will be caused by update
 	return &Result{Result: reconcile.Result{Requeue: false}}
 }
 
-func (action *BaseAction) Failed(err error) *Result {
+func (action *BaseAction) Error(err error) *Result {
```
osmman, line 61: Error function do not work correctly at least it do not behave live ErrorWithStatusUpdate. 
Event our return values are incorrect controller returns this warning:
```
Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler
```

bouskaJ:
This PR is out of date - there is too many conflicting files and the error handler approach is partially handled by https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/common/action/base_action.go#L77 (Terminal/non-Terminal errors)

PR #449, opened 2024-06-18
[SECURESIGN-1037] inject certificates if proxy is enabled

fghanmi:
replaces PR #448


osmman:
The `proxy` parameter in CRD is currently pointless. The operator should be proxy aware and automatically configure proxy for it's operands automatically. This means that it detects `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` env variables above the operator deployment and use them accordingly.

PR #458, opened 2024-06-21
[SECURESIGN-663] Add option to set external ctlog service. Refactor services port name.

fghanmi:
- Add the option to use an external CTLog
- Expose Trillian log server via a route
- Refactor of services port name done on: [Ctlog, Trillian]
Another PR will be created for the rest of components.


internal/controller/common/utils/kubernetes/service.go
```diff
@@ -12,7 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-func CreateService(namespace string, name string, port int, labels map[string]string) *corev1.Service {
+func CreateService(namespace string, name string, portName string, port int, labels map[string]string) *corev1.Service {
```
osmman, line 15: Please create a separate PR for refactoring `CreateService` method and it's required changes.

osmman (changes requested):

fghanmi:
To be declined: 
Will create sub PRs instead.

PR #459, opened 2024-06-24
Add new flag for manager to define target k8s environment

osmman:


internal/controller/common/utils/flag_or_env.go
```diff
@@ -14,3 +15,13 @@ func StringFlagOrEnv(p *string, name string, envName string, defaultValue string
...
+
+// BoolFlagOrEnv defines a bool flag which can be set by an environment variable.
+// Precedence: flag > env var > default value.
+func BoolFlagOrEnv(p *bool, name string, envName string, defaultValue bool, usage string) {
+	envValue := os.Getenv(envName)
+	if envName != "" {
+		defaultValue, _ = strconv.ParseBool(envValue)
+	}
+	flag.BoolVar(p, name, defaultValue, usage)
+}
```
bouskaJ, line 27: Add a new line

cmd/main.go
```diff
@@ -92,6 +93,7 @@ func main() {
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
 	flag.Int64Var(&constants.CreateTreeDeadline, "create-tree-deadline", constants.CreateTreeDeadline, "The time allowance (in seconds) for the create tree job to run before failing.")
+	utils.BoolFlagOrEnv(&constants.Openshift, "openshift", "OPENSHIFT", false, "Enable to ensures the operator applies OpenShift specific configurations.")
```
bouskaJ, line 96: Is the `false` intentional? Do we really want to have different values (main.go/rhtas-operator.clusterserviceversion.yaml)?

osmman: Yes, the `false` value is intentional. The goal is to keep the flag set to false by default for the manager executable to simplify local development and testing.

PR #465, opened 2024-06-25
[SECURESIGN-663] Add option to set an external ctlog service

fghanmi:


api/v1alpha1/common.go
```diff
@@ -34,6 +34,19 @@ type TrillianService struct {
...
 
+// CtlogService configuration to connect Ctlog server
+type CtlogService struct {
+	// Address to Ctlog Log Server End point
+	//+optional
+	Address string `json:"address,omitempty"`
+	// Port of Ctlog Log Server End point
+	//+kubebuilder:validation:Minimum:=1
+	//+kubebuilder:validation:Maximum:=65535
+	//+kubebuilder:default:=6963
```
osmman, line 45: default port fot ctlog is `80`, port 6963 is used for metrics

api/v1alpha1/fulcio_types.go
```diff
@@ -11,6 +11,10 @@ import (
 type FulcioSpec struct {
 	// Define whether you want to export service or not
 	ExternalAccess ExternalAccess `json:"externalAccess,omitempty"`
+	// Ctlog service configuration
+	//+kubebuilder:default:={port: 6963}
```
osmman, line 15:
```suggestion
	//+kubebuilder:default:={port: 80}
```

internal/controller/fulcio/utils/fulcio_deployment.go
```diff
@@ -38,7 +38,23 @@ func CreateDeployment(instance *v1alpha1.Fulcio, deploymentName string, sa strin
...
-		fmt.Sprintf("--ct-log-url=http://ctlog.%s.svc/trusted-artifact-signer", instance.Namespace)}
+	}
+
+	var err error
+	var ctlogUrl string
+	switch {
+	case instance.Spec.Ctlog.Port == nil && instance.Spec.Ctlog.Address != "":
+		err = fmt.Errorf("%w", CtlogPortNotSpecified)
+	case instance.Spec.Ctlog.Address == "":
+		ctlogUrl = fmt.Sprintf("http://ctlog.%s.svc/trusted-artifact-signer", instance.Namespace)
```
bouskaJ, line 49: We can extract https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/ctlog/utils/ctlog_config.go#L177 to constant and reuse for better transparency. (it is not clear from where the URL suffix is getting there).

internal/controller/fulcio/utils/fulcio_deployment.go
```diff
@@ -38,7 +38,23 @@ func CreateDeployment(instance *v1alpha1.Fulcio, deploymentName string, sa strin
 		"/var/run/fulcio-secrets/key.pem",
 		"--fileca-cert",
 		"/var/run/fulcio-secrets/cert.pem",
-		fmt.Sprintf("--ct-log-url=http://ctlog.%s.svc/trusted-artifact-signer", instance.Namespace)}
+	}
+
+	var err error
+	var ctlogUrl string
+	switch {
+	case instance.Spec.Ctlog.Port == nil && instance.Spec.Ctlog.Address != "":
```
osmman, line 46: operator will fails on fatal error if `port` is `nil` and address has any non empty value. It will unable to dereference `*instance.Spec.Ctlog.Port` value if it is nil.

Please do that same way as [rekor deployment](https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/rekor/utils/rekor_deployment.go#L22-L26).

internal/controller/fulcio/utils/fulcio_deployment.go
osmman, line 49: Please move logic to calculate default value for address outside of util method. Here is my reasoning from [previos PR](https://github.com/securesign/secure-sign-operator/pull/196#discussion_r1629065677) for Rekor.

osmman (changes requested):

osmman:
Please run `make generate manifest bundle` command to generate all manifests and bundle metadata

internal/controller/fulcio/actions/deployment.go
```diff
@@ -37,6 +37,12 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Fulcio
 	)
 
 	labels := constants.LabelsFor(ComponentName, DeploymentName, instance.Name)
+
+	if instance.Spec.Ctlog.Address == "" {
+		port := int32(80)
```
osmman, line 42: If address is empty you should not override defined port. Here is example what should be possible to configure:
```
fulcio:
   ctlog:
      port: 9999
```
it should generate 'http://ctlog.namespace.svc:9999/trusted-artifact-signer' but your implementation will override port to `80`

osmman (approved):

PR #469, opened 2024-06-25
[SECURESIGN-1137] refactor services port name

fghanmi:

PR #474, opened 2024-06-26
[SECURESIGN-1137] refactor services port name

fghanmi:
replaces #469


fghanmi:
> /lgtm

seems I have an issue with the build.. checking...

PR #484, opened 2024-06-30
[SECURESIGN-1136] add option to expose trillian logserver service

fghanmi:


bouskaJ:
I am thinking about the use-case. With the full-stack deployment, the communication that the user would do should be done against Rekor or CTLog and NEVER directly to the trillian.

The other case is deploying the Trillian as a single component and connect to it from external deployment. In this case, you can expose the service by your own.

I am considering this because I would like to avoid over-complication the API with exposing the single one and in the most cases internal service.

@JasonPowr @osmman @lance WDYT?

lance:
> The other case is deploying the Trillian as a single component and connect to it from external deployment. In this case, you can expose the service by your own.

This is the motivation behind the change.

> I am considering this because I would like to avoid over-complication the API with exposing the single one and in the most cases internal service.

I don't have very strong feelings about this, but I do recognize that the use case is likely to be very uncommon. As maintainers of this code, I think you all should make the final decision. It would be nice for customers to be able to just deploy trillian using the operator install screen. But that's just a nicety.

PR #488, opened 2024-07-03
feat: provide host name option for CLI and search ui

JasonPowr:
I messed up the commit tree in https://github.com/securesign/secure-sign-operator/pull/414, so reopening here


JasonPowr:
@osmman @bouskaJ From the conversation here: https://github.com/securesign/secure-sign-operator/pull/414#discussion_r1638110085, it seems like we are ok with just adding the host field, can ye ptal :)  thanks

lance (commented):
/lgtm
/hold for review from @osmman

PR #490, opened 2024-07-03
Extract shared actions

bouskaJ:


internal/controller/common/action/transitions/to_initialize.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+
+	"github.com/securesign/operator/internal/controller/common/action"
+	"github.com/securesign/operator/internal/controller/constants"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func NewToInitializeAction[T action.ConditionsAwareObject]() action.Action[T] {
```
osmman, line 12: please use same name format for transition. Create and pending actions has phase keyword vs initialise which is missing it

internal/controller/common/action/conditions_aware.go
```diff
@@ -0,0 +1,13 @@
+package action
```
osmman, line 1: I think it would be better to move this interface to a separate package `conditions` or `apis` package. It will enable us to create helper functions for it. For example `IsFailure(ConditionsAwareObject) bool`

internal/controller/rekor/rekor_controller.go
```diff
@@ -87,14 +88,13 @@ func (r *RekorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		return reconcile.Result{}, err
 	}
 	target := instance.DeepCopy()
-	actions := []action.Action[rhtasv1alpha1.Rekor]{
-		// NONE -> PENDING
-		actions2.NewInitializeConditions(),
+	actions := []action.Action[*rhtasv1alpha1.Rekor]{
+		transitions.NewToPendingPhaseAction[*rhtasv1alpha1.Rekor](actions2.ServerCondition, actions2.RedisCondition, actions2.SignerCondition, actions2.UICondition),
```
osmman, line 92: how it will looks like when I deploy rekor with disabled Search UI? Will it's status be stuck at Pending state?

bouskaJ: It will stay in unknown. I can add condition for that

internal/controller/common/action/transitions/to_pending_phase.go
```diff
@@ -0,0 +1,40 @@
...
+	"context"
+
+	"github.com/securesign/operator/internal/apis"
+	"github.com/securesign/operator/internal/controller/common/action"
+	"github.com/securesign/operator/internal/controller/constants"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func NewToPendingPhaseAction[T apis.ConditionsAwareObject](componentSupplier func(T) []string) action.Action[T] {
```
osmman, line 13: You can define a new type ComponentSupplier. Like:
```go
type ComponentSupplier [T apis.ConditionsAwareObject]func(T) []string 
```

```suggestion
func NewToPendingPhaseAction[T apis.ConditionsAwareObject](supplier ComponentSupplier[T]) action.Action[T] {
```

internal/controller/tuf/tuf_controller.go
```diff
@@ -87,16 +88,23 @@ func (r *TufReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 	}
 
 	target := instance.DeepCopy()
-	acs := []action.Action[rhtasv1alpha1.Tuf]{
-		actions.NewToPendingPhaseAction(),
+	acs := []action.Action[*rhtasv1alpha1.Tuf]{
+		transitions.NewToPendingPhaseAction[*rhtasv1alpha1.Tuf](func(tuf *rhtasv1alpha1.Tuf) []string {
+			keys := make([]string, len(instance.Spec.Keys))
```
osmman, line 93:
```suggestion
			keys := make([]string, len(tuf.Spec.Keys))
```

internal/controller/tuf/tuf_controller.go
```diff
@@ -87,16 +88,23 @@ func (r *TufReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
...
+			for i, k := range instance.Spec.Keys {
```
osmman, line 94:
```suggestion
			for i, k := range tuf.Spec.Keys {
```

PR #491, opened 2024-07-03
Inject CA trust bundle into managed containers and set SSL_CERT_DIR

osmman:
related to https://issues.redhat.com/browse/SECURESIGN-703 and https://issues.redhat.com/browse/SECURESIGN-1037

### DIFF
--- a/internal/controller/annotations/annotations.go
+++ b/internal/controller/annotations/annotations.go
@@ -6,4 +6,23 @@ const (
 
 	// Metrics Annotation is used to control the sending of analytic metrics of the installed services managed by the operator.
 	Metrics = "rhtas.redhat.com/metrics"
+
+	// TrustedCA Annotation to specify name of ConfigMap with additional bundle of trusted CA
+	TrustedCA = "rhtas.redhat.com/trusted-ca"
 )
+
+var inheritable = []string{
+	TrustedCA,
+}
+
+func FilterInheritable(annotations map[string]string) map[string]string {
+	result := make(map[string]string, 0)
+	for key, value := range annotations {
+		for _, ia := range inheritable {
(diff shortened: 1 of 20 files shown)


internal/controller/annotations/annotations.go
```diff
@@ -6,4 +6,23 @@ const (
...
+
+	// TrustedCA Annotation to specify name of ConfigMap with additional bundle of trusted CA
+	TrustedCA = "rhtas.redhat.com/trusted-ca"
 )
+
+var inheritable = []string{
+	TrustedCA,
+}
+
+func FilterInheritable(annotations map[string]string) map[string]string {
```
bouskaJ, line 18: What is purpose of this function? Why you just don't use `annotations[TrustedCA]`?

bouskaJ:
Can you create `e2e` test for this feature?

bouskaJ:
> Can you create `e2e` test for this feature?

I mean can modify the CI to use keycloak (self signed) route instead of the insecure service https://github.com/securesign/secure-sign-operator/blob/main/.github/workflows/main.yml#L187 This should verify `kuberoot` CA injection AFIK

osmman:
> > Can you create `e2e` test for this feature?
> 
> I mean can modify the CI to use keycloak (self signed) route instead of the insecure service https://github.com/securesign/secure-sign-operator/blob/main/.github/workflows/main.yml#L187 This should verify `kuberoot` CA injection AFIK
> 
> I am not sure how to test the annotation injection. But with that ^^ we will know that the volume is correctly mounted and the `ENV` property works fine

Yes, that will be good to modify Keycloack deployment on Kind cluster. I think Openshift CI uses self signed certificates if I remember correctly.

PR #492, opened 2024-07-04
[SECURESIGN-994] Add TLS to Fulcio and CTlog services

fghanmi:


api/v1alpha1/ctlog_types.go
```diff
@@ -48,6 +48,9 @@ type CTlogSpec struct {
 	// publicKeyRef, rootCertificates and trillian will be overridden.
 	//+optional
 	ServerConfigRef *LocalObjectReference `json:"serverConfigRef,omitempty"`
+	// Configuration for enabling TLS (Transport Layer Security) encryption for manged database.
```
osmman, line 51: please fix doc, it is configuration to encrypt CTlog server

internal/controller/fulcio/actions/deployment.go
```diff
@@ -38,11 +38,20 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Fulcio
 
 	labels := constants.LabelsFor(ComponentName, DeploymentName, instance.Name)
 
-	switch {
-	case instance.Spec.Ctlog.Address == "":
-		instance.Spec.Ctlog.Address = fmt.Sprintf("http://ctlog.%s.svc", instance.Namespace)
-	case instance.Spec.Ctlog.Port == nil:
-		port := int32(80)
+	if instance.Spec.Ctlog.Address == "" {
```
bouskaJ, line 41: `instance.Spec.TrustedCA != nil` does not necessary means that you have TLS enabled CTLog service. I would try to load Ctlog service and try to get scheme and port from it.

internal/controller/fulcio/actions/deployment.go
```diff
@@ -58,6 +67,20 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Fulcio
...
+		if err != nil {
+			meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+				Type:    constants.Ready,
+				Status:  metav1.ConditionFalse,
+				Reason:  constants.Failure,
+				Message: err.Error(),
+			})
+			return i.FailedWithStatusUpdate(ctx, fmt.Errorf("could not get CA path: %w", err), instance)
+		}
+		dp.Spec.Template.Spec.Containers[0].Args = append(dp.Spec.Template.Spec.Containers[0].Args, "--ct-log.tls-ca-cert", caPath)
```
bouskaJ, line 81: not even Fulcio is able to use `SSL_CERT_DIR` variable?

internal/controller/fulcio/actions/deployment.go
fghanmi: So, Fulcio needs to depend on CTlog and should only start after CTlog has started, in order to fetch its service. Is that correct?

bouskaJ: 1) there is no dependency on CTLog itself. You can use label selector to find the service
2) your deployment depends on `ctlog.namespace.svc"` anyway so your fulcio Fulcio will not become fully ready until Ctlog starts. The only downside of this solution is that Fulcio would wait for Ctlog service to be created but I don't think that it is a big problem.

@osmman WDYT?

internal/controller/fulcio/actions/deployment.go
fghanmi: the only TLS option I see is this one:
https://github.com/sigstore/fulcio/blob/main/cmd/app/serve.go#L112

internal/controller/fulcio/actions/deployment.go
fghanmi: @bouskaJ , is [it](https://github.com/securesign/secure-sign-operator/pull/492/files#diff-c3e5a43422675e69c95cc8ceb17db71b241505d328d8db63f886d7013d8e0850R41) better now ?

internal/controller/fulcio/actions/deployment.go
bouskaJ: What I see in the implementation, the server should work with system certs. Give it a try, please. https://github.com/securesign/fulcio/blob/main/cmd/app/serve.go#L279-L302

bouskaJ: Fulcio.Spec.TrustedCA does not necessarily mean that you specify CA for CTLog -> I would prefer to mount it as system CA

internal/controller/fulcio/actions/deployment.go
bouskaJ: I wold do it like this
```
var url
	if instance.Spec.Ctlog.Address != "" && instance.Spec.Ctlog.Port != "" {
	url = // use specified
     
	} else {
svc := FindService(ctx, cli, LabelsForComponent(ctl.ComponentName, instance.Name)
// retrieve protocol from service port
// compose url 
url = fmt.Sprintf("%s%s.svc:%d", svc.Name, svc.Namespace, svc.Port)
     }

url += instance.Spec.Ctlog.Preffix

```

I hope the above snapshot is clear enough.

osmman: please do not depends on services, you need to keep support of independent deployment so ctlog and fulcio could be deployed separatly (different namespace, cluster) and Fulcio will sucessfuly start without runnig CTlog only create certificate requests will be rejected which is intended behavior.

So please try to use only information which are present in Fulcio resource

fghanmi: we have `TrustedCA`, but as Jan said, it does not necessarily mean that CTlog has TLS enabled.. 
that's why in the beginning, I was using a dedicated field for that `TLS.CaCert`, I used it to decide whether fulcio should use TLS or not.

bouskaJ: > please do not depends on services, you need to keep support of independent deployment so ctlog and fulcio could be deployed separatly (different namespace, cluster) and Fulcio will sucessfuly start without runnig CTlog only create certificate requests will be rejected which is intended behavior.
> 
> So please try to use only information which are present in Fulcio resource

I was thinking about the like a fail-over solution in case there is no information in Fulcio resource. With Firas solution, you onli _think_ that you should use tls (based on env) but the running CTL does not needs to be TLS secured at all. You can get much more valid information from service itself. Service lookup (based on labels) does not create any hard dependency and we already use it in the TUF key auto-resolution.

internal/controller/fulcio/actions/deployment.go
bouskaJ: I was able to make it working with system-certs. This is my PoC https://github.com/bouskaJ/operator/commit/b9f41545994d6838581167523f6e359a1ef70ff9

fghanmi: @bouskaJ I've cherry-picked your suggestion, and tested it, all works fine!

bouskaJ:
closing this in favour of https://github.com/securesign/secure-sign-operator/pull/1116

PR #501, opened 2024-07-10
[SECURESIGN-1206] Making segment backup job proxy aware 

SequeI:
Inspects environment proxy variables and appends them to Operand environments making them proxy aware.


lance (commented):
/lgtm

PR #502, opened 2024-07-10
SECURESIGN-1207: fix SBJ breaking on update 

JasonPowr:


internal/controller/common/utils/kubernetes/job.go
```diff
@@ -35,3 +39,15 @@ func CreateJob(namespace string, name string, labels map[string]string, image st
 		},
 	}
 }
+
+func GetJob(ctx context.Context, c client.Client, namespace, jobName string) (bool, *batchv1.Job, error) {
```
osmman, line 43: Other `Get**** functions` uses return type `(Object, error)`, so for consistency, please use the return type (*batchv1.Job, error).

internal/controller/securesign/actions/segment_backup_job.go
```diff
@@ -42,13 +56,23 @@ func (i segmentBackupJob) CanHandle(_ context.Context, instance *rhtasv1alpha1.S
 }
 
 func (i segmentBackupJob) Handle(ctx context.Context, instance *rhtasv1alpha1.Securesign) *action.Result {
-
 	var (
 		err error
 	)
 
-	labels := constants.LabelsFor(SegmentBackupJobName, SegmentBackupJobName, instance.Name)
+	jobID := uuid.New().String()[:5]
```
osmman, line 63: It will be better to let K8s api to handle generation of uuid in name. It can be done by modify CreateJob function to use `generateName` instead of `name` on metadata. We are already doing it for Secrets and ConfigMaps

osmman (changes requested):

internal/controller/common/utils/kubernetes/job.go
```diff
@@ -35,3 +44,15 @@ func CreateJob(namespace string, name string, labels map[string]string, image st
 		},
 	}
 }
+
+func GetJob(ctx context.Context, c client.Client, namespace, jobName string) (*batchv1.Job, error) {
+	job := &batchv1.Job{}
+	err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: jobName}, job)
+	if err == nil {
+		return job, nil
+	} else if errors.IsNotFound(err) {
```
osmman, line 53: Please pass the `IsNotFound` error as well, and allow the caller to determine how to handle it.

osmman: you can remove `else if` block because there are two blocks which returns same value `nil, err`

internal/controller/securesign/actions/segment_backup_job.go
```diff
@@ -68,13 +84,32 @@ func (i segmentBackupJob) Handle(ctx context.Context, instance *rhtasv1alpha1.Se
 	// Adding proxy variables to operand
 	env = append(env, proxy.ReadProxyVarsFromEnv()...)
 
+	// Logic to delete old SBJ to avoid SECURESIGN-1207, can be removed after next release
+	log := ctrllog.FromContext(ctx)
```
osmman, line 88: all our actions should have predefined logger in `i.Logger`

internal/controller/common/utils/kubernetes/job.go
JasonPowr: Thank, I was not thinking hah

osmman (approved):

PR #512, opened 2024-07-17
SECURESIGN-568: Extend Rekor API to support sharding configuration

osmman:

PR #513, opened 2024-07-17
doc: Rotation the signer key for Rekor service

osmman:
Depends on: https://github.com/securesign/secure-sign-operator/pull/512


lance (approved):
lgtm - though you might want to put a warning or note at the top that this procedure is incomplete and should not be done until the final step (updating TUF) is possible and documented.

PR #516, opened 2024-07-18
chore: add image preflight checks to pipelines

lance:
Konflux does not have a way of introducing new tasks into our build pipelines automatically. Instead, we need to do these things manually. The image preflight checks are new as of a couple of months ago, and are now being added to all new components as they are onboarded. This commit brings the operator pipelines up to date with that addition. Apparently, it also removes a whitespace from the cel expression.


JasonPowr (approved):
/lgtm

lance:
Closing in favor of https://github.com/securesign/secure-sign-operator/pull/520

PR #517, opened 2024-07-19
[SECURESIGN-574] | Independently deployable CTlog Changes

tommyd450:


osmman (changes requested):
You have to rebase your PR against main. There has been a changes in [resolve_tree](https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/ctlog/actions/resolve_tree.go) which you will need to modify.

It should be very similar to https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/rekor/actions/server/resolve_tree.go#L67-L80

osmman (approved):

PR #518, opened 2024-07-19
SECURESIGN-1250 | Targets are not coming up with monitoring enabled.

JasonPowr:
Pr to fix the issue where service monitor targets are not coming up, caused by a change in the naming for the service ports.

### DIFF
--- a/internal/controller/common/utils/kubernetes/service.go
+++ b/internal/controller/common/utils/kubernetes/service.go
@@ -12,7 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-func CreateService(namespace string, name string, portName string, port int, labels map[string]string) *corev1.Service {
+func CreateService(namespace string, name string, portName string, port int, targetPort int32, labels map[string]string) *corev1.Service {
 	return &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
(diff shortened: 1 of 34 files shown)


internal/controller/common/utils/kubernetes/service.go
```diff
@@ -26,7 +26,7 @@ func CreateService(namespace string, name string, portName string, port int, lab
 					Name:       portName,
 					Protocol:   corev1.ProtocolTCP,
 					Port:       int32(port),
-					TargetPort: intstr.FromInt(port),
+					TargetPort: intstr.FromInt(targetPort),
```
osmman, line 29:
```suggestion
					TargetPort: intstr.FromInt32(targetPort),
```

osmman: FromInt is deprecated

internal/controller/trillian/actions/logsigner/service.go
```diff
@@ -44,7 +40,7 @@ func (i createServiceAction) Handle(ctx context.Context, instance *rhtasv1alpha1
 	)
 
 	labels := constants.LabelsFor(actions.LogSignerComponentName, actions.LogsignerDeploymentName, instance.Name)
-	logsignerService := k8sutils.CreateService(instance.Namespace, actions.LogsignerDeploymentName, actions.MonitoringPortName, actions.MonitoringPort, labels)
+	logsignerService := k8sutils.CreateService(instance.Namespace, actions.LogsignerDeploymentName, actions.MonitoringPortName, actions.MonitoringPort, actions.MonitoringPort, labels)
```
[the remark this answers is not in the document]
osmman: currently there is only 8090 provided

osmman:
maybe we should rename `80-tcp` to be consistent across services to `http`

osmman (approved):

PR #519, opened 2024-07-22
Fix incorrect number of parameters for SBJ error logging

osmman:
Fixes panic error caused by incorrect number of parameters for error logger function


JasonPowr (approved):
/lgtm

PR #521, opened 2024-07-22
github-action: Store OCI image as artifact

osmman:


bouskaJ:
What about the EKS workflow?

PR #534, opened 2024-07-31
Enforce standards by golangci linter

osmman:


bouskaJ:
Can you add github action workflow for linter execution? Something similar to https://github.com/securesign/sigstore-e2e/blob/main/.github/workflows/validate.yml#L21

osmman:
> Can you add github action workflow for linter execution? Something similar to https://github.com/securesign/sigstore-e2e/blob/main/.github/workflows/validate.yml#L21

I transformed this PR to enforce golangci linter via Github Action. It contains fixes to this project to pass defined rules, each of them has separate commit.

PR #536, opened 2024-08-01
Add proxy e2e test

bouskaJ:


test/e2e/custom_install/proxy.go
```diff
@@ -0,0 +1,498 @@
...
+	ctx := context.TODO()
+
+	var namespace *v1.Namespace
+	var securesign *v1alpha1.Securesign
+	var hostname string
+
+	AfterEach(func() {
+		if CurrentSpecReport().Failed() {
+			if val, present := os.LookupEnv("CI"); present && val == "true" {
+				if val, present := os.LookupEnv("CI"); present && val == "true" {
```
osmman, line 58: this is duplicated if statement

test/e2e/custom_install/proxy.go
```diff
@@ -0,0 +1,498 @@
...
+					Name:       "https",
+					Port:       443,
+					TargetPort: intstr.FromString("https"),
+				},
+			},
+		},
+	}
+}
+
+func createProxyServer(cli runtimeCli.Client, ctx context.Context, hostname string, ns string) {
```
osmman, line 373: please put `ctx` as first parameter. It is common practice in go.

test/e2e/custom_install/proxy.go
```diff
@@ -0,0 +1,498 @@
...
+						PeriodSeconds:       10,
+					},
+				},
+			},
+			ServiceAccountName: "operator-controller-manager",
+		},
+	}
+}
+
+func installOperatorWithProxyConf(cli runtimeCli.Client, ctx context.Context, proxyHostname string, ns string) {
```
osmman, line 489: please put `ctx` as first parameter. It is common practice in go. please put `ctx` as first parameter. It is common practice in go.

test/e2e/custom_install/proxy.go
```diff
@@ -0,0 +1,498 @@
...
+		} else {
+			objects = append(objects, u)
+		}
+	}
+
+	return objects
+}
+func installOperatorPodWithProxyConf(proxyHostname string, ns string) *v1.Pod {
+	image := os.Getenv("TEST_MANAGER_IMAGE")
+	Expect(image).ToNot(BeNil(), "TEST_MANAGER_IMAGE can't be nil")
```
osmman, line 427: `Getenv` return `string` which could not be `nil`, if it is not set it will be default value `""`.  It will be better to use `LookupEnv` which return `(string, bool)`.

test/e2e/custom_install/proxy.go
```diff
@@ -0,0 +1,498 @@
+//go:build custom_install
```
osmman, line 1: currently this file do not pass `linter` check. Please fix the errors. Currently you have to run it manualy because of custom_install tag

osmman (changes requested):

test/e2e/custom_install/custom_install_suite_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/format"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+func TestCustomInstall(t *testing.T) {
```
osmman, line 13: we need to force go test to not run test parallel. it happens because of a new TestCustomInstall and it will to collision on ingress resources.

osmman (changes requested):

test/e2e/custom_install/custom_install_suite_test.go
bouskaJ: `-p 1` in the go test command disable parallel execution

osmman: we need to add that parameter here too https://github.com/securesign/secure-sign-operator/blob/main/Makefile#L133

osmman (approved):
/lgtm

bouskaJ:
rebased

PR #550, opened 2024-08-12
[SECURESIGN-1238] Align with new TUF server

bouskaJ:


api/v1alpha1/tuf_types.go
```diff
@@ -20,6 +20,9 @@ type TufSpec struct {
 	//+kubebuilder:default:={{name: rekor.pub},{name: ctfe.pub},{name: fulcio_v1.crt.pem}}
 	//+kubebuilder:validation:MinItems:=1
 	Keys []TufKey `json:"keys,omitempty"`
+	// PVC configuration
+	//+kubebuilder:default:={size: "5Gi", retain: true}
```
bkabrda, line 24: 100Mi is more than enough for the TUF repository.

bkabrda (commented):
Hi :wave: thanks for working on this. I think overall this looks very reasonable. I think we can reserve much less space for the PVC, a TUF repo, even a very active one, shouldn't need more than 100Mi long-term, so let's decrease the size (I think the size is mentioned on multiple places, so let's change them all).

I'm not approving or disapproving yet, because we need to wait for upstream to review the TUF server PR to see if some changes to it will need to reflected here. I'm also not expert on the operator, I'm just saying that overall this seems to make sense.

internal/controller/tuf/utils/tuf_init_job.go
```diff
@@ -0,0 +1,77 @@
...
+								},
+							},
+						},
+					},
+					Containers: []core.Container{
+						{
+							Name:  "tuf-init",
+							Image: constants.TufImage,
+							Env:   env,
+							Args: []string{
```
bkabrda, line 55: We should provide the -keyssecret flag here to export the TUF signing keys: https://github.com/sigstore/scaffolding/blob/main/cmd/tuf/server/main.go#L53

api/v1alpha1/common.go
```diff
@@ -98,6 +98,10 @@ type Pvc struct {
 	// The name of the StorageClass to claim a PersistentVolume from.
 	//+optional
 	StorageClass string `json:"storageClass,omitempty"`
+	// PVC AccessModes
+	//+kubebuilder:default:={ReadWriteOnce}
+	//+kubebuilder:validation:MinItems:=1
```
osmman, line 103: Please add Enum validation rule with listed options

internal/controller/common/utils/kubernetes/job.go
osmman, line 1: it will be better to move it into separate package like kubernetes/job/job.go. Currently, it collide with other k8s resources. Other option is to add resource type into funtion name like GetJob

internal/controller/common/utils/kubernetes/pvc.go
```diff
@@ -10,7 +10,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func CreatePVC(namespace string, pvcName string, pvcSize resource.Quantity, storageClass string, labels map[string]string) *corev1.PersistentVolumeClaim {
+func CreatePVC(namespace string, pvcName string, pvcSize resource.Quantity, storageClass string, accessModes []corev1.PersistentVolumeAccessMode, labels map[string]string) *corev1.PersistentVolumeClaim {
```
osmman, line 13: The function is starting to have a lot of parameters, so maybe it would be better to pass Pvc object and reduce 4 parameters into single one. Problem with two types could be solved by transforming TufPvc into Pvc.

internal/controller/tuf/actions/constants.go
```diff
@@ -6,4 +6,7 @@ const (
 	RBACName       = "tuf"
 	PortName       = "http"
 	Port           = 8080
+	InitJobName    = "tuf-repository-init"
+
+	RepositoryCondition = "tufRepository"
```
osmman, line 11: you can remove tuf from condition name, it should be clear that it is connected to tuf component from CRD

internal/controller/common/utils/kubernetes/pvc.go
bouskaJ: I am quite unsure about this. There is no safe way transforming one struct to another. Simple modification of the `TufPvc` struct can lead to panic. I would like to avoid it. Holding 4 parameters is still something we can do.

osmman: It thought this dummy method:
```go
pvc := Pvc{
  Name: tufPvc.Name,
  StorageClass: tufPvc,StorageClass,
  AccessModes: tufPvc.AccessModes,
  Size: tufPvc.Size,
}
_ = CreatePVC(namespace, pvc, labels)
```
This transformation of type will be needed only in one place

bouskaJ:
/unhold

ready for review

api/v1alpha1/trillian_types_test.go
osmman, line 1: TrillianSpec is missing defaulter for accessMode like RekorStruct
https://github.com/securesign/secure-sign-operator/blob/a47b60f1f6128c646bffa9f360576954c9ff4071/api/v1alpha1/trillian_types.go#L28

api/v1alpha1/tuf_types.go
```diff
@@ -20,6 +21,39 @@ type TufSpec struct {
 	//+kubebuilder:default:={{name: rekor.pub},{name: ctfe.pub},{name: fulcio_v1.crt.pem},{name: tsa.certchain.pem}}
 	//+kubebuilder:validation:MinItems:=1
 	Keys []TufKey `json:"keys,omitempty"`
+	// Secret object reference that will hold you repository root keys. This parameter will be used only with operator-managed repository.
+	//+kubebuilder:default:=tuf-root-keys
+	RootKeySecretRef string `json:"rootKeySecretRef,omitempty"`
```
osmman, line 26: Please use `LocalObjectReference` or `SecretKeySelector` to reference k8s objects rather than string type. It will make our API more consistent

test/e2e/support/tas/tuf/tuf.go
```diff
@@ -48,3 +56,104 @@ func GetServerPod(ctx context.Context, cli client.Client, ns string) func() *v1.
...
+		return cli.Update(ctx, tufDeployment)
+	},
+	).WithTimeout(1 * time.Second).Should(Succeed())
+
+	// wait for controller to start loop again
+	time.Sleep(5 * time.Second)
+}
+
+func refreshTufJob(instance *v1alpha1.Tuf) *v12.Job {
+	return &v12.Job{
```
osmman, line 100: How much different is this Job structure when compare to actions implementation 
https://github.com/securesign/secure-sign-operator/pull/550/files#diff-674b2405144e8f5172fb07a4e85b8b3c8571ebe4a78a327871b3d8c731967f23R21.

If they are mostly same than we could reuse it, for example like:
```go
func refreshTufJob(instance *v1alpha1.Tuf) *v12.Job {
   job := tufUtils.CreateTufInitJob(....)
   job.GenerateName = "tuf-refresh-",
   // other modification
   return job
}
```

api/v1alpha1/tuf_types.go
bouskaJ: I was trying but I hit problems with default values. Can you suggest me defaulter that would work? I was not successful.

osmman: this should work
```go
	//+kubebuilder:default:={name: tuf-root-keys}
	RootKeySecretRef *LocalObjectReference `json:"rootKeySecretRef,omitempty"`
```

osmman: I think that important part is that LocalObjectReference is pointer

osmman (approved):
/lgtm

PR #557, opened 2024-08-16
SECURESIGN-842 | Operator's metrics are not consumed by OCP monitor

JasonPowr:
# Testing 
To test this, you just need to build and deploy the operator, you should see a new target being created under `Observe -> Targets` called `rhtas-operator-controller-manager-metrics-monitor`, you should then be able to query the metrics of the operator.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -125,6 +125,12 @@ jobs:
           kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml
           kubectl wait --namespace ingress-nginx --for=condition=ready pod --selector=app.kubernetes.io/component=controller --timeout=90s
 
+      - name: Install prometheus
+        run: |
+          LATEST=$(curl -s https://api.github.com/repos/prometheus-operator/prometheus-operator/releases/latest | jq -cr .tag_name)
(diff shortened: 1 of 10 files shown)


JasonPowr:
@osmman @bouskaJ Quick question, I have been thinking about this a bit, like with the ci we needed to install Prometheus prior to installing the operator so it has a definition of the service monitor type, will this cause an issue? I believe it will, as it would fail to install on clusters without an install of Prometheus.

Instead should this be tied to the securesign resource, as in:
```
spec:
  monitoring:
    enabled: true
  rekor:
    externalAccess:
    ......
``` 
then if enabled create the service monitors, service etc ?

osmman:
> @osmman @bouskaJ Quick question, I have been thinking about this a bit, like with the ci we needed to install Prometheus prior to installing the operator so it has a definition of the service monitor type, will this cause an issue? I believe it will, as it would fail to install on clusters without an install of Prometheus.
> 
> Instead should this be tied to the securesign resource, as in:
> 
> ```
> spec:
>   monitoring:
>     enabled: true
>   rekor:
>     externalAccess:
>     ......
> ```
> 
> then if enabled create the service monitors, service etc ?

I think that at this moment we can live with it, for example a lot of operators depends on cert-manager to make their admission webhooks to work correctly. So unfortunatly right not it is prerequisity. Definatly it will be prefered to make it optional.

/lgtm

JasonPowr:
@osmman Can I get another /lgtm pls :)

osmman (approved):

PR #493, opened 2024-07-05
[SECURESIGN-994] Add TLS to Rekor and Trillian services

fghanmi:

PR #558, opened 2024-08-17
[SECURESIGN-994] Add TLS support for trillian db

fghanmi:


osmman (changes requested):
I do not like to that overusage of TlS word in the API.
```
spec:
  serverTLS:
    tls:
      privateKeyRef:
```

Make new sections for signer and server which will contains all related configs for that sub component. Here is example of API
```
spec:
  server:
    tls:
  signer:
    tls:
  database:
    tls:     
```

I prefer this approach because next release we can simply extend these section by adding replication or other options to configure HA.

api/v1alpha1/trillian_types.go
```diff
@@ -27,6 +27,12 @@ type TrillianSpec struct {
 	//+kubebuilder:validation:XValidation:rule=((!self.create && self.databaseSecretRef != null) || self.create),message=databaseSecretRef cannot be empty
 	//+kubebuilder:default:={create: true, pvc: {size: "5Gi", retain: true}}
 	Db TrillianDB `json:"database,omitempty"`
+	//+optional
+	TrillianServerTLS TrillianServerTLS `json:"serverTLS,omitempty"`
+	//+optional
+	TrillianSignerTLS TrillianSignerTLS `json:"signerTLS,omitempty"`
```
osmman, line 33: Why there is a TLS setting for signer when it do not provide any service which could be secured by TLS?

fghanmi: because the signer is communicating with the DB that could has TLS enabled

osmman: ok so these are only for clients to trust db's server cert. Than we should design it a differently move that setting into same place where is our database connection options and use this one place for server and signer.

osmman:
> I do not like to that overusage of TLS word in the API.
> 
> ```
> spec:
>   serverTLS:
>     tls:
>       privateKeyRef:
> ```
> 
> Make new sections for signer and server which will contains all related configs for that sub component. Here is example of API
> 
> ```
> spec:
>   server:
>     tls:
>   signer:
>     tls:
>   database:
>     tls:     
> ```
> 
> I prefer this approach because next release we can simply extend these section by adding replication or other options to configure HA.

Due received a new information from https://github.com/securesign/secure-sign-operator/pull/558#discussion_r1722960170 comment here is a new proposal which has only single point of configuration:

```
spec:
  database:
    tls:
```

api/v1alpha1/trillian_types.go
fghanmi: in this PR, yes, but in other PR (https://github.com/securesign/secure-sign-operator/pull/493/files#diff-33ed5299c8ab77e1be584ffd6faca7ef31c9053d7bd30d6f83aaeaaf7acf7e06R129)
Trillian logserver needs also to enable its TLS (using its cert/key) 

in TLSCert struct, we can use only ca cert, or only cert/key, or both.

internal/controller/trillian/utils/config_map.go
```diff
@@ -0,0 +1,85 @@
+package trillianUtils
```
osmman, line 1: this file should be in actions package

internal/controller/trillian/actions/db/deployment.go
```diff
@@ -49,7 +50,9 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Trilli
 		scc = &v1.PodSecurityContext{FSGroup: utils.Pointer(int64(1001)), FSGroupChangePolicy: utils.Pointer(v1.FSGroupChangeOnRootMismatch)}
 	}
 
-	db, err := trillianUtils.CreateTrillDb(instance, actions.DbDeploymentName, actions.RBACName, scc, labels)
+	signingKeySecret, _ := k8sutils.GetSecret(i.Client, "openshift-service-ca", "signing-key")
```
osmman, line 53: I am not sure if I understand correctly usage of: 
```
signingKeySecret, _ := kubernetes.GetSecret(cli, "openshift-service-ca", "signing-key")
```
in the code. Is it used to check if operator is running in OCP? If yes than please use [IsOpenshift](https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/common/utils/kubernetes/common.go#L76) method

osmman: I am even surprised that your operator has access to that resource

internal/controller/trillian/actions/logserver/deployment.go
```diff
@@ -63,6 +65,105 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Trilli
...
+			})
+	} else {
+		i.Logger.V(1).Info("Communication between services is insecure")
+	}
+
+	if instance.Spec.TrillianServerTLS.TLSCertificate.CertRef != nil || signingKeySecret != nil {
+		server.Spec.Template.Spec.Containers[0].VolumeMounts = append(server.Spec.Template.Spec.Containers[0].VolumeMounts,
+			corev1.VolumeMount{
+				Name:      "tls-cert",
+				MountPath: "/etc/ssl/certs",
```
osmman, line 160: /etc/ssl/certs  will not work, our images are rhel based and they used a different location of ca trusted source. Currently it works because we already OCP default ca-bundel generated in namespace. 

https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/trillian/actions/logsigner/deployment.go#L50
https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/trillian/actions/logserver/deployment.go#L49

https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/common/utils/set_trusted_ca.go#L12-L34

osmman: so we can remove this injection of certs and just depends on existing implementation which is already present and works

api/v1alpha1/trillian_types.go
osmman: What is purpose of CACertRef? I understand PrivateKeyRef and CertRef are used to enable enription on server side.

PR #561, opened 2024-08-19
[SECURESIGN-574] Independently Deployable CT-log

tommyd450:
After testing and signing and veryfying it should be up to scrath and working now. 

To test you it is advised ctlog and trillian not be in the same namespace as the main change is to check that ctlog can successfully reach trillian elsewhere.


internal/controller/ctlog/ctlog_controller_test.go
```diff
@@ -115,12 +115,12 @@ var _ = Describe("CTlog controller", func() {
 				return meta.IsStatusConditionPresentAndEqual(found.Status.Conditions, constants.Ready, metav1.ConditionFalse)
 			}).Should(BeTrue())
 
-			By("Pending phase until Trillian svc is created")
+			/*By("Pending phase until Trillian svc is created")
```
osmman, line 118: you can remove it from source code.

osmman (approved):
/lgtm

osmman:
please squash commits into single one

PR #563, opened 2024-08-19
chore: patch-tsa-upgrade-test

JasonPowr:
Small fix the upgrade test, adds a retry to the GetTSACertificateChain() func

### DIFF
--- a/test/e2e/byodb_test.go
+++ b/test/e2e/byodb_test.go
@@ -164,8 +164,10 @@ var _ = Describe("Securesign install with byodb", Ordered, func() {
 
 			tsa := tas.GetTSA(ctx, cli, namespace.Name, securesign.Name)()
 			Expect(tsa).ToNot(BeNil())
-			err := tas.GetTSACertificateChain(ctx, cli, tsa.Namespace, tsa.Name, tsa.Status.Url)
-			Expect(err).ToNot(HaveOccurred())
+
+			Eventually(func() error {
+				return tas.GetTSACertificateChain(ctx, cli, tsa.Namespace, tsa.Name, tsa.Status.Url)
+			}).Should(Succeed())
 
 			oidcToken, err := support.OidcToken(ctx)
 			Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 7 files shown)


test/e2e/byodb_test.go
```diff
@@ -164,7 +164,7 @@ var _ = Describe("Securesign install with byodb", Ordered, func() {
 
 			tsa := tas.GetTSA(ctx, cli, namespace.Name, securesign.Name)()
 			Expect(tsa).ToNot(BeNil())
-			err := tas.GetTSACertificateChain(ctx, cli, tsa.Namespace, tsa.Name, tsa.Status.Url)
+			err := tas.GetTSACertificateChain(ctx, cli, tsa.Namespace, tsa.Name, tsa.Status.Url, 5)
```
osmman, line 167: I think that it is better to use `Eventualy` function. Example: 
```go
Eventually(func() error {
  return tas.GetTSACertificateChain(ctx, cli, tsa.Namespace, tsa.Name, tsa.Status.Url)
}).To(Succeed())
```

test/e2e/support/tas/tsa.go
```diff
@@ -53,15 +54,28 @@ func GetTSA(ctx context.Context, cli client.Client, ns string, name string) func
...
 }
 
-func GetTSACertificateChain(ctx context.Context, cli client.Client, ns string, name string, url string) error {
-	req, err := http.NewRequestWithContext(ctx, "GET", url+"/api/v1/timestamp/certchain", nil)
-	if err != nil {
-		return fmt.Errorf("failed to create HTTP request: %w", err)
+func GetTSACertificateChain(ctx context.Context, cli client.Client, ns string, name string, url string, retries int) error {
+	var resp *http.Response
+	var err error
+	for i := 0; i < retries; i++ {
```
osmman, line 60: right now you can remove retries logic because it is handled by eventually

JasonPowr: Yes ofc, thank you I was not thinking :)

PR #564, opened 2024-08-20
fix: check observed, available and new replica set statuses to decide if deployment is ready

osmman:
… if deployment is ready


JasonPowr (approved):
/lgtm

PR #582, opened 2024-08-29
fix(rekor): create new secret only for generated signer key

osmman:
It fixes:
1. create new immutable secret only for generated signer key
2. identify changes in spec.signer.kms and apply them
3. ignore privateKey if kms is not empty 
4. simplify generation function signer key, because it will newer generate key by password


JasonPowr (approved):
/lgtm

PR #587, opened 2024-09-02
fix: Make sure SBJ installation runs only once

JasonPowr:
Small bug fix, I noticed sometimes SBJ would run twice on re-installation, as in two jobs would run at the same time, this negates the issue

### DIFF
--- a/internal/controller/securesign/actions/segment_backup_job.go
+++ b/internal/controller/securesign/actions/segment_backup_job.go
@@ -16,7 +16,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	rhtasv1alpha1 "github.com/securesign/operator/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


internal/controller/securesign/actions/segment_backup_job.go
```diff
@@ -37,8 +36,8 @@ func (i segmentBackupJob) Name() string {
 }
 
 func (i segmentBackupJob) CanHandle(_ context.Context, instance *rhtasv1alpha1.Securesign) bool {
-	if c := meta.FindStatusCondition(instance.Status.Conditions, SegmentBackupJobName); c != nil {
-		return c.Reason != constants.Ready
+	if c := meta.FindStatusCondition(instance.Status.Conditions, SBJCondition); c != nil {
```
osmman, line 39: you can simplify it to
```
return !meta.IsStatusConditionTrue(instance.Status.Conditions, SBJCondition)
```

Benefit is that it will handle nil value and both **False** and **Unknow** value

JasonPowr: Please correct me if I am wrong, but If I make this change, I need to move this line to the bottom of the function, which wont work as if the annotation is there and set to true the can handle function will continuously return true eg:

```
func (i segmentBackupJob) CanHandle(_ context.Context, instance *rhtasv1alpha1.Securesign) bool {
	val, found := instance.Annotations[annotations.Metrics]
	if !found {
		return true
	}
	if boolVal, err := strconv.ParseBool(val); err == nil {
		return boolVal
	}
	return !meta.IsStatusConditionTrue(instance.Status.Conditions, SBJCondition)
}
```

osmman: Sorry, you are right. I completely missed whole logic with annotation.

PR #592, opened 2024-09-03
[SECURESIGN-994] Add TLS support for trillian db #558

fghanmi:
Replaces: https://github.com/securesign/secure-sign-operator/pull/558 (e2e corrupted)


internal/controller/trillian/utils/tls.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// external DB
+	if !utils.IsEnabled(instance.Spec.Db.Create) && instance.Spec.TrustedCA != nil {
+		return true
+	}
+
+	return false
+}
+
+func CAPath(ctx context.Context, cli client.Client, instance *rhtasv1alpha1.Trillian) (string, error) {
```
bouskaJ, line 32: It would be nice if the deployment could leverage system configuration https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/common/utils/set_trusted_ca.go#L22 and we could remove this function.

internal/controller/trillian/utils/trillian-db.go
```diff
@@ -49,6 +56,10 @@ func CreateTrillDb(instance *v1alpha1.Trillian, dpName string, sa string, secCon
 					},
 					Containers: []core.Container{
 						{
+							Command: []string{
+								"run-mysqld",
+							},
+							Args:  []string{},
```
bouskaJ, line 62: I guess, there is no need to overwrite the default command. Is it?

test/e2e/byodb_test.go
```diff
@@ -151,7 +154,43 @@ var _ = Describe("Securesign install with byodb", Ordered, func() {
...
+	if err != nil {
+		return err
+	}
+
+	if kubernetes.IsOpenShift() {
+		if mysql.Annotations == nil {
+			mysql.Annotations = make(map[string]string)
+		}
+		mysql.Annotations["service.beta.openshift.io/serving-cert-secret-name"] = "my-trillian-db-tls-secret"
+		err := cli.Update(ctx, mysql)
```
bouskaJ, line 187: Move annotation configuration to the Service creation phase. Updates may fail on `the object has been modified; please apply your changes to the latest version and try again ` error. Better to avoid updates.

internal/controller/trillian/actions/logserver/deployment.go
```diff
@@ -63,6 +70,31 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Trilli
...
+		}
+
+		server.Spec.Template.Spec.Containers[0].Args = append(server.Spec.Template.Spec.Containers[0].Args, "--mysql_tls_ca", caPath)
+		mysql_server_name := "$(MYSQL_HOSTNAME)." + instance.Namespace + ".svc"
+		if !*instance.Spec.Db.Create {
+			mysql_server_name = "$(MYSQL_HOSTNAME)"
+		}
+		server.Spec.Template.Spec.Containers[0].Args = append(server.Spec.Template.Spec.Containers[0].Args, "--mysql_server_name", mysql_server_name)
+	}
+
```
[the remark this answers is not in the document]
fghanmi: @bouskaJ 
CreateTrillDeployment creates the deployment for both trillian log server and signer, in TLS, these two have different configuration, actually, only logserver needs to enabled TLS as a server (not in the PR, we need it here https://github.com/securesign/secure-sign-operator/pull/493)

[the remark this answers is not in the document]
fghanmi: Also, I am not sure if it's possible to handle errors in the CreateTrillDeployment, as we do in this file where we have the Action.

internal/controller/trillian/actions/logserver/deployment.go
```diff
@@ -63,6 +70,31 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Trilli
...
+				Message: err.Error(),
+			})
+			i.StatusUpdate(ctx, instance)
+			if apiErrors.IsNotFound(err) {
+				return i.Requeue()
+			}
+			return i.Failed(err)
+		}
+
+		server.Spec.Template.Spec.Containers[0].Args = append(server.Spec.Template.Spec.Containers[0].Args, "--mysql_tls_ca", caPath)
```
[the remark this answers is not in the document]
fghanmi: in trillian: https://github.com/google/trillian/blob/master/storage/mysql/provider.go#L38
only `--mysql_tls_ca` is recognized (not `SSL_CERT_DIR`)

internal/controller/trillian/utils/trillian-db.go
fghanmi: In case we don't, we'll have this error:
```
>> oc logs trillian-db-69dcf6f96c-hf8ks
/usr/bin/container-entrypoint: line 2: exec: --: invalid option
exec: usage: exec [-cl] [-a name] [command [argument ...]] [redirection ...]
```
Before, we weren't using it because we haven't any args.

internal/controller/trillian/actions/logserver/deployment.go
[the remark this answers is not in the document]
osmman: I found that there is an option `tls=true` in mysql driver which by `crypto/tls` doc should use host's root CA

```
// RootCAs defines the set of root certificate authorities
// that clients use when verifying server certificates.
// If RootCAs is nil, TLS uses the host's root CA set.
RootCAs *[x509](https://pkg.go.dev/crypto/x509).[CertPool](https://pkg.go.dev/crypto/x509#CertPool)
``` 

https://github.com/go-sql-driver/mysql/blob/master/dsn.go#L175 
https://pkg.go.dev/crypto/tls#Config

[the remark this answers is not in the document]
osmman: I tried it and it looks like that `tcp(trillian-mysql.demo.svc:3306)/database?tls=true` is using RootCAs from `SSL_CERT_DIR`

internal/controller/trillian/actions/logserver/deployment.go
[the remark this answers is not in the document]
bouskaJ: I see that each deployment have its own set of modifications. I would consider creating `CreateLogServerDeployment` and `CreateLogSignerDeployment` functions that would fit to the desired state. More code but it would be much easier for readability and maintenance. But of course it is up to you.

internal/controller/trillian/utils/trillian-db.go
bouskaJ: Ok, I see there is something like 
```
ENTRYPOINT ["container-entrypoint"]
CMD ["run-mysqld"]
``` at the end of mariaDB dockerfile that cause it.

File: test/e2e/byodb_test.go [hunk 6]
```diff
@@ -18,6 +20,7 @@ import (
 )
 
 var _ = Describe("Securesign install with byodb", Ordered, func() {
+	utils.BoolFlagOrEnv(&constants.Openshift, "openshift", "OPENSHIFT", false, "Enable to ensures the operator applies OpenShift specific configurations.")
```
bouskaJ, line 23: Just one note, otherwise looks good.

Better to use `IsOpenshift` function because user can forget to specify the ENV variable

fghanmi: Actually, `IsOpenshift` is used here, this line is just to fill the &constants.Openshift from OPENSHIFT.

bouskaJ: oh, it is not working OOTB. got it

fghanmi:
@bouskaJ , @osmman , is it okay to merge ? 
Thanks!

PR #602, opened 2024-09-05
Migrate to tough

bouskaJ:


osmman (approved):

api/v1alpha1/tuf_types.go
```diff
@@ -59,7 +59,7 @@ type TufPvc struct {
 type TufKey struct {
 	// File name which will be used as TUF target.
 	//+required
-	//+kubebuilder:validation:Pattern:="^[-._a-zA-Z0-9]+$"
+	// +kubebuilder:validation:Enum:=rekor.pub;ctfe.pub;fulcio_v1.crt.pem;tsa.certchain.pem
```
bkabrda, line 62: So IIUC this can be changed in upcoming releases without larger issues, is that correct? The reason I'm asking is that we might do the full "TUF repo auto management" feature and hence this would likely change again.

bouskaJ: yes, from my experience there is no issue with changing validation rules.

bouskaJ: @osmman do you see any possible issue?

bouskaJ: According to what I understand from @bkabrda files needs to be strictly named because the cosign CLI expect that. In other words, If you put different name the cosign CLI will not work as expected. I would say it is more a bug-fix.

PR #604, opened 2024-09-06
SECURESIGN-1342 | Unable to delete upgraded Securesign object

JasonPowr:
The validation rules I removed here were stopping deletion after an upgrade, I tried to configure new rules however it just would not work reliable, validation for this particular rule is still done inside the controller here: https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/tsa/utils/tsa_cert_chain.go#L238


osmman:
Hi, I am thinking about making tsa optional in Securesign resource, by changing it to pointer. https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/securesign_types.go#L35

```
TimestampAuthority *TimestampAuthoritySpec `json:"tsa,omitempty"`
```

@JasonPowr @bouskaJ What do you think?

JasonPowr:
> Hi, I am thinking about making tsa optional in Securesign resource, by changing it to pointer. https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/securesign_types.go#L35
> 
> ```
> TimestampAuthority *TimestampAuthoritySpec `json:"tsa,omitempty"`
> ```
> 
> @JasonPowr @bouskaJ What do you think?

Off the Top of my head, I can't see any major issue with this, making it a pointer should allow for a nil value so I should be able to do some better api validation with it, but I will let @bouskaJ comment before making changes.

bouskaJ:
yes, I don't see any issue. @JasonPowr go ahead.

osmman (approved):

PR #613, opened 2024-09-11
SECURESIGN-1015 | Make sure route selector labels and managed annos are recconciled

JasonPowr:


bouskaJ:
Do we need to do this by operator? Resource labels are not updated by `Ensure` function at all so user can freely modify them. This is just first question that popped up in my head. Maybe I am missing something.

JasonPowr:
> Do we need to do this by operator? Resource labels are not updated by `Ensure` function at all so user can freely modify them. This is just first question that popped up in my head. Maybe I am missing something.

@osmman @bouskaJ, You make a good point, I just tested this to 100% confirm and adding the labels manually works without issue. I guess the only thing we loose here by not doing it is the convince of being able to configure it through the API. I suppose it may also be confusing for the customer, if they see the option to configure routeSelectorLabels for ingress sharding  they might assume they can add it through the api and update exsisting ingress and route objects.

@osmman WDYT?

JasonPowr:
@osmman @bouskaJ When you guys have a minute can you take a look? thanks

internal/controller/common/action/base_action.go
```diff
@@ -142,3 +139,139 @@ func (action *BaseAction) Ensure(ctx context.Context, obj client2.Object) (bool,
...
+		// DeepEqual does not honor default values
+		if !equality.Semantic.DeepDerivative(expectedSpec.Interface(), currentSpec.Interface()) {
+			currentSpec.Set(expectedSpec)
+		}
+		return nil
+	}
+}
+
+func EnsureRouteSelectorLabels(managedLabels ...string) EnsureOption {
+	return func(current client.Object, expected client.Object) error {
```
bouskaJ, line 169: I would convert both objects to `Ingress` or `Route` or throw error. You can avoid reflrection.

internal/controller/common/action/base_action.go
```diff
@@ -142,3 +139,139 @@ func (action *BaseAction) Ensure(ctx context.Context, obj client2.Object) (bool,
...
+			expectedLabels = map[string]string{}
+		}
+		currentLabels := current.GetLabels()
+		if currentLabels == nil {
+			currentLabels = map[string]string{}
+		}
+
+		for label := range currentLabels {
+			if _, ok := expectedLabels[label]; !ok {
+				delete(currentLabels, label)
```
bouskaJ, line 211: you can create empty map and add only managed labels

internal/controller/common/action/base_action.go
```diff
@@ -142,3 +139,139 @@ func (action *BaseAction) Ensure(ctx context.Context, obj client2.Object) (bool,
...
+			expectedAnno = map[string]string{}
+		}
+		currentAnno := current.GetAnnotations()
+		if currentAnno == nil {
+			currentAnno = map[string]string{}
+		}
+
+		for anno := range currentAnno {
+			if _, ok := expectedAnno[anno]; !ok {
+				delete(currentAnno, anno)
```
bouskaJ, line 237: I am not sure, we want to delete all k8s generated labels/annotations. Maybe we should touch `managedAnnotations` (`managedLabels`) only.

internal/controller/common/action/base_action.go
JasonPowr: If I understand you correctly, I cant convert to Ingress or Route objects here, as an example I will use fulcio, when the resource is fully deployed, and I make a change to the route selector labels fulcio does not reconcile again due to https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/securesign/actions/ensure_fulcio.go#L52 return updated to false, I believe this to be a similar issue to  deepequal vs DeepDerivative, I make sure the update is picked up here https://github.com/securesign/secure-sign-operator/blob/457bd3faa49b93eb11b7bab3cda5f92876481642/internal/controller/common/action/base_action.go#L180

after this conversion would take place

internal/controller/common/action/base_action.go
osmman: We need to cover these scenarios in test:
1. remove managed label if it is not present in expected object
2. label is not managed so it should be ignored by any changes in expected object

Input:
```
EnsureLabels("managed")(generatrion1, generation2)
```

Expected result:
```yaml
metadata:
  generation: 1
  labels:
    managed: "values"
    unmanaged: "value"
    
---

# it is expected to remove managed label
metadata:
  generation: 2
  labels:
    unmanaged: "value"
```

internal/controller/common/action/base_action.go
osmman: yes we should touch only `managedAnnotations` and `manageLabels`. 
https://github.com/securesign/secure-sign-operator/pull/613#discussion_r1766612816

JasonPowr:
@bouskaJ @osmman This is good for a re-review, when you guys have a minute, thanks :)

osmman (approved):

PR #617, opened 2024-09-13
SBJ | Make sure SBJ reconciles after upgrade

JasonPowr:
SBJ was not getting updated when the operator was getting updated, this fixes that


internal/controller/securesign/actions/constants.go
```diff
@@ -7,7 +7,7 @@ const (
 	RekorCondition           = "RekorAvailable"
 	TrillianCondition        = "TrillianAvailable"
 	CTlogCondition           = "CTlogAvailable"
-	SBJCondition             = "SBJCondition"
```
osmman, line 10: Why do you not use `MetricsCondition` for segment_backup_job action? It is condition already used by cronjob and rbac action.

osmman (approved):

PR #618, opened 2024-09-13
Unifying CLI Servers

tommyd450:
All of our CLI Servers are now one again, tuftool lacks multiarch so I added the entry manually rather than using the method as it would add multiple empty entries to the CLI Downloads Page on cluster.


internal/clidownload/component.go
```diff
@@ -78,9 +76,29 @@ func (c *Component) Start(ctx context.Context) error {
 			"gitsign":         "gitsign is a CLI tool that allows you to digitally sign and verify git commits.",
 			"ec":              "Enterprise Contract CLI. Set of commands to help validate resources with the Enterprise Contract.",
 			"fetch-tsa-certs": "fetch-tsa-certs is a cli used to configure the kms and tink signer types for Timestamp Authority.",
+			"createtree":     "create-tree is a CLI tool which is used for creating new trees within trillian.",
+			"updatetree":     "update-tree is a CLI tool which is used for managing existing tress within trillian.",
```
JasonPowr, line 80: These two lines are causing the linter to fail, you need to either save the file so the will be formatted correctly, or run `gofmt -s -w internal/clidownload/component.go` which should also do the trick.

cmd/main.go
```diff
@@ -116,9 +116,7 @@ func main() {
 	utils.StringFlagOrEnv(&constants.TufImage, "tuf-image", "TUF_IMAGE", constants.TufImage, "The image used for TUF.")
 	utils.StringFlagOrEnv(&constants.CTLogImage, "ctlog-image", "CTLOG_IMAGE", constants.CTLogImage, "The image used for ctlog.")
 	utils.StringFlagOrEnv(&constants.HttpServerImage, "client-server-image", "CLIENT_SERVER_IMAGE", constants.HttpServerImage, "The image used to serve our cli binary's.")
-	utils.StringFlagOrEnv(&constants.ClientServerImage_cg, "client-server-cg-image", "CLIENT_SERVER_CG_IMAGE", constants.ClientServerImage_cg, "The image used to serve cosign and gitsign.")
-	utils.StringFlagOrEnv(&constants.ClientServerImage_re, "client-server-re-image", "CLIENT_SERVER_RE_IMAGE", constants.ClientServerImage_re, "The image used to serve rekor-cli and the ec binary.")
-	utils.StringFlagOrEnv(&constants.ClientServerImage_f, "client-server-f-image", "CLIENT_SERVER_F_IMAGE", constants.ClientServerImage_f, "The image used to serve fetch-tsa-certs binary.")
+	utils.StringFlagOrEnv(&constants.ClientServerImage, "client-server-cg-image", "CLIENT_SERVER_CG_IMAGE", constants.ClientServerImage, "The image used to serve cosign and gitsign.")
```
JasonPowr, line 119: we should updating the naming here, `client-server-cg-image` should be client-server-image. I would also update 

```
utils.StringFlagOrEnv(&constants.HttpServerImage, "client-server-image", "CLIENT_SERVER_IMAGE", constants.HttpServerImage, "The image used to serve our cli binary's.")
```

to 
```
utils.StringFlagOrEnv(&constants.HttpServerImage, "http-server-image", "HTTP_SERVER_IMAGE", constants.HttpServerImage, "A http image used for TUF")
```

internal/clidownload/component.go
```diff
@@ -78,9 +76,29 @@ func (c *Component) Start(ctx context.Context) error {
...
 		} {
 			obj = append(obj, c.createConsoleCLIDownload(ns.Name, name, protocol+ingress.Spec.Rules[0].Host, description, labels))
 		}
+		x := &consolev1.ConsoleCLIDownload{
```
JasonPowr, line 84: Can we use a better name here than x

tommyd450: yes sir we can

osmman:
Please this code into cli e2e test. It will dump state of common objects from namespace which could be useful to find root of problem.

```
AfterEach(func() {
    if CurrentSpecReport().Failed() && support.IsCIEnvironment() {
	    support.DumpNamespace(ctx, cli, "trusted-artifact-signer")
    }
})
```

osmman:
@tommyd450 I see this error message in events:
```
'Failed to apply default image tag "quay.io/securesign/client-server@sha256:8d98cc4648153c1aebbf2772139d05e560a150bf5070fd8e49057b2ac6dab67":
  couldn''t parse image reference "quay.io/securesign/client-server@sha256:8d98cc4648153c1aebbf2772139d05e560a150bf5070fd8e49057b2ac6dab67":
  invalid checksum digest length'
```

osmman (approved):

PR #622, opened 2024-09-16
test: add namespaced custom installation test

osmman:


bouskaJ:
`custom_install` means custom **operator** installation (for example enhanced with proxy ENV). I think your test should go under `integration` test because it is executable on regular operator installation

PR #627, opened 2024-09-17
Update Ensure to use CreateOrUpdate function

bouskaJ:


JasonPowr (approved):
/lgtm

internal/controller/common/action/base_action.go
```diff
@@ -90,68 +90,52 @@ func (action *BaseAction) Requeue() *Result {
...
-	currentSpec.Set(expectedSpec)
-	action.Logger.Info("Updating object",
-		"kind", reflect.TypeOf(currentObj).Elem().Name(), "Namespace", key.Namespace, "Name", key.Name)
-	if err := action.Client.Update(ctx, currentObj); err != nil {
+		return nil
+	})
+
+	if err != nil {
 		if strings.Contains(err.Error(), OptimisticLockErrorMsg) {
 			return action.Ensure(ctx, obj)
```
osmman, line 135: this could lead to stack overflow error due recursion we should use `https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict`

bouskaJ: yes, I was looking for that function. I remember that I have seen it but I could not find it so I kept the previous implementation.

osmman (approved):

PR #655, opened 2024-09-27
SECURESIGN-1405 | Fulcio key-rotation procedure does not work as expected

JasonPowr:


internal/controller/fulcio/actions/generate_cert.go
```diff
@@ -54,6 +56,9 @@ func (g handleCert) Handle(ctx context.Context, instance *v1alpha1.Fulcio) *acti
 		)
 		return g.StatusUpdate(ctx, instance)
 	}
+	meta.FindStatusCondition(instance.Status.Conditions, CertCondition).Reason = constants.Creating
+	g.StatusUpdate(ctx, instance)
```
bouskaJ, line 60: I am not sure that this is really the solution. I check the code and I do see following scenario that cause this issue:

1. two reconcile requests are queued
2. first one is processed - the secret is created, status update at the end is done
3. second one start processing **before** the status update is propagated to the cache

as a result we have two certificates

With your PR, you make the status update **before** the secret is created so you lower the probability of the issue but it is still possible to hit it.

bouskaJ: I did not find anything better.
/lgtm

PR #658, opened 2024-09-27
fix: recreate ctlog config when spec has changed

osmman:
related to [SECURESIGN-1227](https://issues.redhat.com//browse/SECURESIGN-1227)

CT log server config has not been updated when spec has changed. Because that configuration depends mostly on all fields in spec I solved that problem by comparing latest observed generation for Ready phase. `serverConfig` action will be executed on Ready phase when ObservedGeneration is different than Generation of resource.

### DIFF
--- a/internal/controller/ctlog/actions/server_config.go
+++ b/internal/controller/ctlog/actions/server_config.go
@@ -41,6 +41,8 @@ func (i serverConfig) CanHandle(_ context.Context, instance *rhtasv1alpha1.CTlog
 		return true
 	case instance.Spec.ServerConfigRef != nil:
 		return !equality.Semantic.DeepEqual(instance.Spec.ServerConfigRef, instance.Status.ServerConfigRef)
+	case c.Reason == constants.Ready:
+		return instance.Generation != c.ObservedGeneration
 	default:
 		return false
 	}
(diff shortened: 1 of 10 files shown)


bouskaJ:
yes, I like it. We should apply the same pattern to other components

internal/controller/ctlog/actions/server_config.go
```diff
@@ -101,14 +111,25 @@ func (i serverConfig) Handle(ctx context.Context, instance *rhtasv1alpha1.CTlog)
...
 		})
 		return i.FailedWithStatusUpdate(ctx, fmt.Errorf("could not create CTLog configuration: %w", err), instance)
 	}
 
+	if instance.Status.ServerConfigRef != nil {
+		currentSecret, err := utils.GetSecret(i.Client, instance.Namespace, instance.Status.ServerConfigRef.Name)
+		if client.IgnoreNotFound(err) != nil {
+			return i.Failed(err)
+		}
+		if currentSecret != nil && equality.Semantic.DeepEqual(cfg, currentSecret.Data) {
```
bouskaJ, line 128: Generated configs contains **salted** public key. It will be never equal.

osmman: I completely forget that private keys has to be encrypted by password. So I will remove this and generate new config for every generation.

PR #661, opened 2024-10-03
Resolve SECURESIGN-1405

bouskaJ:


internal/controller/fulcio/actions/generate_cert.go
```diff
@@ -125,11 +118,36 @@ func (g handleCert) Handle(ctx context.Context, instance *v1alpha1.Fulcio) *acti
...
+			Message: err.Error(),
+		})
+		meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+			Type:    constants.Ready,
+			Status:  metav1.ConditionFalse,
+			Reason:  constants.Pending,
+			Message: "Resolving keys",
+		})
+		g.StatusUpdate(ctx, instance)
+		// swallow error and retry
```
JasonPowr, line 142: Only something small, but should we log the error here? People might be more inclined to check the controller logs, before checking the status? just a passing thought

JasonPowr (commented):
Seems fine so far

internal/controller/fulcio/actions/generate_cert.go
```diff
@@ -40,10 +39,8 @@ func (g handleCert) Name() string {
 
 func (g handleCert) CanHandle(_ context.Context, instance *v1alpha1.Fulcio) bool {
 	c := meta.FindStatusCondition(instance.Status.Conditions, constants.Ready)
-	cert := meta.FindStatusCondition(instance.Status.Conditions, CertCondition)
-	return (c.Reason == constants.Pending || c.Reason == constants.Ready) && (instance.Status.Certificate == nil ||
-		!equality.Semantic.DeepDerivative(instance.Spec.Certificate, *instance.Status.Certificate)) &&
-		cert.Reason != constants.Creating
+	return (c.Reason == constants.Pending || c.Reason == constants.Ready) && (!meta.IsStatusConditionTrue(instance.GetConditions(), CertCondition) ||
```
osmman, line 42: can you please transform it to switch statement for better readability. Example:
```go
c := meta.FindStatusCondition(instance.Status.Conditions, constants.Ready)

switch {
case c == nil:
return false
case !(c.Reason == constants.Pending || c.Reason == constants.Ready):
return false
case !meta.IsStatusConditionTrue(instance.GetConditions(), CertCondition):
return true
default:
return !equality.Semantic.DeepDerivative(instance.Spec.Certificate, *instance.Status.Certificate)
}
```

internal/controller/fulcio/actions/generate_cert.go
```diff
@@ -75,42 +74,36 @@ func (g handleCert) Handle(ctx context.Context, instance *v1alpha1.Fulcio) *acti
...
+	//Check if a secret for the  fulcio cert already exists and validate
+	partialSecret, err := k8sutils.FindSecret(ctx, g.Client, instance.Namespace, FulcioCALabel)
 	if err != nil && !apierrors.IsNotFound(err) {
 		g.Logger.Error(err, "problem with finding secret", "namespace", instance.Namespace)
 	}
 
-	if secret != nil {
-		if err := constants.RemoveLabel(ctx, secret, g.Client, FulcioCALabel); err != nil {
+	if partialSecret != nil {
+		if maps.Equal(g.certMatchingAnnotations(instance), partialSecret.GetAnnotations()) {
```
osmman, line 90: I think that we should check if it is derivate. The secret can have more annotations which could be added by other managers.

internal/controller/fulcio/actions/generate_cert.go
```diff
@@ -240,3 +231,68 @@ func (g handleCert) setupCert(ctx context.Context, instance *v1alpha1.Fulcio) (*
...
+		instance.Status.Certificate.CARef = &v1alpha1.SecretKeySelector{
+			Key: "cert",
+			LocalObjectReference: v1alpha1.LocalObjectReference{
+				Name: secret.Name,
+			},
+		}
+	}
+}
+
+func (g handleCert) calculateHostname(ctx context.Context, instance *v1alpha1.Fulcio) error {
```
osmman, line 264: please reformat this method to better if statements readability. Example:
```go
func (g handleCert) calculateHostname(ctx context.Context, instance *v1alpha1.Fulcio) (string, error) {
	if instance.Status.Certificate.CommonName != "" {
		return instance.Status.Certificate.CommonName, nil
	}

	if !instance.Spec.ExternalAccess.Enabled {
		return fmt.Sprintf("%s.%s.svc.local", DeploymentName, instance.Namespace), nil
	}

	if instance.Spec.ExternalAccess.Host != "" {
		return instance.Spec.ExternalAccess.Host, nil
	}

	return k8sutils.CalculateHostname(ctx, g.Client, DeploymentName, instance.Namespace)
}
```

internal/controller/fulcio/actions/generate_cert.go
bouskaJ: There is a log message just few lines above (L128)

JasonPowr: Apologies completely missed that

JasonPowr:
Done a quick test, on a cluster ran the cmd `oc delete fulcio --all` the fulcio cert secret does not seem to be getting deleted and the label is not being removed

bouskaJ:
The certificate is being revoked only if it needs to be changed.

JasonPowr:
> The certificate is being revoked only if it needs to be changed.

Does this not block key rotation for operator generated certificates then?, Or maybe we don't care about the operator generated ones since, its more for dev?

JasonPowr:
I guess it can still be done if we remove the label then, delete the fulcio instance?

bouskaJ:
oh, yes, I was focusing on the `SECURESIGN-1405` and totally forgot about key rotation. You need to invalidate key by removing `"rhtas.redhat.com/fulcio_v1.crt.pem"` label from the key before fulcio removal. Otherwise, the operator will try to compare the labeled secret and try to bind it with the new deployment.

You can use `oc label secret -l rhtas.redhat.com/fulcio_v1.crt.pem  rhtas.redhat.com/fulcio_v1.crt.pem-`

JasonPowr:
@bouskaJ can you update https://github.com/securesign/secure-sign-operator/blob/main/docs/fulcio-key-rotation.md with this :), thanks

JasonPowr:
@bouskaJ I think I found an issue with this, wdyt?, here is reproducers

1. Do a default deploy & install
2.  Create the keys and patch as per https://gitlab.cee.redhat.com/software-supply-chain-documentation/doc-rhtas/-/merge_requests/28/diffs#dd4e4d702e5901308da5d6e314eed5afd9b09c0a
3. check the status for the secret:
```
➜  secure-sign-operator git:(SECURESIGN-1405) ✗ oc get ctlog securesign-sample -o jsonpath='{.status.rootCertificates[*].name}'
fulcio-config%                                                                                                    
➜  secure-sign-operator git:(SECURESIGN-1405) ✗ oc get fulcio securesign-sample -o jsonpath='{.status.certificate.caRef.name}'
fulcio-config%                                                                                                    
➜  secure-sign-operator git:(SECURESIGN-1405) ✗ 
```

that secret is the one you just created however it does not have the rhtas.redhat.com/fulcio_v1.crt.pem: cert

bouskaJ:
Ready for review. @JasonPowr @osmman

PR #671, opened 2024-10-08
[SECURESIGN-1455] Fix rekor-server component

bouskaJ:


internal/controller/rekor/actions/server/resolve_pub_key.go
```diff
@@ -63,49 +63,28 @@ func (i resolvePubKeyAction) Handle(ctx context.Context, instance *rhtasv1alpha1
...
-		Kind:    "Secret",
-	})
-
-	if err = k8sutils.FindByLabelSelector(ctx, i.Client, scrl, instance.Namespace, RekorPubLabel); err != nil {
+	if partialSecret, err = k8sutils.FindSecret(ctx, i.Client, instance.Namespace, RekorPubLabel); client.IgnoreNotFound(err) != nil {
 		return i.Failed(fmt.Errorf("ResolvePubKey: find secrets failed: %w", err))
 	}
 
-	// Search if exists a secret with rhtas.redhat.com/rekor.pub label
-	for _, secret := range scrl.Items {
```
osmman, line 78: There has been a reason to fetch all secrets with rekor public label. It was added to handle unexpected cases when there is for some reason multiple secrets with that label (due some failure, user error...). Unfortunately I did not add test case for it. Here is reproducer:

```diff
diff --git a/internal/controller/rekor/actions/server/resolve_pub_key_test.go b/internal/controller/rekor/actions/server/resolve_pub_key_test.go
index a11da227..c9768f59 100644
--- a/internal/controller/rekor/actions/server/resolve_pub_key_test.go
+++ b/internal/controller/rekor/actions/server/resolve_pub_key_test.go
@@ -140,6 +140,32 @@ func TestResolvePubKey_Handle(t *testing.T) {
 				publicKey: testPublicKey,
 			},
 		},
+		{
+			name: "remove label from old secret and use existing secret",
+			env: env{
+				objects: []client.Object{
+					kubernetes.CreateSecret("old-secret-1", "default", map[string][]byte{
+						"public": testPublicKey2,
+					}, map[string]string{
+						RekorPubLabel: "public",
+					}),
+					kubernetes.CreateSecret("secret", "default", map[string][]byte{
+						"public": testPublicKey,
+					}, map[string]string{
+						RekorPubLabel: "public",
+					}),
+					kubernetes.CreateSecret("old-secret-2", "default", map[string][]byte{
+						"public": testPublicKey2,
+					}, map[string]string{
+						RekorPubLabel: "public",
+					}),
+				},
+			},
+			want: want{
+				result:    testAction.StatusUpdate(),
+				publicKey: testPublicKey,
+			},
+		},
 		{
 			name: "unable to resolve public key",
 			want: want{

```

internal/controller/rekor/actions/server/sharding_config.go
```diff
@@ -67,6 +70,34 @@ func (i shardingConfig) Handle(ctx context.Context, instance *rhtasv1alpha1.Reko
 		}
 	}
 
+	// try to discover existing config
+	partialSecrets, err := kubernetes.ListConfigMaps(ctx, i.Client, instance.Namespace, labels2.SelectorFromSet(labels).String())
```
osmman, line 74:
```suggestion
	partialConfigs, err := kubernetes.ListConfigMaps(ctx, i.Client, instance.Namespace, labels2.SelectorFromSet(labels).String())
```

internal/controller/rekor/actions/server/sharding_config.go
```diff
@@ -67,6 +70,34 @@ func (i shardingConfig) Handle(ctx context.Context, instance *rhtasv1alpha1.Reko
...
+	if err != nil && !apierrors.IsNotFound(err) {
```
osmman, line 75: I think that list function will not return NotFound error. If there is not records than it will return empty list

internal/controller/rekor/actions/server/sharding_config.go
```diff
@@ -67,6 +70,34 @@ func (i shardingConfig) Handle(ctx context.Context, instance *rhtasv1alpha1.Reko
...
+			i.Recorder.Eventf(instance, v1.EventTypeNormal, "ShardingConfigDiscovered", "Existing ConfigMap with sharding configuration discovered: %s", cm.Name)
+			instance.Status.ServerConfigRef = &rhtasv1alpha1.LocalObjectReference{Name: cm.Name}
+
+			meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+				Type:    actions.ServerCondition,
+				Status:  metav1.ConditionFalse,
+				Reason:  constants.Creating,
+				Message: "Sharding config discovered",
+			})
+			return i.StatusUpdate(ctx, instance)
```
osmman, line 93: we should continue to process all items in the list. There can be a case when first item is valid and second should be deleted.

internal/controller/rekor/actions/server/generate_signer_test.go
osmman, line 1: You removed this test case after changes https://github.com/securesign/secure-sign-operator/commit/b9433a8104e879e9cf0fa2c0c802256213c9f641, please add it back.

PR #673, opened 2024-10-08
refactor: constants for app.kubernetes.io labels

osmman:


internal/controller/common/utils/kubernetes/common.go
```diff
@@ -22,25 +22,14 @@ import (
...
-	out := map[string]string{}
-	for key, value := range labels {
-		if key == "app.kubernetes.io/part-of" || key == "app.kubernetes.io/instance" {
-			out[key] = value
-		}
-	}
-	return out
-}
-
 func FilterOutCommonLabels(labels map[string]string) map[string]string {
```
bouskaJ, line 27: I don't think that _kubernetes/common.go_ is correct place to have this function in. We already have similar functions in https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/constants/labels.go#L42 but again - I don't like `constants` package for such functions. Maybe we can extract it to some util package.

internal/controller/common/utils/kubernetes/secret.go
```diff
@@ -68,6 +68,25 @@ func GetSecretData(client client.Client, namespace string, selector *rhtasv1alph
 	return nil, nil
 }
 
+func FindSecrets(ctx context.Context, c client.Client, namespace string, labelSelector string) (*metav1.PartialObjectMetadataList, error) {
```
bouskaJ, line 71: Please move this function to another PR. It does not have anything common with  `refactor: constants for app.kubernetes.io labels`

internal/controller/common/utils/kubernetes/common.go
osmman: I'm considering moving all the Labels functionality into a dedicated labels package, similar to what we've done for annotations. However, this would involve significant changes during the refactor, so I'd prefer to do it after we've addressed most of our actions to avoid dealing with too many conflicts.

internal/controller/common/utils/kubernetes/secret.go
osmman: I will remove that

PR #675, opened 2024-10-08
SECURESIGN-1460 | Ensure that TSA component does not create duplicated resources

JasonPowr:
@bouskaJ @osmman Just looking for some feed back on this approach to ensuring only one resources is created at a time, if its acceptable Ill apply it to other TSA resources, needs some more testing


internal/controller/tsa/actions/common.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+
+	"github.com/securesign/operator/api/v1alpha1"
+	"github.com/securesign/operator/internal/controller/constants"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (g generateSigner) validateExistingObject(obj metav1.Object, instance *v1alpha1.TimestampAuthority) bool {
+	expectedAnnotations := g.computeObjectAnnotations(instance)
+	for key, expectedValue := range expectedAnnotations {
```
bouskaJ, line 15: You use DeepDerivate function. There could be more annotations on existing object then just expected.

internal/controller/tsa/actions/common.go
```diff
@@ -0,0 +1,41 @@
...
+		if value, exists := obj.GetAnnotations()[key]; !exists || value != expectedValue {
+			return false
+		}
+	}
+	return true
+}
+
+func (g generateSigner) computeObjectAnnotations(instance *v1alpha1.TimestampAuthority) map[string]string {
+	annotations := make(map[string]string)
+	specHash, err := computeSpecHash(instance.Spec.Signer)
```
bouskaJ, line 25: This is working but I would prefer to have something readable in annotations. We can easily determine context of the secret during debugging. Can you use something similar to https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/fulcio/actions/generate_cert.go#L293-L309?

internal/controller/tsa/actions/common.go
JasonPowr: I get what you are saying here, but in this implementation, the only one I cared about was  `[constants.LabelNamespace+"/specHash"]`

internal/controller/tsa/actions/common.go
JasonPowr: Sure no worries :)

internal/controller/common/action/base_action.go
```diff
@@ -253,6 +254,18 @@ func EnsureAnnotations(managedAnnotations ...string) EnsureOption {
 	}
 }
 
+func EnsureNTPConfig() EnsureOption {
```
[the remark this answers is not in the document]
JasonPowr: @bouskaJ In this case no, when I remove a item from the config e.g, when I remove a server https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/timestampauthority_types.go#L160, it doesn't trigger an updated, related to https://issues.redhat.com/browse/SECURESIGN-1393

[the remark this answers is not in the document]
osmman: I think that for TSA ntp monitoring config will be better to compare ObserveGeneration and reseource Generation. https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/ctlog/actions/server_config.go#L47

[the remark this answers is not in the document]
JasonPowr: @osmman The ntp monitoring action wont run unless it is picked up here https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/securesign/actions/ensure_tsa.go, can we still use this? I need to look into the ObserveGeneration and resourceGeneration more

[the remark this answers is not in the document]
bouskaJ: Good point Jason. I think that the Generation can be used in CanHandle function but it have no impact on the Ensure function. The EnsureSpec will not work after timestamp config removal

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -51,30 +59,103 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
+				ObservedGeneration: instance.Generation,
+			})
+			meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+				Type:               constants.Ready,
+				Status:             metav1.ConditionFalse,
+				Reason:             constants.Failure,
+				Message:            err.Error(),
+				ObservedGeneration: instance.Generation,
+			})
+			return g.FailedWithStatusUpdate(ctx, err, instance)
```
bouskaJ, line 105: I don't think we need to fail the whole TSA deployment. I think it is enough to throw `g.Fail()` in this case and give it another try in few seconds.

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -51,30 +59,103 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
```
bouskaJ, line 134: I don't think we need to fail the whole TSA deployment. I think it is enough to throw `g.Fail()` in this case and give it another try in few seconds.

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -51,30 +59,103 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
+		meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+			Type:               TSASignerCondition,
+			Status:             metav1.ConditionFalse,
+			Reason:             constants.Creating,
+			ObservedGeneration: instance.Generation,
+		})
 		return g.StatusUpdate(ctx, instance)
 	}
 
+	instance.Status.Signer = instance.Spec.Signer.DeepCopy()
```
bouskaJ, line 76: add 
```
if instance.Status.Signer != nil {
// check signer validity
}
```
block

bouskaJ, line 76: We can end-up right here if `instance.Spec.Signer != nil`

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -86,111 +167,74 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
+			ObservedGeneration: instance.Generation,
+		})
+		meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+			Type:               constants.Ready,
+			Status:             metav1.ConditionFalse,
+			Reason:             constants.Failure,
+			Message:            err.Error(),
+			ObservedGeneration: instance.Generation,
+		})
+		return g.FailedWithStatusUpdate(ctx, err, instance)
```
bouskaJ, line 204: I don't think we need to set Status to Failure in this case. `g.Fail()` should be enough IMHO.

internal/controller/tsa/actions/generate_signer_test.go
```diff
@@ -51,13 +53,6 @@ func Test_SignerCanHandle(t *testing.T) {
 			},
 			expected: true,
 		},
-		{
```
bouskaJ, line 54: Why did you remove this test? I think it make sense to keep it here.

internal/controller/tsa/actions/generate_signer_test.go
```diff
@@ -123,6 +118,10 @@ func Test_SignerHandle(t *testing.T) {
 					t.Errorf("Secret name mismatch: expected %v, got %v", instance.Status.Signer.File.PrivateKeyRef.Name, secret.Name)
 					return false
 				}
+				if !g.Expect(secret.Annotations).To(Equal(generateSecretAnnotations(instance))) {
```
bouskaJ, line 121: Do you need to use `if` condition? Isn't the `gomega.Expect` enough to verify?

internal/controller/tsa/actions/ntpMonitoring.go
```diff
@@ -32,19 +39,23 @@ func (i ntpMonitoringAction) Name() string {
...
+		return !equality.Semantic.DeepEqual(instance.Spec.NTPMonitoring, instance.Status.NTPMonitoring)
+	case c.Reason == constants.Ready:
+		return instance.Generation != c.ObservedGeneration
+	default:
+		return false
+	}
 }
 
 func (i ntpMonitoringAction) Handle(ctx context.Context, instance *rhtasv1alpha1.TimestampAuthority) *action.Result {
-	if meta.FindStatusCondition(instance.Status.Conditions, constants.Ready).Reason != constants.Creating {
```
bouskaJ, line 40: What will happen if the user do the change on `Ready` resource? We should move it back to the `Creating` status to ensure that all next actions can handle.

internal/controller/tsa/actions/generate_signer.go
JasonPowr: @bouskaJ Why do we need this check here? If instance.Status.Signer is nil it will be instantiated by DeepCopy(), then we can proceed with the validity check? Please correct me if I am wrong or misunderstanding :)

internal/controller/tsa/actions/ntpMonitoring.go
JasonPowr: @bouskaJ https://github.com/securesign/secure-sign-operator/blob/455f8bcdbff5c026f3fedb40b3d2f17e21b8ebab/internal/controller/tsa/actions/ntpMonitoring.go#L113-L118

we still do change to a creating phase here, if one is discovered, and here when a new one is created https://github.com/securesign/secure-sign-operator/blob/455f8bcdbff5c026f3fedb40b3d2f17e21b8ebab/internal/controller/tsa/actions/ntpMonitoring.go#L147-L152

internal/controller/tsa/actions/generate_signer.go
JasonPowr: @bouskaJ I have added this can you please take a look

internal/controller/tsa/actions/ntpMonitoring.go
bouskaJ: It make sense. Thank you for explaining.

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -332,3 +404,83 @@ func (g generateSigner) handleCertificateChain(ctx context.Context, instance *v1
...
+			}
+		}
+	}
+}
+
+func (g generateSigner) secretAnnotations(instance *v1alpha1.TimestampAuthority) (map[string]string, error) {
+	annotations := make(map[string]string)
+	chain := instance.Status.Signer.CertificateChain
+
+	marshalAndSetAnnotation := func(key string, value interface{}) error {
```
bouskaJ, line 449: Do we need all this fields to be json values? In all other components, we use direct references like `_tas_namespace_/secretKeyRef: ref`

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -51,30 +59,131 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
 	}
 
+	if instance.Status.Signer != nil {
+		secret, err := k8sutils.GetSecret(g.Client, instance.Namespace, instance.Status.Signer.CertificateChain.CertificateChainRef.Name)
+		if err != nil {
+			return g.Failed(fmt.Errorf("can't load CA secret %w", err))
+		}
+		anno, err := g.secretAnnotations(instance)
+		if err != nil {
+			meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
```
bouskaJ, line 83: I would not be so strict and fail the deployment in this case. Maybe it worth to throw `g.Fail()` and retry next time.

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -51,30 +59,131 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
+				Type:               constants.Ready,
+				Status:             metav1.ConditionFalse,
+				Reason:             constants.Failure,
+				Message:            err.Error(),
+				ObservedGeneration: instance.Generation,
+			})
+			return g.Failed(err)
+		}
+		if equality.Semantic.DeepDerivative(anno, secret.GetAnnotations()) {
+			return g.Continue()
```
bouskaJ, line 100: We can get there by fulfilling any of following
```
case !meta.IsStatusConditionTrue(instance.GetConditions(), TSASignerCondition):
		return true
	default:
		return !equality.Semantic.DeepDerivative(instance.Spec.Signer, *instance.Status.Signer)
```

I would change order of actions there
- first check the Spec != nil && Spec == Status - if they do not match take action and return
- then use [this](https://github.com/securesign/secure-sign-operator/pull/675/files#diff-6de0224d464917eb428b6ef8b58cc714a3f4a47e6b885535d33a33b1d2e25883R76-R103) block and do not forget to set `TimestampCondition` to true if secret is resolved
- autodiscovery
- generate new

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -51,30 +59,131 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
+		}
+	}
+
+	instance.Status.Signer = instance.Spec.Signer.DeepCopy()
+	//Check if a secret for the TSA cert already exists and validate
+	partialSecret, err := k8sutils.FindSecret(ctx, g.Client, instance.Namespace, TSACertCALabel)
```
bouskaJ, line 106: Find secret will fail if multiple secrets are found. It is more safe to list them all and invalidate (unlabel) non-valid ones
See https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/ctlog/actions/handle_keys.go#L150-L175

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -51,30 +59,131 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		g.Logger.Error(err, "problem with finding secret", "namespace", instance.Namespace)
+	}
+
+	if partialSecret != nil {
+		secret, err := k8sutils.GetSecret(g.Client, partialSecret.Namespace, partialSecret.Name)
```
bouskaJ, line 112: you can work with partialSecret object and do not load the whole secret into memory.

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -51,30 +59,131 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
+				Status:             metav1.ConditionTrue,
+				Reason:             "Resolved",
+				ObservedGeneration: instance.Generation,
+			})
+			return g.StatusUpdate(ctx, instance)
+		}
+
+		// invalidate certificate
+		if err := constants.RemoveLabel(ctx, partialSecret, g.Client, TSACertCALabel); err != nil {
+			meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
```
bouskaJ, line 148: Remove status condition settings. Just throw g.Failed()

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -86,111 +195,74 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
+			ObservedGeneration: instance.Generation,
 		})
 		g.StatusUpdate(ctx, instance)
 		// swallow error and retry
 		return g.Requeue()
 	}
 
+	anno, err := g.secretAnnotations(instance)
+	if err != nil {
+		meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
```
bouskaJ, line 218: Do not set conditions. Just g.Failed

internal/controller/tsa/actions/ntpMonitoring.go
```diff
@@ -54,52 +65,92 @@ func (i ntpMonitoringAction) Handle(ctx context.Context, instance *rhtasv1alpha1
 	ntpConfig, err := i.handleNTPMonitoring(instance)
```
bouskaJ, line 65: I would add 
	```
	if instance.Spec.NTPMonitoring.Config.NtpConfigRef != nil {
// update status fields
		return g.StatusUpdate
	}
	```
	before any computation. You can cut-off one execution branch right away (and it is very readable)

internal/controller/tsa/actions/ntpMonitoring.go
```diff
@@ -54,52 +65,92 @@ func (i ntpMonitoringAction) Handle(ctx context.Context, instance *rhtasv1alpha1
...
+			Type:               constants.Ready,
+			Status:             metav1.ConditionFalse,
+			Reason:             constants.Failure,
+			Message:            err.Error(),
+			ObservedGeneration: instance.Generation,
 		})
 		return i.FailedWithStatusUpdate(ctx, err, instance)
 	}
 
 	if ntpConfig != nil {
```
bouskaJ, line 77: You do not need to apply this if after my previous comment application

internal/controller/tsa/actions/ntpMonitoring.go
```diff
@@ -54,52 +65,92 @@ func (i ntpMonitoringAction) Handle(ctx context.Context, instance *rhtasv1alpha1
...
 		labels := constants.LabelsFor(ComponentName, DeploymentName, instance.Name)
-		configMap := kubernetes.CreateImmutableConfigmap(NtpCMName, instance.Namespace, labels, map[string]string{"ntp-config.yaml": string(ntpConfig)})
-		if err = controllerutil.SetControllerReference(instance, configMap, i.Client.Scheme()); err != nil {
-			return i.Failed(fmt.Errorf("could not set controller reference for ConfigMap: %w", err))
+		labels[constants.LabelResource] = ntpConfigLabel
+
+		if instance.Status.NTPMonitoring == nil {
+			instance.Status.NTPMonitoring = instance.Spec.NTPMonitoring.DeepCopy()
```
bouskaJ, line 82: I do not like this approach too much because in case of any error (and setting the status into failure) We will lost all data from current status.

I do prefer to create new variable for new status and set it to the instance after successful execution (https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/ctlog/actions/handle_keys.go#L71)

osmman:
I think that rootCA, intermediateCA, leafCA should not be created if signer uses key from is KMS or Tink service. These should be using certificate chain stored in `certificateChainRef`.

JasonPowr:
@osmman Does it create rootCA, intermediateCA, leafCA keys when other signer types are specified ? Originally it wasn't supposed to, here: https://github.com/securesign/secure-sign-operator/blob/6052c539ab07df9cf8918a9fa0a08f7826568039/internal/controller/tsa/actions/generate_signer.go#L129, We only handle keys e.g  rootCA, intermediateCA, leafCA if it is a file signer type. Maybe I am misunderstanding ?

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -51,30 +57,92 @@ func (g generateSigner) Handle(ctx context.Context, instance *v1alpha1.Timestamp
...
+	for _, partialSecret := range partialSecrets.Items {
+		if equality.Semantic.DeepDerivative(anno, partialSecret.GetAnnotations()) {
+			g.alignStatusFields(partialSecret.Name, instance)
+			meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+				Type:               TSASignerCondition,
+				Status:             metav1.ConditionTrue,
+				Reason:             "Resolved",
+				ObservedGeneration: instance.Generation,
+			})
+			return g.StatusUpdate(ctx, instance)
```
osmman, line 116: this will interupt evaluation of other partialSecrets. For example if there are 2 secrets, 1. is valid and 2. should be invalidated than in this case we end up with two secrets with TSACertCALabel because 2. secret will not we processed

osmman:
> @osmman Does it create rootCA, intermediateCA, leafCA keys when other signer types are specified ? Originally it wasn't supposed to, here:
> 
> https://github.com/securesign/secure-sign-operator/blob/6052c539ab07df9cf8918a9fa0a08f7826568039/internal/controller/tsa/actions/generate_signer.go#L129
> 
> , We only handle keys e.g rootCA, intermediateCA, leafCA if it is a file signer type. Maybe I am misunderstanding ?
> and we will only create a cert chain if one is not specified
> 
> https://github.com/securesign/secure-sign-operator/blob/6052c539ab07df9cf8918a9fa0a08f7826568039/internal/controller/tsa/actions/generate_signer.go#L329-L343
> 
> , if using a tink or kms, a certchain should be provided

I think that the action will fail if KMS or Tink resource and cert chain ref is empty/configured to autocreate certs. It will try to resolve certificate chain https://github.com/securesign/secure-sign-operator/blob/6052c539ab07df9cf8918a9fa0a08f7826568039/internal/controller/tsa/actions/generate_signer.go#L153 
which need RootPrivateKey. That value do not exists because it is resolved only for File based signer.

PR #679, opened 2024-10-10
fix: Resolving rekor pub key locally against openshift

JasonPowr:
When running the operator locally against an openshift cluster the Rekor pub key fails to resolve, this fixes that

### DIFF
--- a/internal/controller/rekor/actions/server/resolve_pub_key.go
+++ b/internal/controller/rekor/actions/server/resolve_pub_key.go
@@ -17,6 +17,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 )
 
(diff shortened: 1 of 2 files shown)


internal/controller/rekor/actions/server/resolve_pub_key.go
```diff
@@ -127,9 +128,19 @@ func (i resolvePubKeyAction) resolvePubKey(instance rhtasv1alpha1.Rekor) ([]byte
 	var (
 		data []byte
 		err  error
+		url  = fmt.Sprintf("http://%s.%s.svc", actions.ServerDeploymentName, instance.Namespace)
 	)
 
-	if data, err = i.requestPublicKey(fmt.Sprintf("http://%s.%s.svc", actions.ServerDeploymentName, instance.Namespace)); err == nil {
+	inContainer, err := k8sutils.ContainerMode()
+	if err == nil {
+		if !inContainer && k8sutils.IsOpenShift() {
```
osmman, line 136: Please remove `k8sutils.IsOpenShift()` from this statement. There shouldn't be any distinction between generic K8s and OpenShift in this case. In both instances, the operator should be allowed to use cluster internal DNS names (`%s.%s.svc`) when it's running inside the cluster.

osmman:
/override "Red Hat Konflux / operator-enterprise-contract / rhtas-operator"

PR #693, opened 2024-10-22
fix(ctlog): cleanup only linked server configs

osmman:
### Fix handling of cleanup of ServerConfig secrets and invalidation mechanism

**Handling of Deleted Generated Secrets:**
In cases where generated secrets were being deleted due to duplicate events during the operator's reconciliation process, I have adjusted order of execution. Now, any secrets that have the expected resource labels will be removed after successful creation of new one.

**ServerConfig Invalidation Mechanism:**
I have modified the mechanism that invalidates the current server configuration in response to external actions, such as the creation of a Trillian tree, issuance of Fulcio certificates, or the generation of signer keys. When any of these external actions result in a change of data, the ServerConfigAvaliable status is now correctly set to `false`. This triggers the necessary reconciliation to maintain the correct state of the server configuration.

### DIFF
--- a/internal/controller/ctlog/actions/constants.go
+++ b/internal/controller/ctlog/actions/constants.go
@@ -8,7 +8,13 @@ const (
 	RBACName           = "ctlog"
 	MonitoringRoleName = "prometheus-k8s-ctlog"
 
-	CertCondition    = "FulcioCertAvailable"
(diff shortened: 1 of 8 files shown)


bouskaJ:
/lgtm
Can you apply similar approach to other components?

PR #728, opened 2024-11-07
chore: bump version of operator to 1.1.1

osmman:


bouskaJ:
LGTM but I would keep the final merge to some Konflux expert

JasonPowr (approved):
This Looks good to me, we will need to apply the ProjectDevelopmentStream to the konflux cluster, so the application in konlfux is created :)

PR #740, opened 2024-11-12
PoC rework ensure

bouskaJ:


internal/controller/common/utils/kubernetes/common.go
```diff
@@ -133,3 +137,23 @@ func FindByLabelSelector(ctx context.Context, c client.Client, list client.Objec
 
 	return c.List(ctx, list, client.InNamespace(namespace), listOptions)
 }
+
+func CreateOrUpdate[T client.Object](ctx context.Context, cli client.Client, obj T, fn func(object T) error) (result controllerutil.OperationResult, err error) {
```
osmman, line 141: lets allow to use multiple mutate function so we can simply reuse common operations. For example to have:

```
kubernets.CreateIrUpdate(ctx, cli, &Deployment{},
   // examples for migration tasks
   migration.RemoveTUFContainer_1_1()
   migration.RemoveDeprecatedAnnotations_1_1()
   // specific mutation function
   deployment(...)
   // generic mutation fucntions
   mutate.Labels(),
   mutate.Annotations(),
   mutate.Proxy(),
   mutate.TrustedCA(),
```

internal/controller/common/utils/ensure.go
```diff
@@ -0,0 +1,24 @@
+package utils
+
+import (
+	"golang.org/x/exp/maps"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func EnsureLabels(obj client.Object, labels map[string]string) {
```
osmman, line 8: I I correctly remember these function will need to be more complex, we need a functionality to allow remove some of annotations and labels which we manage. 

https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/common/action/base_action.go#L232

internal/controller/tuf/actions/deployment.go
```diff
@@ -32,34 +36,105 @@ func (i deployAction) CanHandle(_ context.Context, tuf *rhtasv1alpha1.Tuf) bool
...
+		}
+
+		template := &spec.Template
+		template.Labels = labels
+		template.Spec.ServiceAccountName = sa
+
+		if template.Spec.Volumes == nil {
+			template.Spec.Volumes = make([]core.Volume, 1)
+		}
+		volume := &template.Spec.Volumes[0]
```
osmman, line 87: we should not use specific position in lists like `0` but we should try to find correct object based on `name`. It will be good to implement some helper method to make it easier.

internal/controller/tuf/actions/deployment.go
```diff
@@ -32,34 +36,105 @@ func (i deployAction) CanHandle(_ context.Context, tuf *rhtasv1alpha1.Tuf) bool
...
+		volume.VolumeSource = core.VolumeSource{
+			PersistentVolumeClaim: &core.PersistentVolumeClaimVolumeSource{
+				ClaimName: instance.Status.PvcName,
+			},
+		}
+
+		if template.Spec.Containers == nil {
+			template.Spec.Containers = make([]core.Container, 1)
+		}
+		container := &template.Spec.Containers[0]
```
osmman, line 98: used hardcoded position

osmman (changes requested):
Hi @bouskaJ, the core functionality looks solid. However, I would like us to move this to a more reasonable package structure to avoid the common/utils clutter.

internal/controller/common/utils/ensure.go
bouskaJ: you are right. I forget to implement `remove` functionality :+1:

osmman (changes requested):
It looks good, please migrate these test scenarios for new implementation

https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/common/action/base_action_test.go

osmman (approved):

PR #745, opened 2024-11-12
ci: modify Konflux pipelines

osmman:
Migrated Konflux pipeline based on OLM operator example 

https://github.com/konflux-ci/olm-operator-konflux-sample


tommyd450:
@osmman is there a reason you are chaining the pipelines? Rather than having them be in the one tekton file? Im also going to re run Konflux builds as the tasks on Konflux are not reporting much .

osmman:
> @osmman is there a reason you are chaining the pipelines? Rather than having them be in the one tekton file? Im also going to re run Konflux builds as the tasks on Konflux are not reporting much .

I’m chaining the pipelines to avoid duplicating code for push and pull_request pipelines. This approach makes it easier to add new steps, like unit tests, in future updates. I followed the Konflux example for this setup.

osmman:
> @osmman is there a reason you are chaining the pipelines? Rather than having them be in the one tekton file? Im also going to re run Konflux builds as the tasks on Konflux are not reporting much .

tommyd450 (approved):

PR #786, opened 2024-12-11
Add integration test

bouskaJ:


bouskaJ:
I will keep the IntegrationTest present only on the `main` stream that is managed manually. The reason is that I am afraid that the oldest stream (v1.0) test suite is not compatible with the pipeline implementation

PR #793, opened 2024-12-18
[SECURESIGN-1393] Migrate to the new ensure logic

bouskaJ:


osmman (changes requested):
We need to create unit tests for `Ensure*` functions to define how we expecting from them to modify existing objects. I think about these scenarios should be mandatory for all these functions:

- empty object -> modify to expected values
- existing object without all correct values -> no changes
- existing object contains extra values (port, rule, ...) -> **keep** or **delete** that values, based on our implementation requirements which will be important for handling upgrades
- existing object with missing values -> values will be added

bouskaJ:
> We need to create unit tests for `Ensure*` functions to define how we expecting from them to modify existing objects. I think about these scenarios should be mandatory for all these functions:
> 
> * empty object -> modify to expected values
> * existing object without all correct values -> no changes
> * existing object contains extra values (port, rule, ...) -> **keep** or **delete** that values, based on our implementation requirements which will be important for handling upgrades
> * existing object with missing values -> values will be added

@osmman tests were added but the expected behavior with existing object is a big question. I keep values that are not subject of the operator configuration unchanged with respect to default values set by another controllers but that means that we will not revert misconfiguration on those fields. What is your opinion on this?

osmman:
> > We need to create unit tests for `Ensure*` functions to define how we expecting from them to modify existing objects. I think about these scenarios should be mandatory for all these functions:
> > 
> > * empty object -> modify to expected values
> > * existing object without all correct values -> no changes
> > * existing object contains extra values (port, rule, ...) -> **keep** or **delete** that values, based on our implementation requirements which will be important for handling upgrades
> > * existing object with missing values -> values will be added
> 
> @osmman tests were added but the expected behavior with existing object is a big question. I keep values that are not subject of the operator configuration unchanged with respect to default values set by another controllers but that means that we will not revert misconfiguration on those fields. What is your opinion on this?

Exactly, that’s the problem—the choice of behavior depends heavily on the object, and I don’t expect we’ll have a uniform approach for all k8s objects:
- For example, we might override all RoleBindings.
- For labels and annotations, we only manage our own annotations and ignore others.
- For deployment, there could be cases where another operator adds a sidecar or a mounted volume.

Currently, I’m unsure how all these scenarios will be handled by our operator. I think to a straightforward approach where, for most objects, we simply override the values if it makes sense to do so. However, since the behavior can differ depending on the object type and context, it’s critical to define these cases explicitly in unit tests as documentation for our feature cahnges.

osmman (approved):

PR #799, opened 2025-01-10
[SECURESIGN-1393] Migrate svc

bouskaJ:


internal/controller/trillian/actions/db/svc.go
```diff
@@ -39,42 +43,32 @@ func (i createServiceAction) CanHandle(_ context.Context, instance *rhtasv1alpha
...
+		kubernetes.EnsureServiceSpec(labels, v1.ServicePort{
+			Name:       host,
+			Protocol:   v1.ProtocolTCP,
+			Port:       port,
+			TargetPort: intstr.FromInt32(port),
+		}),
+		ensure.ControllerReference[*v1.Service](instance, i.Client),
+		ensure.Labels[*v1.Service](maps.Keys(labels), labels),
+		//TLS: Annotate service
+		ensure.Optional(kubernetes.IsOpenShift() && instance.Spec.Db.TLS.CertRef == nil, ensure.Annotations[*v1.Service](maps.Keys(tlsAnnotations), tlsAnnotations)),
```
osmman, line 66: There is a scenario which will end up with undesired `tlsAnnotation` when `CertRef` will change from `nil` to some value on Openshift cluster.

```
trillian:
  db:
     create: true
```

modify to

```
trillian:
  db:
     create: true
     tls:
      certificateRef:
        key: secret
        name: cert
      privateKeyRef:
        key: secret
        name: key
```

osmman (changes requested):

internal/controller/trillian/actions/db/svc.go
bouskaJ: Good point!. fixed

osmman (approved):

PR #803, opened 2025-01-13
[SECURESIGN-1393] migrate ConfigMap

bouskaJ:
- migrate ConfigMap to new ensure
- eliminate CreateConfigMap to avoid usage in source-code


bouskaJ:
The `Red Hat Konflux / rhtas-operator-e2e-test / rhtas-operator ` fails on cluster provisioning. We will need to wait till konflux is in better shape but It should be ready for review.

osmman (approved):

PR #812, opened 2025-01-15
feat: use RELATED_IMAGES_ environment variables to configure operand's images

osmman:
This pull request introduces a new feature that allows the Operator manager to utilize `RELATED_IMAGES_*` environment variables for image configurations. This change allows operator-sdk to generate bundle image with relatedImages meta data which are used for disconnected environment. Additionally, it includes several refactoring changes to improve deployment for different environments Kubernetes/Openshift.

#### Key Features
- **Environment Variables**: Use `RELATED_IMAGES_` environment variables for image configurations.
- **Makefile Updates**: Added `CONFIG_DEFAULT` for dynamic configuration based on the `OPENSHIFT` variable.
- **Codebase Enhancements**: Centralized image management by importing `internal/images`.


internal/images/images.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	Registry = registry{
+		data: data,
+	}
+
+}
+
+// parseConfigFile parses an embedded `.env` content and returns a map of key-value pairs.
+func parseConfigFile(envContent string) (map[Image]string, error) {
```
bouskaJ, line 55: Did you considered using Viper https://github.com/spf13/viper ?

osmman: Did not thing for this case. We can move our configuration of operator to be managed by it.

internal/images/images.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("error reading env content: %w", err)
+	}
+
+	return data, nil
+}
+
+type registry struct {
+	mutex sync.RWMutex
```
bouskaJ, line 85: why do we need synchronization? do you expect images to be changed on runtime?

ci/related-image-patch.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+#
+# Usage: ./related-image-patch.sh inputfile
```
bouskaJ, line 3: Where do you use this script?

bundle/manifests/rhtas-related-images_v1_configmap.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+data:
+  RELATED_IMAGE_BACKFILL_REDIS: registry.redhat.io/rhtas/rekor-backfill-redis-rhel9@sha256:6aa3ca40e0f9e32a0a211a930b21ff009b83e46609bfa5bb328979e4799d13c7
```
bouskaJ, line 3: I know that this is needed by Kustomize but I don't think it is necessary to keep it as part of the bundle. Maybe we can add it to .gitignore WDYT?

ci/related-image-patch.sh
osmman: is will be used in https://github.com/openshift/release/pull/61505

internal/images/images.go
osmman: I added synchronisation because the values are configured via callbacks (async).

osmman: Unfortunatly I forget the reason why I have to do that way. It has been a long time ago when I develop that.

osmman: I reworked that registry to remove async setter but I think that it is bettor to keep synchronisation because registry is singleton. The async setter is looking to be left over from prototyping and hacking flag package.

osmman:
/override Red Hat Konflux / rhtas-operator-e2e-test / rhtas-operator

osmman:
/override "Red Hat Konflux / rhtas-operator-e2e-test / rhtas-operator"

bouskaJ (approved):

PR #815, opened 2025-01-15
PoC 2: feat: use RELATED_IMAGES_ environment variables for manager

osmman:


osmman:
Rejecting proposal to use PoC from https://github.com/securesign/secure-sign-operator/pull/812

PR #838, opened 2025-02-06
Migrate segment jobs and securesign resources

bouskaJ:


internal/controller/common/utils/set_trusted_ca_test.go
osmman, line 1: Are these scenarios covered with some other test?

bouskaJ: https://github.com/securesign/secure-sign-operator/blob/remove_ensure/internal/controller/common/utils/kubernetes/ensure/deployment_test.go#L21

PR #871, opened 2025-02-20
feat: Default TLS encryption on Trillian services

osmman:


internal/controller/common/action/tree/action.go
```diff
@@ -0,0 +1,415 @@
...
+	if _, err = kubernetes.CreateOrUpdate(ctx, i.Client, &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      rbacName,
+			Namespace: instance.GetNamespace(),
+		},
+	},
+		ensure.ControllerReference[*corev1.ServiceAccount](instance, i.Client),
+		ensure.Labels[*corev1.ServiceAccount](maps.Keys(labels), labels),
+	); err != nil {
+		return i.Error(ctx, reconcile.TerminalError(fmt.Errorf("could not create SA: %w", err)), instance)
```
bouskaJ, line 93: Error without Terminal Error reconcile with exponential backoff. IMHO, we should throw the TerminalError only in cases in which we can't recover by any means (with wrong user-provided configuration). In this case, it is still possible that the ServiceAccount will be created in the new round. It is even possible that the user will align his cluster RBAC settings during the deployment, so enabling the exponential backoff makes sense for me. WDYT?

internal/controller/common/action/tree/action.go
```diff
@@ -0,0 +1,415 @@
...
+
+	labels := labels.For("createtree", i.component, instance.GetName())
+
+	configMapName := fmt.Sprintf(configMapResultMask, i.component, instance.GetName())
+	configMap, err := kubernetes.GetConfigMap(ctx, i.Client, instance.GetNamespace(), configMapName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return i.Requeue()
+		}
+		return i.Error(ctx, reconcile.TerminalError(fmt.Errorf("could not get configmap: %w", err)), instance)
```
bouskaJ, line 189: See my comment about errors above. I think it makes sense to retry in this case.

internal/controller/common/action/tree/action.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+
+	if val, ok := configMap.GetAnnotations()[jobReferenceAnnotation]; ok && val != "" {
```
bouskaJ, line 192: Is it possible to switch to ownership references?

internal/controller/common/action/tree/action.go
bouskaJ: yes, that would make sense. I kept original implementation during migration RBACs to the `error` function.

internal/controller/common/action/tree/action.go
osmman: I will look at it

internal/controller/common/action/tree/action.go
osmman: https://github.com/securesign/secure-sign-operator/pull/871#discussion_r1977261656

internal/controller/common/action/tree/action.go
osmman: I modified implementation to use ownership reference

internal/controller/ctlog/actions/resolve_tree.go
```diff
@@ -1,116 +1,24 @@
...
-		})
-		meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
-			Type:               constants.Ready,
-			Status:             metav1.ConditionFalse,
-			Reason:             constants.Failure,
-			Message:            err.Error(),
-			ObservedGeneration: instance.Generation,
+func NewResolveTreeAction() action.Action[*rhtasv1alpha1.CTlog] {
+	wrapper := tree.Wrapper[*rhtasv1alpha1.CTlog](
+		func(rekor *rhtasv1alpha1.CTlog) *int64 {
```
bouskaJ, line 11: Creating the struct directly with named arguments would be more readable. Take it or not...
```
w := &Wrapper[*rhtasv1alpha1.CTlog] {
	callTree:            func()....
	callStatusTree:      func()...
.....
}

Comment (osmman) [reply internal/controller/ctlog/actions/resolve_tree.go:11]: Unfortunately, that approach would require making all fields in the `wrapper` structure public. I’m concerned that this could lead to potential misuse, such as directly manipulating fields or invoking the callback function. Which isn’t ideal.

Comment (osmman) [reply internal/controller/ctlog/actions/resolve_tree.go:11]: I can create a new structure only for constructor `tree.wrapperOpts` like:

```
wrapper := tree.Wrapper[*rhtasv1alpha1.Rekor](tree.WrapperOpts[*rhtasv1alpha1.Rekor]{
		CallTree: func(rekor *rhtasv1alpha1.Rekor) *int64 {
			return rekor.Spec.TreeID
		},
		CallStatusTree: func(rekor *rhtasv1alpha1.Rekor) *int64 {
			return rekor.Status.TreeID
		},
		CallSetStatusTree: func(rekor *rhtasv1alpha1.Rekor, i *int64) {
			rekor.Status.TreeID = i
		},
		CallTrillianService: func(rekor *rhtasv1alpha1.Rekor) *rhtasv1alpha1.TrillianService {
			return &rekor.Spec.Trillian
		},
	})
```

PR #878, opened 2025-02-26
feat(retrievePodLogs): fetch logs from pods and add to zip 

kdacosta0:
- retrieves logs from every pod in the given namespace
- adds each log file to a zip folder under "pod-logs/"


osmman:
@kdacosta0 can you rebase your PR agains main branch. It should solve the problem of not passing Build-fbc check

PR #899, opened 2025-03-05
feat(fips): build manager for strict fips runtime

osmman:
Modify manager's Red Hat's dockerfile to build binary which runs on an Operating System in FIPS mode. 

Tool to verify resulted operator https://github.com/openshift/check-payload


osmman:
I removed installation of OpenSSL package because it is not needed. OpenSSL is available via libcrypto found in /usr/lib64

PR #928, opened 2025-03-17
fix: Instability of cleanup process of old configmaps and secrets

osmman:
CTlog, Rekor and Fulcio has broblem when newly created object has been deleted during clean up process of old objects.


bouskaJ:
just a minor comment. Otherwise LGTM

PR #932, opened 2025-03-19
fix: mount kubernetes api server certificate to /var/run/fulcio/ca.crt

osmman:
fixes issue [SECURESIGN-1989](https://issues.redhat.com//browse/SECURESIGN-1989)

### DIFF
--- a/internal/controller/fulcio/actions/deployment.go
+++ b/internal/controller/fulcio/actions/deployment.go
@@ -5,8 +5,14 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/securesign/operator/internal/controller/common/utils/kubernetes/ensure/deployment"
-	"github.com/securesign/operator/internal/images"
+	"golang.org/x/exp/maps"
+	v1 "k8s.io/api/apps/v1"
+	core "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


internal/controller/fulcio/actions/deployment.go
```diff
@@ -218,6 +222,27 @@ func (i deployAction) ensureDeployment(instance *rhtasv1alpha1.Fulcio, sa string
 			},
 		}
 
+		oidcInfo := kubernetes.FindVolumeByNameOrCreate(&template.Spec, "oidc-info")
+		if oidcInfo.Projected == nil {
```
bouskaJ, line 226: as you are mounting only one ConfigMap you do not need to create a Projected volume but it has no impact overall on functionality. 

**Key Considerations:**
_Overhead:_ Using a Projected Volume with just one ConfigMap may introduce unnecessary complexity. The Projected Volume is meant to combine multiple sources into one volume, so using it for a single source (like a single ConfigMap) might add unnecessary overhead without any clear benefit.

_Simplicity:_ If you're only using one ConfigMap, directly referencing it using a ConfigMap Volume Source is more straightforward and simpler. The Projected approach is better suited when you plan to combine multiple sources like ConfigMap, Secrets, and other sources.

PR #943, opened 2025-03-21
[release-1.2] backport changes from main branch

osmman:


cmd/main.go
```diff
@@ -128,19 +125,6 @@ func main() {
 
 	ctrl.SetLogger(klog.NewKlogr())
 
-	// Register custom panic handlers
```
bouskaJ, line 131: Why do you remove the panic handler?

docs/Restore.md
```diff
@@ -62,6 +63,17 @@ EOF
...
 
+If the restore is done on a different cluster, few more steps needs to be done. First, delete the secret for Trillian DB which will be recreated by operator,
+and restart the pod:
+
+```sh
+oc delete secret securesign-sample-trillian-db-tls
+oc delete pod trillian-db-xxx
+```
+
+After the restore process is finished and all the pods are running, run the [restoreOwnerReferences.sh](../hack/restoreOwnerReferences.sh) script to recreate
```
bouskaJ, line 74: It seems we do not have the script backported to `release-1.2`

cmd/main.go
osmman: The functionality of controller_runtime_reconcile_panics_total metric has been added into new version of controller-runtime library. So registration and inrementation of that metric is already handled by `Controller.RecoverPanic`.
https://github.com/securesign/secure-sign-operator/pull/943/commits/2e76207d6f3aac8e02490ed792c31df7254dd450

PR #971, opened 2025-03-27
Update channel lables 1.2

JasonPowr:
Small pr to update the `operators.operatorframework.io.bundle.channels.v1` label


osmman (changes requested):
Need to modify channels values in makefile
https://github.com/securesign/secure-sign-operator/blob/release-1.2/Makefile#L14

PR #1089, opened 2025-05-05
[SECURESIGN-2268] Add rekor monitor

fghanmi:


osmman (commented):
Currently, I don't see a way to configure rekor-monitor for alerting on verification failures. Adding this functionality, and specifically integrating with OpenShift Alertmanager, would significantly improve our ability to respond to security issues.

For the periodic execution of rekor-monitor, switching to a CronJob could offer several advantages. It would utilize a built-in Kubernetes feature and potentially reduce resource consumption compared to a continuously running Deployment. It would be possible to simply create Alert when CronJob failed in redirect it by AlertManager to slack/email/webhook,pager.

internal/controller/rekor/actions/monitor/deployment.go
```diff
@@ -0,0 +1,143 @@
...
+		template.Spec.ServiceAccountName = sa
+
+		container := kubernetes.FindContainerByNameOrCreate(&template.Spec, actions.MonitorDeploymentName)
+		container.Image = images.Registry.Get(images.RekorMonitor)
+
+		container.Command = []string{
+			"/bin/sh",
+			"-c",
+			fmt.Sprintf(`
+		while true; do
```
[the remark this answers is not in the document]
fghanmi: I've updated the execution method for rekor-monitor, it's now running within a wrapper, allowing us to adjust the monitoring interval as needed."

bouskaJ:
Please do not add the OpenTelemetry controller. Integrate with native OpenShift stack.
1) Your monitor should expose metrics endpoint
2) create ServiceMonitor to connect your enpoint with buitin prometheus.
3) Create AlertingRule (PrometheusRule) to propagate alerts (this is probably task for end-user not for our app so we should document the recommended settings)

fghanmi:
> Hi Firas, I have a few questions:
> 
> * 1- Is this component required, or does it make sense to make it optional?
> * 2- Do  we really need an ingress service for it? If so, we should offer option to disable it to keep traffic internal.
> * 3- there is no serviceMonitor created (https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/rekor/actions/server/monitoring.go)  It is crucial part for integration with cluster monitoring stack.
> 
> Otherwise, it looks good 👍

1- Done
2- `Do we really need an ingress service for it? If so, we should offer option to disable it to keep traffic internal`: we already have externalAccess struct that can be used to disable/enable ingress routes (including rekor monitor's) 
3- Done

Thanks @bouskaJ

osmman (changes requested):
I've reviewed the proposed API changes for Rekor configuration and have a few suggestions regarding the schema design.

### My main concerns are:
- Unnecessary "**Rekor**" Repetition: The API context already implies Rekor. Repeating "rekor" in sections like rekorMonitor or other fields is redundant and verbose. We should aim for conciseness.
- Duplicative **rekorMonitor** Section: We already have a general monitoring section. Introducing a new high-level **rekorMonitor** section for Rekor's monitoring duplicates concerns and could fragment future monitoring configurations.

### Proposed Alternative API Schema:

I suggest integrating Rekor's transparency log (tlog) monitoring directly into the existing monitoring section. This approach consolidates all monitoring-related configurations and avoids the "rekor" repetition.

Here's how I envision the updated schema:
```yaml
rekor:
  monitoring:
    tlog:
      enabled: true
      interval: 10m
```

### In this proposed structure:

- The monitoring section remains the single source for all monitoring configurations.
- A new tlog subsection is added under monitoring specifically for Rekor's transparency log settings.
- enabled and interval options are directly nested within tlog, clearly indicating their application to transparency log monitoring.

api/v1alpha1/common.go
```diff
@@ -18,18 +19,25 @@ type ExternalAccess struct {
...
-type RekorMonitorConfig struct {
-	// If true, the Operator will create the Rekor log monitor resources
+type MonitoringConfig struct {
+	// If true, the Operator will create monitoring resources
 	//+kubebuilder:validation:XValidation:rule=(self || !oldSelf),message=Feature cannot be disabled
 	//+kubebuilder:default:=true
 	Enabled bool `json:"enabled"`
+	// Configuration for Rekor transparency log monitoring
+	//+optional
+	Tlog TlogMonitoring `json:"tlog"`
```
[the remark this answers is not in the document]
fghanmi: Thank you @osmman for the review!

bouskaJ:
I am still not able to deploy rekor-monitor with your default image. Please update the image to the latest one from konflux. After the image sha change, I am able configure the platform to trigger an alert when validation fails :+1: . 

Please finish Tomas's review. LGTM for me.

internal/images/images.go
```diff
@@ -24,6 +24,7 @@ const (
 	RekorRedis    Image = "RELATED_IMAGE_REKOR_REDIS"
 	RekorServer   Image = "RELATED_IMAGE_REKOR_SERVER"
 	RekorSearchUi Image = "RELATED_IMAGE_REKOR_SEARCH_UI"
+	RekorMonitor  Image = "RELATED_IMAGE_REKOR_MONITOR"
```
osmman, line 27: You need to extend operator cli flags by newly added image. https://github.com/securesign/secure-sign-operator/blob/main/cmd/main.go#L105-L120

internal/controller/rekor/actions/monitor/deployment.go
```diff
@@ -0,0 +1,138 @@
...
+	v1 "k8s.io/api/apps/v1"
+	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	rhtasv1alpha1 "github.com/securesign/operator/api/v1alpha1"
+)
+
+const storageVolumeName = "monitor-storage"
```
osmman, line 28: I noticed the use of `EmptyDirVolume`, which isn't persistent, meaning data will be lost on pod restarts. Is data persistence a requirement here, or is this intended to be ephemeral storage for the lifetime of a single pod?

internal/controller/rekor/actions/monitor/deployment.go
```diff
@@ -0,0 +1,138 @@
...
+		template.Spec.ServiceAccountName = sa
+
+		container := kubernetes.FindContainerByNameOrCreate(&template.Spec, actions.MonitorDeploymentName)
+		container.Image = images.Registry.Get(images.RekorMonitor)
+
+		container.Command = []string{
+			"/bin/sh",
+			"-c",
+			fmt.Sprintf(`
+				until curl -sf %s > /dev/null 2>&1; do
```
osmman, line 113: Please move the Rekor waiting script into an `initContainer`.

osmman (changes requested):

internal/controller/rekor/actions/monitor/deployment.go
osmman: @fghanmi Could you please check the size of the stored data? If it's consistently under 1.5MB, a ConfigMap might be a more lightweight and simpler solution for maintenance. I noticed rekor-monitor has functionality for deleting old checkpoints, which could be relevant here.

osmman: https://github.com/sigstore/rekor-monitor/blob/main/pkg/rekor/verifier.go#L124-L131

osmman: You're right, **ConfigMaps** are mounted as read-only filesystems.

To address this, please modify your changes to use a **StatefulSet** and its **volumeClaimTemplates** instead of a Deployment. This will allow the cluster to automatically manage and correctly attach **PersistentVolumeClaims** (PVCs) for each specific pod.

Additionally, please relocate the new PVC configuration from **spec.monitorpvcName** to **spec.monitoring.tlog.pvc**.

fghanmi: @osmman why do we need to consider the cluster to automatically attach PVCs for each specific pod ? 
we should have one single pod for the rekor-monitor. When a pod dies, another would get provisioned and be attached to the same PVC.

osmman: You're right to bring up the single pod, but for a stateful service like **rekor-monitor**, a **StatefulSet** is crucial, even with one replica.

Here's why:
- Stateful Nature: rekor-monitor needs to maintain persistent data. A StatefulSet guarantees stable, unique storage for each pod, ensuring data integrity.
- Rolling Upgrades: This is the big one. A **Deployment** will likely fail during a rolling upgrade because the new pod won't be able to attach to the PVC while the old one is still holding onto it. **StatefulSets** are designed to handle this gracefully, ensuring the new pod correctly acquires the PVC after the old one detaches.
- Automated PVC Management: With **volumeClaimTemplates**, the **StatefulSet** automatically provisions and manages the PVC for you, simplifying operator no need for **pvc** manager action.

internal/controller/rekor/actions/monitor/statefulset.go
```diff
@@ -0,0 +1,174 @@
...
+        echo "TreeSize is 0, resetting checkpoint log file"
+        > /data/checkpoint_log.txt
+      elif [ -n "$tree_size" ]; then
+        break
+      fi
+    fi
+    sleep %.0f
+  done
+) &
+exec /rekor_monitor --file=/data/checkpoint_log.txt --once=false --interval=%s --url=%s
```
osmman, line 125: Remove `exec`, I do not see a reason to use it.

internal/controller/rekor/actions/monitor/statefulset.go
```diff
@@ -0,0 +1,174 @@
...
+		container := kubernetes.FindContainerByNameOrCreate(&template.Spec, actions.MonitorStatefulSetName)
+		container.Image = images.Registry.Get(images.RekorMonitor)
+
+		interval := instance.Spec.Monitoring.TLog.Interval.Duration
+		backgroundSleep := interval / 2
+		container.Command = []string{
+			"/bin/sh",
+			"-c",
+			fmt.Sprintf(`(
+  while true; do
```
osmman, line 112: Why is this hack needed? I thought that it will be fixed directly in rekor-monitor.

osmman (changes requested):

internal/controller/rekor/actions/monitor/statefulset.go
fghanmi: as per this comment: https://github.com/sigstore/rekor-monitor/issues/698#issuecomment-3090691396
as long as the rekor server has 0 entries, the rekor_monitor would fail/stop (so will future attempts). So this background loop check keeps running and initializing log file until tree size is > 0.

PR #1094, opened 2025-05-07
refactor: make config/default/images.env single source of truth

osmman:
Uses go generate feature to copy images.env into embed folder. Which content will be added into embeded file system.

## Summary by Sourcery

Refactor the images configuration to use config/default/images.env as the single source of truth, leveraging go generate to manage the embedded configuration

Bug Fixes:
- Ensured consistent images configuration across different build and deployment processes

Enhancements:
- Moved images configuration from internal/images to config/default directory
- Implemented go generate to copy images.env into embedded filesystem

Chores:
- Updated workflow files to reference new images.env location
- Modified Makefile and Dockerfile to support new configuration approach


JasonPowr (approved):
/lgtm

PR #1115, opened 2025-05-15
fix: ServiceMonitor use https scheme to pull metrics

osmman:
Use correct schema for pulling Trillian metrics when TLS is enabled. It required extension of Trillain API to have separate TLS configuration for Server and Signer service.

Fixes [SECURESIGN-2450](https://issues.redhat.com//browse/SECURESIGN-2450)

## Summary by Sourcery

Enable per-service TLS configuration for Trillian database, log server, and log signer, and switch ServiceMonitor endpoints to HTTPS when TLS is enabled by extending the CRDs, API, and controllers, and updating manifests.

New Features:
- Add separate TLS configuration fields for Trillian log server, log signer, and database in the CRD and API types
- Introduce ServiceMonitorHttpsEndpoint to support HTTPS metric scraping with custom TLS settings
- Add new reconciliation actions to resolve and propagate TLS status for each Trillian service

Bug Fixes:
- Fix ServiceMonitor to use HTTPS scheme for scraping when TLS is enabled

Enhancements:
- Refactor controllers and deep-copy logic to use per-service TLS fields and remove the legacy monolithic TLS spec
- Update ServiceMonitor creation to dynamically choose HTTP or HTTPS based on service TLS status
- Bump related container image digests and update operator CSV metadata

Tests:
- Extend ServiceMonitor unit tests to cover HTTPS endpoint configuration


api/v1alpha1/trillian_types.go
```diff
@@ -29,14 +29,26 @@ type TrillianSpec struct {
 	Db TrillianDB `json:"database,omitempty"`
 	// Enable Monitoring for Logsigner and Logserver
 	Monitoring MonitoringConfig `json:"monitoring,omitempty"`
-	// Configuration for enabling TLS (Transport Layer Security) encryption for manged log-server and log-signer services.
-	//+optional
-	TLS TLS `json:"tls,omitempty"`
```
bouskaJ, line 34: This is an API change. I believe we should keep the `tls` field present for backward compatibility.

osmman: We do not need backward compatibility for this field because it is not introduced in any released version. It is new feature for 1.3.0

bouskaJ: ok, now I see it was added 2 mons ago https://github.com/securesign/secure-sign-operator/commit/0ce937c84ce5e0b2ecfe2c0ab0483faef48a3644

PR #1116, opened 2025-05-16
TLS for CTLOG/Fulcio communication

bouskaJ:


api/v1alpha1/common.go
```diff
@@ -46,7 +46,6 @@ type CtlogService struct {
 	// Port of Ctlog Log Server End point
 	//+kubebuilder:validation:Minimum:=1
 	//+kubebuilder:validation:Maximum:=65535
-	//+kubebuilder:default:=80
```
osmman, line 49: I expect that we cannot define default value for this field. The reason is that depends on current configuration. If run on Openshift than default should be 443.

bouskaJ: yes, that is the reason why I removed that.

osmman: I sorry, it looks like I am colour blind 😄

test/e2e/namespaced.go
```diff
@@ -129,14 +131,22 @@ var _ = Describe("Install components to separate namespaces", Ordered, func() {
 			},
 		}
 
+		protocol := "http"
+		ocp, err := kubernetes2.IsRemoteClusterOpenshift(config.GetConfigOrDie())
```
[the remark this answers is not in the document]
bouskaJ: OPENSHIFT env is set on operator deployment. The test is not executed by the operator so it does not work by default.

We have options to set the env variable on the test runner or resolve it from the cluster. Having **test** function that resolve that from cluster should not make any troubles with starting cluster (the cluster is already running during the test phase). The benefit of the automatic resolution is that you do not need to configure any additional variables on your local environment.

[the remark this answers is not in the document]
osmman: that function is already used in e2e tests byodb_test.go and common_install_test.go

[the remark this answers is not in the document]
bouskaJ: and I believe that IsRemoteClusterOpenshift is better than the strange ENV var reading https://github.com/securesign/secure-sign-operator/blob/main/test/e2e/byodb_test.go#L25 and maintaining the ENV var on all pipelines and developer stations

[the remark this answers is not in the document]
osmman: Can you create a new commit to migrate other tests to use a new implementation. I do not want to have multiple methods which handle same thing.

Can you add comment to `IsRemoteClusterOpenshift` method that it can be used only for tests and not for runtime because of instability of results.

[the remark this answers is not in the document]
bouskaJ: the function is in `"github.com/securesign/operator/test/e2e` package so I guess it is clear enough.

internal/controller/fulcio/actions/deployment.go
```diff
@@ -88,6 +83,30 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Fulcio
...
+	if instance.Spec.Ctlog.Prefix == "" {
+		return "", futils.CtlogPrefixNotSpecified
+	}
+
+	if instance.Spec.Ctlog.Address != "" {
+		url := instance.Spec.Ctlog.Address
+		if instance.Spec.Ctlog.Port != nil {
+			url = fmt.Sprintf("%s:%d", url, *instance.Spec.Ctlog.Port)
+		}
+		return fmt.Sprintf("%s/%s", url, instance.Spec.Ctlog.Prefix), nil
```
osmman, line 96: I think that there is missing case when port is nil.
It will output `address/prefix` but compare to other results which are `adress:port/prefix` and `http?://ctlog.namespace.svc/prefix`.

I am uncertain what is correct output.

internal/controller/fulcio/actions/deployment.go
```diff
@@ -88,6 +83,30 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1alpha1.Fulcio
 	}
 }
 
+func (i deployAction) resolveCTlogUrl(instance *rhtasv1alpha1.Fulcio) (string, error) {
```
osmman, line 86: We need unit tests for this method with all options.

internal/controller/fulcio/actions/deployment.go
bouskaJ: The port number is optional according to the URL specification (you can rely on the default ports for protocols).

osmman (approved):

PR #1135, opened 2025-06-04
chore(SECURESIGN-1855): Initialize TUF repository with user-provided PVC

bouskaJ:


test/e2e/securesign_1855_test.go
osmman, line 1: To speed up our CI pipeline and improve clarity, please reduce `securesign_1855_test.go` to the minimum required to reproduce the issue. You can accomplish this by deploying only the `Tuf` CRD with a static secret, rather than the entire TAS stack which took min 2minutes.

osmman (changes requested):
The current commit message doesn't follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard. A chore commit type is intended for routine maintenance and tasks that don't modify production code, so feat or fix would be more appropriate here. Please rewrite the commit message to reflect the nature of the changes.

osmman (approved):

PR #1136, opened 2025-06-06
fix: change red hat trusted artifact signer icon

osmman:


osmman:
New icon looks like this in OLM
Screenshot 2025-06-06 at 12 41 10

PR #1149, opened 2025-06-12
fix: changing CA common names

SequeI:
A library I am working with for model signing is extremely strict with self signed CA's, and will not allow them whatsoever. This change makes it possible for TAS to work out the box with no issues.

### DIFF
--- a/config/samples/rhtas_v1alpha1_securesign.yaml
+++ b/config/samples/rhtas_v1alpha1_securesign.yaml
@@ -60,12 +60,12 @@ spec:
         rootCA:
           organizationName: Red Hat
           organizationEmail: jdoe@redhat.com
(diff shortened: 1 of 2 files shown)


JasonPowr (commented):
You will need to update the bundle also, please run 'make bundle'

SequeI:
@osmman @bouskaJ Would like more eyes on this please, I am not entirely sure of the ramifications this could cause for automatic upgrades etc or if this would be an ok change

osmman:
> @osmman @bouskaJ Would like more eyes on this please, I am not entirely sure of the ramifications this could cause for automatic upgrades etc or if this would be an ok change

It should be ok, because samples are used only for new instances.

PR #1174, opened 2025-06-27
Changing image mapping Trillian Create Tree to createTree

tommyd450:
## Summary by Sourcery

Rename the Trillian Create Tree image reference key to simplify naming, updating all related constants and config entries.

Enhancements:
- Rename the TrillianCreateTree image constant to use RELATED_IMAGE_CREATETREE in code.
- Update kustomization and manager_images_patch configs to reference RELATED_IMAGE_CREATETREE instead of RELATED_IMAGE_TRILLIAN_CREATE_TREE.


osmman (changes requested):
There is a application argument which need  to be changed too. 
https://github.com/securesign/secure-sign-operator/blob/main/cmd/main.go#L109

petrpinkas:
How about https://github.com/securesign/secure-sign-operator/blob/tdalton/createTreemappingFix/ci/dev-images.sed#L8 @osmman ? Looks good to me, just not sure, why it was not updated :-)

PR #1181, opened 2025-07-03
fix(SECURESIGN-2058): Fix conditions for creating ingress objects

bkabrda:
Ingress objects for Fulcio, Rekor Search UI and TSA will now only be created when they should be. Additionally, tests are added for all CRDs that have externalAccess config that validate this correct behavior.

### DIFF
--- a/internal/controller/fulcio/actions/ingress_test.go
+++ b/internal/controller/fulcio/actions/ingress_test.go
@@ -1,0 +1,59 @@
+package actions
+
+import (
+	"context"
+	"testing"
+
+	. "github.com/onsi/gomega"
+
+	rhtasv1alpha1 "github.com/securesign/operator/api/v1alpha1"
+	"github.com/securesign/operator/internal/constants"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestIngress_CanHandle(t *testing.T) {
+	ctx := context.TODO()
+	g := NewWithT(t)
+	tests := []struct {
+		name           string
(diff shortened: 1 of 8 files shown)


internal/controller/fulcio/actions/ingress_test.go
```diff
@@ -0,0 +1,59 @@
...
+	. "github.com/onsi/gomega"
+
+	rhtasv1alpha1 "github.com/securesign/operator/api/v1alpha1"
+	"github.com/securesign/operator/internal/constants"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestIngress_CanHandle(t *testing.T) {
+	ctx := context.TODO()
+	g := NewWithT(t)
```
osmman, line 16: Initializing `g := NewWithT(t)` once at the start of `TestIngress_CanHandle` can cause issues with `t.Run` subtests. If an assertion fails in a subtest, Gomega's `t.FailNow()` will terminate the parent test, stopping all subsequent sibling subtests from running.

**Solution:**
To allow subtests to fail independently, initialize a new Gomega object `g` within each `t.Ru`n block, using that subtest's `*testing.T` instance.

osmman (changes requested):

internal/controller/fulcio/actions/ingress_test.go
bkabrda: @osmman good point, thanks. I fixed it in a subsequential commit across all the test files I added.

PR #1200, opened 2025-07-17
fix: Fix providing custom CA for the rekor createtree job

bkabrda:
This fixes two issues:
* First, the path provided to the createtree binary was incorrect (missed a file separator)
* Second, the configmap wasn't actually mounted in because no container name was provided to the TrustedCA function.
  * This was possible because the ellipsis argument to TrustedCA would allow the caller to pass in no container names; I fixed this by adding an extra argument before ellipsis, so that the compilation would fail if the user doesn't provide at least one container name.

## Summary by Sourcery

Fix custom CA provisioning for the Trillian create-tree (rekor) job by correcting path construction and guaranteeing config map mounting

Bug Fixes:
- Construct the CA bundle path with filepath.Join to include path separators
- Mount the custom CA config map to the createtree container by enforcing a required container name in TrustedCA helpers and passing it where needed

Enhancements:
- Introduce createTreeContainerName constant and update TrustedCA function signatures to require at least one explicit container name

### DIFF
--- a/internal/utils/tls/ensure/client.go
+++ b/internal/utils/tls/ensure/client.go
@@ -9,11 +9,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// TrustedCA mount config map with trusted CA bundle to all pod template's containers.
-func TrustedCA(lor *v1alpha1.LocalObjectReference, containerNames ...string) func(template *corev1.PodTemplateSpec) error {
+// TrustedCA mount config map with trusted CA bundle to the provided pod template's containers.
+func TrustedCA(lor *v1alpha1.LocalObjectReference, containerName string, moreNames ...string) func(template *corev1.PodTemplateSpec) error {
+	// NOTE: the "containerName" argument ensures that this function is never called with
 	return func(template *corev1.PodTemplateSpec) error {
 		for i, c := range template.Spec.Containers {
-			if slices.Contains(containerNames, c.Name) {
+			if slices.Contains(append(moreNames, containerName), c.Name) {
 				env := kubernetes.FindEnvByNameOrCreate(&template.Spec.Containers[i], "SSL_CERT_DIR")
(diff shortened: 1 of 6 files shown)


internal/utils/tls/ensure/client.go
```diff
@@ -9,11 +9,12 @@ import (
...
 
-// TrustedCA mount config map with trusted CA bundle to all pod template's containers.
-func TrustedCA(lor *v1alpha1.LocalObjectReference, containerNames ...string) func(template *corev1.PodTemplateSpec) error {
+// TrustedCA mount config map with trusted CA bundle to the provided pod template's containers.
+func TrustedCA(lor *v1alpha1.LocalObjectReference, containerName string, moreNames ...string) func(template *corev1.PodTemplateSpec) error {
+	// NOTE: the "containerName" argument ensures that this function is never called with
 	return func(template *corev1.PodTemplateSpec) error {
 		for i, c := range template.Spec.Containers {
-			if slices.Contains(containerNames, c.Name) {
+			if slices.Contains(append(moreNames, containerName), c.Name) {
```
osmman, line 17: You are creating a new slice for every iteration of `for` loop which is waste of resources. You can construct slice before loop and reuse it.

internal/utils/tls/tls.go
```diff
@@ -28,7 +29,7 @@ func CAPath(ctx context.Context, cli client.Client, instance objectWithTlsClient
 			err = fmt.Errorf("%s ConfigMap can contain only 1 record", lor.Name)
 			return "", err
 		}
-		return CATrustMountPath + slices.Collect(maps.Keys(cfgTrust.Data))[0], nil
+		return filepath.Join(CATrustMountPath, slices.Collect(maps.Keys(cfgTrust.Data))[0]), nil
```
osmman, line 32: please provide unit test

osmman (changes requested):

PR #1418, opened 2025-10-31
feat: SECURESIGN-3183: add operator config and auto-detect platform

knrc:
Add an operator configuration resource, auto detect the platform and create the resource on first startup.

Restarts will retrieve the configuration from the configuration resource, which should remain static.

## Summary by Sourcery

Add a new operator configuration resource (TASOperatorConfig) with startup logic to auto-detect or apply user-specified platform (OpenShift vs Kubernetes) and persist it, update CRDs with EnvFiles support and documentation fixes, remove per-environment overlays, and update RBAC, build and CI workflows accordingly

New Features:
- Introduce TASOperatorConfig custom resource and API types to store operator configuration
- Implement platform detection logic in the operator to auto-detect or respect command-line flags and persist the result in TASOperatorConfig

Enhancements:
- Add fileKeyRef support under EnvFiles feature gate in CRDs and update CRD documentation (WeightedPodAffinityTerm semantics, env var naming, feature gate descriptions)
- Introduce IsFlagProvided utility to detect explicit flag or environment variable usage
- Remove environment-specific overlays and consolidate default configuration path

Build:
- Adjust Makefile to remove OPENSHIFT default and set unified config default

CI:
- Remove hardcoded OPENSHIFT environment flags from GitHub workflows

Chores:
- Update RBAC rules and controller permissions for TASOperatorConfig
- Generate deepcopy methods for new TASOperatorConfig types
- Add TASOperatorConfig CRD to CRD kustomization


knrc:
This is interesting feedback for a number of reasons, but lets take these one at a time.

> Hello, I believe it introduces an architectural anti-pattern that leads to unnecessary complexity and fragility.
> For this reason, I am recommending we do not go down this path and opt for a simpler solution instead.

This is certainly an unconventional use of CRDs, but it's not without precedent.  Personally, I wouldn't call this an anti-pattern.

> ### TASOperatorConfig
> The core design issue is that the operator is configured by a **TASOperatorConfig** resource that it also manages. This creates a "chicken-and-egg" circular dependency:
> 
> 1. The operator's main.go logic depends on this resource to function.
> 2. The operator itself is responsible for creating this resource on its first startup.

A "chicken-and-egg" problem would be one where the operator could not function without the existence of the resource and the resource could only be created by the operator.  This is not the case with what is being suggested.  The resource is just storage for the decision, it's no different from using a database but obviously relies on etcd rather than mysql/postgres.

> This results in a fragile, non-deterministic startup process. This core anti-pattern is further highlighted by the need to enforce a static, hard-coded name (**rhtas-operator-config**) to create a "singleton" resource. This is a strong indicator that a CRD is the wrong tool for the job, as it's misusing a CRD as a simple key-value store—a role better suited for a **ConfigMap**.

In this instance a ConfigMap and a custom resource are no different, they provide the same functionality.  I have no preference for whether we create a ConfigMap or a resource.  The differences between the two approaches are
- CRD is structured, ConfigMap is unstructured
- A ConfigMap can be used as a volume and 'mounted' into a pod

We do not need the second capability, and I would favour structured over unstructured every time but that is a personal choice and not a hard preference.  I tend to only use a ConfigMap when it needs to be used as a Volume.  Both CRD and ConfigMap approaches would be accessible as soon as the API server starts, so once created there would be no race.

> ### Operator Configuration
> The Kubernetes ecosystem already has solutions for passing configuration to operators (OLM, Helm, Kustomize). These tools are designed to template the operator's Deployment with the correct platform-specific configuration (via env vars or args) before it's applied. This approach bypasses these standard mechanisms, reinventing a less-reliable configuration system inside the operator.

Except that the whole point of this issue is that the 'reliable' approach isn't reliable.  Passing in the parameter via OLM doesn't work, we just moved from one problem to another when choosing the parameter approach.  The parameter approach is always right in some use cases, always wrong in other use cases.  Both approaches fail, hence the problem we are exploring here.

> ### Detection flaw
> The timing window (e.g., hibernation) is the most critical flaw and the primary reason we moved to our current static configuration. This was an existing, reproducible issue (especially on large clusters with lot of operators), and while this PR makes it less probable, it does not solve it. This new "detect-and-cache" approach has timing flaw. Scenario:

I haven't seen a reproducer for this issue, please provide the details.  The sequence is understood, and the timing window you are documenting (small race window between installation and hibernation) is an issue.  In fact this single instance is the only scenario I can think of where we would still encounter the problem, while solving all the original reports of the issue.

> This dynamic, internal detection logic is fundamentally racy and non-deterministic.

That would depend on the detection logic we choose to implement.  At the moment it's looking for `openshift.io` resources, but perhaps this is where we should be focussing.

> ### Recommendations
> The ideal path forward is to first explore of OCP's functionality. We should investigate if OCP/Kubernetes offers a standard, platform-level way to enforce pod scheduling order (e.g., delaying our manager pod until core OCP services are ready), which would solve the root cause.

OCP does have a notion of run-levels, however this is not a kubernetes feature.

> If that exploration does not yield a useful, standard solution, our default should be to keep our current static configuration and avoid platform-detection logic. This static parameter is a dead-simple 100% predictable behavior. In that case, we would focus on properly documenting this existing solution for our OLM and Kustomize manifests and making minor UX adjustments.

Yes, it's 100%, predictably wrong in the use cases we are looking at for EKS.  Staying with this approach is not really an option, we need to find a solution which works 100% of the time.

Talking through this has given me a better idea for detection, certainly one that would work in kubernetes and should work in OpenShift.  I'll spin up an OCP instance to look.

I suspect we can get this to work in all scenarios, if we fix the detection logic.

knrc:
> I suspect we can get this to work in all scenarios, if we fix the detection logic.

@osmman the detection logic used in this PR will work for all OCP/k8s instances, since there are openshift.io resources which are local.  Just in case this changes in future releases of OCP, we can look at the aggregation and get a guaranteed result.

Given the auto-detection will always work, the race between install/hibernation is irrelevant.  We can still simplify this to always detect on startup, I'll make those changes now.

knrc:
@osmman Looking at the original code, the race problem exists because the wrong resource was chosen for detection.  If a local one had been chosen then it would have always worked.

knrc:
Closing this to open a new PR, based on discussions in this one

PR #1421, opened 2025-11-04
feat: add auto-detect for OpenShift platform

knrc:
This replaces #1418, implementing auto detection based on the registered API Services.

The original race condition was caused by kubernetes  aggregated API servers, with the resources chosen by the original detection method supported by the openshift-apiserver.  Since the openshift-apiserver was not guaranteed to have been running at the start of the operator, this led to the race.

If a `Local` resource had been chosen instead, i.e. one supported by kube-apiserver, there would have been no race.

This PR removes the config CRD proposed in #1418, and implements a detection method based on the APIServices.  The detection will query all APIServices and search for one which has an associated service (i.e. aggregated endpoint) with the openshift-apiserver name or namespace.

I checked this behaviour on a number of OCP versions, and also on kind.  The versions and APIServices are
- openshift-apiserver/api
-- AWS 4.12
-- AWS 4.20
- default/openshift-apiserver
-- rosa 4.16
-- hypershift 4.19
-- hypershift 4.20

## Summary by Sourcery

Add APIService-based auto-detection for OpenShift platform, replacing manual flag configuration and outdated overlays, while updating RBAC, flags, CI workflows, and build dependencies.

New Features:
- Automatically detect OpenShift platform via aggregated APIService matching instead of manual configuration

Bug Fixes:
- Resolve race condition in platform detection by querying APIService resources

Enhancements:
- Introduce OpenshiftAPIServerName flag and IsFlagProvided utility for conditional platform logic
- Consolidate kustomize overlays and remove deprecated env config directories
- Extend operator RBAC to grant get/list/watch permissions on apiservices

Build:
- Add k8s.io/kube-aggregator dependency

CI:
- Remove explicit OPENSHIFT environment variables from GitHub workflows to rely on auto-detection

Chores:
- Remove deprecated OpenShift config CRDs and related env overlays


knrc:
Moving this PR to the `securesign` repo, so the checks will work

PR #1439, opened 2025-11-10
fix: SECURESIGN-3246: restrict SBJ to OpenShift environments

knrc:
This PR adds guards around the SBJ integration, since this requires resources within the `openshift-monitoring` namespace

## Summary by Sourcery

Guard SBJ integration to only operate in OpenShift environments and disable related actions on non-OpenShift clusters

Enhancements:
- Add kubernetes.IsOpenShift guard around Role and RoleBinding cleanup in openshift-monitoring namespace
- Restrict segment backup CronJob, backup Job, and RBAC actions to OpenShift clusters only


osmman:
Segment Backup Job has been removed from operator by https://github.com/securesign/secure-sign-operator/pull/1457

PR #1578, opened 2026-02-11
fix: update envtest version

bouskaJ:


internal/controller/tuf/tuf_controller_test.go
osmman: Is the change needed due update of envtest?

osmman (approved):

PR #1684, opened 2026-03-25
test: TSA key rotation

bouskaJ:


bouskaJ:
this is expected to fail on Konflux because we still use cosign 2. We should fix that ASAP https://github.com/securesign/pipelines/blob/main/pipelines/integration-test/rhtas-fbc-e2e.yaml#L364

osmman (approved):

PR #1723, opened 2026-04-01
feat: TUF repository migration job to handle v1.3.x to v1.4.0 upgrades

bouskaJ:
Refs: [SECURESIGN-4122](https://redhat.atlassian.net/browse/SECURESIGN-4122)


bouskaJ:
Ignore the `crdify` linter error - The tuf.clog.preffix was introduced in https://github.com/securesign/secure-sign-operator/commit/a01b2eabe47f6b1f89e5a065918394b01ac0a9fd but it was never correctly defaulted in Securesign. That means, we already made the API changes and this is only a fix.

We agreed to take the `1.4.0` release as a baseline and we will NOT ignore the crdify issues in next releases.

internal/controller/predicate/ignore_failure_predicate.go
```diff
@@ -0,0 +1,31 @@
...
+		if !ok {
+			return true
+		}
+
+		c := meta.FindStatusCondition(conditionsAwareObject.GetConditions(), constants.ReadyCondition)
+
+		if c != nil && c.Status == metav1.ConditionFalse && c.Reason == state.Failure.String() {
+			return false
+		}
+		return true
```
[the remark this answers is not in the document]
osmman: This introduce bug when user will unable fix misconfigurations because any edit of spec will be ignored and reconciliation skipped.

bouskaJ:
@qodo /review

internal/apis/TasService.go
```diff
@@ -0,0 +1,27 @@
...
+// TasService represents a tas service that can be used to get the address, port and suffix.
+type TasService interface {
+	GetAddress() string
+	GetPort() *int32
+
+	SetAddress(address string)
+	SetPort(port *int32)
+}
+
+func ServiceAsUrl(service TasService) (string, error) {
```
osmman: This function is not behave correctly. For example:
```go
ServiceAsUrl(&v1alpha1.FulcioService{Address: "path.org", Port: ptr.To(int32(8080))}}) == "//:8080/path.org"
```

it should be path.org:8080

https://github.com/securesign/secure-sign-operator/pull/1723#issuecomment-4197461168

internal/controller/tuf/constants/constants.go
```diff
@@ -7,8 +7,12 @@ const (
...
 	Port                = 8080
 	InitJobName         = "tuf-repository-init"
+	MigrationJobName    = "tuf-repository-migration"
 	RBACInitJobName     = "tuf-repository-init"
 	ContainerName       = "tuf-server"
 	VolumeName          = "repository"
 	RepositoryCondition = "repository"
+
+	RepositoryVersionAnnotation = "rhtas.redhat.com/tuf-version"
+	TufVersionV1                = "v1"
```
osmman, line 17: What is driver of this version number? It is something from tuf specification of other source?

internal/apis/TasService.go
bouskaJ: Good point. It is because the `path.org.`  format does not follow the RFC https://datatracker.ietf.org/doc/html/rfc3986#autoid-19

I've added a fix

internal/controller/tuf/constants/constants.go
bouskaJ: no, I was thinking about how to mark the migrated version and I decided to call it `v1`.  It is TAS specific marker open to add new generations.

internal/apis/TasService.go
osmman: I am currently do not remember how we are exactly using `Address` fields maybe solution is to enforce shema via validation.

internal/controller/predicate/ignore_failure_predicate.go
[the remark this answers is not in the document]
bouskaJ: resolved by https://github.com/securesign/secure-sign-operator/pull/1723/changes/5ea4e12d6c3b52dbf04600bfb9740c48b449e4be

internal/controller/tuf/constants/constants.go
osmman: Is it based on this versioning? https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto#L16

internal/controller/tuf/utils/service_url_test.go
```diff
@@ -0,0 +1,182 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/securesign/operator/api/v1alpha1"
+	tsa "github.com/securesign/operator/internal/controller/tsa/actions"
+	v1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var (
+	ctx = context.TODO()
```
osmman: Do not use TODO context, use context from `testing.T.Context()`.

internal/controller/tuf/actions/migration_job_test.go
```diff
@@ -0,0 +1,283 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var migrateJobTestContext = context.TODO()
```
osmman: Use context from testing.T

internal/controller/tuf/utils/tuf_migration_v1.sh
```diff
@@ -0,0 +1,281 @@
...
+fix_valid_for_start() {
+    local target_file="$1"
+    echo "Checking for missing validFor.start entries..."
+    
+    local missing_count
+    missing_count=$(python3 -c "
+import json, sys
+data = json.load(open(sys.argv[1]))
+count = 0
+for key in ['tlogs', 'ctlogs']:
```
osmman, line 72: What about `certificateAuthorities` and `timestampAuthorities`?

internal/controller/tuf/utils/tuf_migration_v1.sh
```diff
@@ -0,0 +1,281 @@
...
+        return
+    fi
+
+    echo " -> signing_config.v0.2.json is missing. Generating it via Cosign..."
+
+    local cmd=("cosign" "signing-config" "create")
+
+    # Using the new TAS_START constant everywhere
+    if [ -n "$FULCIO_URL" ]; then
+        cmd+=("--fulcio=url=$FULCIO_URL,api-version=1,start-time=$TAS_START,operator=sigstore.dev")
```
osmman, line 196: **Sigstore.dev** will not be operator of customer's logs. We will need to change value to something else to not collide with public infrastructure.

internal/controller/tuf/utils/service_urls.go
```diff
@@ -0,0 +1,112 @@
...
+	protocol := "http"
+	if len(ingress.Spec.TLS) > 0 {
+		protocol = "https"
+	}
+	return fmt.Sprintf("%s://%s", protocol, ingress.Spec.Rules[0].Host), nil
+}
+
+func ResolveOIDCIssuers(ctx context.Context, c client.Client, namespace string) []string {
+	fulcioList := &rhtasv1alpha1.FulcioList{}
+	if err := c.List(ctx, fulcioList, client.InNamespace(namespace)); err != nil {
```
osmman, line 95: This could be problematic if some one deploy two instances of Securesign or Fulcio in one namespace it can cause unexpected behavior (it can be common problem in our operator).

Example:
```
kind: Fulcio
metadata:
   name: fulcio-1
   namespace: space
spec:
  externalAccess: 
    enabled: true
  config:
    OIDCIssuers:
      - Issuer: 'https://your-oidc-issuer-url'
---
kind: Fulcio
metadata:
   name: fulcio-2
   namespace: space
spec:
  externalAccess: 
    enabled: true
    host: fulcio2.test.org
  config:
    OIDCIssuers:
      - Issuer: 'https://oauth2.sigstore.dev/auth'
---
kind: Tuf
metadata:
  name: tuf
spec: ...  
```

osmman: One solution could be interrupt migration if more than one Fulcio resource is found. User will have to decide if to manualy migrate or solve problem and retry.

bouskaJ: This part of code was just moved from https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/tuf/actions/tuf_init_job.go#L166. It is used in the `init-job` too (cc @SequeI ) Can you open a new issue to solve this problem? I do not want to make this PR even more complicated.

internal/controller/tuf/utils/tuf_migration_v1.sh
bouskaJ: I know this is an issue and I used what the `init-job` generate to be consistent (https://github.com/securesign/tough/blob/develop/tuftool/src/rhtas.rs#L719). :+1:Would you mind to open a Jira to fix that?  @SequeI can you do the fix it in the `tuffer` repository? I can fix the operator part but we should be consistent here.

internal/controller/tuf/utils/tuf_migration_v1.sh
bouskaJ: this was taken from https://docs.google.com/document/d/1DZ8VnF6Dn18-7kmPuLemZDgePLtY5OsGLYvy4S0sU4A/edit?tab=t.hyrs5lpiigcc#heading=h.3bf47ewo04t9 @kdacosta0 can you explain?

osmman: Here is Sigstore's public instance metadata for example https://github.com/sigstore/root-signing/blob/main/targets/trusted_root.json

internal/controller/tuf/utils/tuf_migration_v1.sh
osmman: https://redhat.atlassian.net/browse/SECURESIGN-4219

internal/controller/tuf/utils/service_urls.go
SequeI: https://redhat.atlassian.net/browse/SECURESIGN-4220

internal/controller/tuf/utils/tuf_migration_v1.sh
bouskaJ: Will be fixed in separate PR

osmman (approved):

PR #1724, opened 2026-04-01
fix: skip TSA ingress resolution when TSA key is not configured

miyunari:
## Summary
- When `signingConfigURLMode: external`, `resolveServiceURLs()` unconditionally resolves the TSA ingress, causing TUF init to fail if TSA is not deployed
- This fix makes TSA ingress resolution conditional on whether `tsa.certchain.pem` is present in the configured TUF keys
- Adds unit tests for the new `hasTSAKey` helper

## Details
The bug manifests when a user deploys RHTAS without TSA (removing `tsa.certchain.pem` from TUF keys). In `external` mode, `resolveServiceURLs()` always tries to look up the TSA ingress, which doesn't exist, causing TUF initialization to fail with an ingress-not-found error.

The fix aligns `resolveServiceURLs()` with the existing pattern in `utils/tuf_init_job.go`, where TSA args are only added when the `tsa.certchain.pem` key is present in the spec.

Fixes: SECURESIGN-4125

## Test plan
- [ ] Unit tests pass (`go test ./internal/controller/tuf/actions/...`)
- [ ] E2E: Deploy RHTAS without TSA, verify TUF init succeeds with `signingConfigURLMode: external`
- [ ] E2E: Deploy RHTAS with TSA, verify TUF init still resolves TSA ingress correctly

---
Jira: https://issues.redhat.com/browse/SECURESIGN-4125

### DIFF
--- a/internal/controller/tuf/actions/tuf_init_job.go
+++ b/internal/controller/tuf/actions/tuf_init_job.go
@@ -134,8 +134,17 @@ func (i initJobAction) resolveServiceURLs(ctx context.Context, instance *rhtasv1
 	}{
 		{&instance.Spec.Fulcio.Address, fulcio.DeploymentName, ""},
 		{&instance.Spec.Rekor.Address, rekor.ServerDeploymentName, ""},
-		{&instance.Spec.Tsa.Address, tsa.DeploymentName, tsa.TimestampPath},
 	}
+
+	// Only resolve TSA ingress if tsa.certchain.pem key is configured
+	if hasTSAKey(instance.Spec.Keys) {
+		services = append(services, struct {
+			address     *string
+			ingressName string
+			suffix      string
+		}{&instance.Spec.Tsa.Address, tsa.DeploymentName, tsa.TimestampPath})
+	}
+
 	for _, svc := range services {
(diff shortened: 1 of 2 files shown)


internal/controller/tuf/actions/tuf_init_job.go
```diff
@@ -148,6 +157,15 @@ func (i initJobAction) resolveServiceURLs(ctx context.Context, instance *rhtasv1
 	return nil
 }
 
+func hasTSAKey(keys []rhtasv1alpha1.TufKey) bool {
```
osmman, line 160: Hi @miyunari, I think that same problem is present for other services too (Fulcio, Rekor). Example: Namespace has only Rekor and TUF (valid deployment).

```
kind: Rekor
metadata:
  name: rekor-sample
spec:
  externalAccess:
    enabled: true
---
kind: Tuf
metadata:
  name: tuf
spec:
  signingConfigURLMode: external
  keys:
    - name: rekor.pub
```

miyunari:
Continue here:
- https://github.com/securesign/secure-sign-operator/pull/1725

PR #1725, opened 2026-04-07
fix: skip TSA ingress resolution when TSA key is not configured

miyunari:
When signingConfigURLMode is "external", resolveServiceURLs() unconditionally attempts to resolve the TSA ingress. If TSA is not deployed (i.e., tsa.certchain.pem is not in the TUF keys), the ingress lookup fails and blocks TUF initialization entirely.

This change makes TSA ingress resolution conditional on whether tsa.certchain.pem is present in the configured TUF keys, matching the existing pattern in the TUF init job args builder.

Fixes: SECURESIGN-4125

### DIFF
--- a/internal/controller/tuf/actions/tuf_init_job.go
+++ b/internal/controller/tuf/actions/tuf_init_job.go
@@ -134,8 +134,17 @@ func (i initJobAction) resolveServiceURLs(ctx context.Context, instance *rhtasv1
 	}{
 		{&instance.Spec.Fulcio.Address, fulcio.DeploymentName, ""},
 		{&instance.Spec.Rekor.Address, rekor.ServerDeploymentName, ""},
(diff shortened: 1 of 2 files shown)


miyunari:
Continue on this branch. Since CI seems not to work as expected from a fork:
See https://github.com/securesign/secure-sign-operator/pull/1724#discussion_r3043601155 regarding the latest change.

osmman:
Hi @miyunari ! Thanks so much for putting this together. It turns out that during the recent big changes caused by introduction of TUF migration job, @bouskaJ actually integrated this fix into the codebase as part of https://github.com/securesign/secure-sign-operator/pull/1723. Since the bug is now resolved on the latest version, I’m going to close this one out. Really appreciate the effort here!

PR #1732, opened 2026-04-12
fix: guard against nil RootKeySecretRef in TUF init job

miyunari:
## Summary
- During an upgrade from an older operator version, `RootKeySecretRef` may be nil if the field wasn't present in the original CRD. `EnsureTufInitJob` dereferences it without a nil check, causing a panic.
- Added a nil guard that returns a clear error (`rootKeySecretRef is not set`) instead of panicking.
- Added unit tests for both the nil case (safe error state) and the normal case (correct `--export-keys` argument).

## Test plan
- [x] `TestEnsureTufInitJob_NilRootKeySecretRef` — verifies nil `RootKeySecretRef` produces a clean error, no panic
- [x] `TestEnsureTufInitJob_WithRootKeySecretRef` — verifies normal operation and correct job args
- [x] All existing TUF controller and action tests pass


miyunari:
Superseded by #1733 (rebased onto main)

PR #1733, opened 2026-04-12
fix: guard against nil RootKeySecretRef in TUF init job

miyunari:
## Summary
- During an upgrade from an older operator version, `RootKeySecretRef` may be nil if the field wasn't present in the original CRD. `EnsureTufInitJob` dereferences it without a nil check, causing a panic.
- Added a nil guard that returns a clear error (`rootKeySecretRef is not set`) instead of panicking.
- Added unit tests for both the nil case (safe error state) and the normal case (correct `--export-keys` argument).

Supersedes #1732
Related: SECURESIGN-4125

## Test plan
- [x] `TestEnsureTufInitJob_NilRootKeySecretRef` — verifies nil `RootKeySecretRef` produces a clean error, no panic
- [x] `TestEnsureTufInitJob_WithRootKeySecretRef` — verifies normal operation and correct job args
- [x] All existing TUF controller and action tests pass

### DIFF
--- a/internal/controller/tuf/utils/tuf_init_job.go
+++ b/internal/controller/tuf/utils/tuf_init_job.go
@@ -23,6 +23,10 @@ const (
 
 func EnsureTufInitJob(instance *rhtasv1alpha1.Tuf, sa string, labels map[string]string, oidcIssuers []string) func(*batchv1.Job) error {
 	return func(job *batchv1.Job) error {
+		if instance.Spec.RootKeySecretRef == nil {
+			return fmt.Errorf("rootKeySecretRef is not set")
+		}
+
 		// prepare args
 		args := []string{"--operator", constants.OperatorName, "--export-keys", instance.Spec.RootKeySecretRef.Name}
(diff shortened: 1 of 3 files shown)


internal/controller/tuf/utils/tuf_init_job.go
```diff
@@ -23,6 +23,10 @@ const (
 
 func EnsureTufInitJob(instance *rhtasv1alpha1.Tuf, sa string, labels map[string]string, oidcIssuers []string) func(*batchv1.Job) error {
 	return func(job *batchv1.Job) error {
+		if instance.Spec.RootKeySecretRef == nil {
+			return fmt.Errorf("rootKeySecretRef is not set")
```
bouskaJ, line 27: Throw a Terminal error - this is a configuration error and we should move to `Failure` status.

bouskaJ (changes requested):

lance:
@miyunari what is the status of this PR? Can it be updated or closed?

PR #1752, opened 2026-04-27
feat: add gRPC client-side load balancing for trillian-logserver

kdacosta0:
## Summary
- Switch `trillian-logserver` service to headless to enable gRPC client-side 
  load balancing (round_robin), fixes severely uneven CPU distribution 
  across replicas caused by the default `pick_first` policy
- Add `grpc_default_service_config` flag to `rekor-server` for round_robin.
- Handle ClusterIP -> headless migration (ClusterIP is immutable, so existing 
  services are deleted and recreated)

## Test
- Verified with rhtas-benchmark: `trillian-logserver` load balances evenly 
  across replicas after the change
- Upgrade from previous version tested,  migration from ClusterIP to 
  headless service works correctly

## Notes
- `ctlog` already load balances by [default](https://github.com/securesign/certificate-transparency-go/blob/727cb5d8e6c477051497b587b853899b54b6a1d9/trillian/ctfe/ct_server/main.go#L208), no changes needed 
- `fulcio` -> `ctlog` uses HTTP/2 via a regular ClusterIP service; improving 
  that path is out of scope for this fix

## Known limitation
New `trillian-logserver` replicas will not receive traffic until an existing connection is lost, which triggers DNS re-resolution

ref: [SECURESIGN-2935](https://redhat.atlassian.net/browse/SECURESIGN-2935)

### DIFF
--- a/internal/controller/rekor/actions/server/deployment.go
+++ b/internal/controller/rekor/actions/server/deployment.go
@@ -140,6 +140,7 @@ func (i deployAction) ensureServerDeployment(instance *rhtasv1alpha1.Rekor, sa s
 			"--trillian_log_server.address", instance.Spec.Trillian.Address,
 			"--trillian_log_server.port", strconv.Itoa(int(*instance.Spec.Trillian.Port)),
 			"--trillian_log_server.sharding_config", "/sharding/sharding-config.yaml",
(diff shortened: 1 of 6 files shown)


kdacosta0:
Waiting for https://github.com/sigstore/rekor/pull/2812

osmman:
@kdacosta0 fix https://github.com/sigstore/rekor/pull/2812 should be in downstream

osmman (approved):

PR #1755, opened 2026-04-28
fix(SECURESIGN-4123): reduce cache memory footprint with label selectors

kdacosta0:
## Summary

Restrict `controller-runtime` informer caches for high-cardinality core types (`Deployment`, `ReplicaSet`, `StatefulSet`, `Pod`, `Service`, `Ingress`, `CronJob`) to only objects labeled `app.kubernetes.io/part-of: trusted-artifact-signer`. Without this, each `Owns()` watch caches ALL objects of that type cluster-wide, causing OOM on production clusters

Bump memory limit to 512Mi and request to 128Mi

Peak RSS: >256 MiB (OOM) → ~80 MiB with all filters active

ref: [SECURESIGN-4123](https://redhat.atlassian.net/browse/SECURESIGN-4123)

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -168,8 +168,6 @@ func main() {
 	// Restrict informer caches for high-cardinality core types to only objects managed by this operator.
 	// Without this, each Owns() watch caches ALL objects of that type cluster-wide, causing OOM on
 	// production clusters with hundreds of Deployments/Services from other workloads (SECURESIGN-4123).
-	// Secrets and ConfigMaps are intentionally excluded: user-provided refs (e.g. Spec.RootKeySecretRef,
-	// Spec.Signer.File.PrivateKeyRef, Spec.Db.DatabaseSecretRef) may lack operator labels.
 	operatorLabelSelector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
 		MatchLabels: map[string]string{
 			labels.LabelAppPartOf: constants.AppName,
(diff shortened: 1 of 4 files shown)


cmd/main.go
```diff
@@ -158,8 +165,32 @@ func main() {
...
+		Label: operatorLabelSelector,
+	}
+
 	cacheOpts := cache.Options{
-		ByObject: map[client.Object]cache.ByObject{},
+		ByObject: map[client.Object]cache.ByObject{
+			&appsv1.Deployment{}:    operatorCacheSelector,
+			&appsv1.ReplicaSet{}:    operatorCacheSelector,
+			&appsv1.StatefulSet{}:   operatorCacheSelector,
+			&corev1.Pod{}:           operatorCacheSelector,
```
osmman: We will need to propagate label to pods which are generated by PodTemplateSpec other way operator will not be able to access them. I found that backfill redis cronjob and Createtree/Updatetree Job is missing it.

cmd/main.go
```diff
@@ -158,8 +165,32 @@ func main() {
...
+			&corev1.Service{}:       operatorCacheSelector,
+			&networkingv1.Ingress{}: operatorCacheSelector,
+			&batchv1.CronJob{}:      operatorCacheSelector,
```
osmman: Why Job is not needed?

kdacosta0: It is needed, I forgot about it. And also pvc I believe

osmman: That could be more problematic when user provide it's own PVC (not managed by operator). These will be invisible to our operator.

kdacosta0: I tried it with my own pvc and it found it

User provided returns without `client.Get` https://github.com/securesign/secure-sign-operator/blob/main/internal/action/pvc/action.go#L75-L84 and I believe it is then referenced via `ClaimName` by deployments

But still it could be a maintenance risk

osmman (approved):

PR #1756, opened 2026-04-29
ci: wait for namespace deletion in e2e tests

osmman:
Refs: [SECURESIGN-3862](https://redhat.atlassian.net/browse/SECURESIGN-3862)


tommyd450 (approved):

osmman:
Rejecting: it is not solving core problem and cause more problems in CI.

PR #1777, opened 2026-05-05
fix: Fix internal signing config

bouskaJ:


bouskaJ:
We usually ignore Crdify issues because we try to be careful with API changes. However, constantly upgrading the API without introducing a new version and proper migration steps causes our operator to be non-upgradable under OLMv1, which has much stricter API validation. We discussed this on the team call and agreed that version 1.4.0 will be our new baseline, and we will follow best practices for API migration moving forward. @osmman I believe we should introduce a new API version along with the conversion webhooks. WDYT?

api/v1alpha1/tuf_types.go
```diff
@@ -24,9 +24,9 @@ type TufSpec struct {
 	// Define whether you want to export service or not
 	ExternalAccess ExternalAccess `json:"externalAccess,omitempty"`
 	// Controls which URLs are used in the signing config:
-	// "external" (default) resolves URLs from Ingress routes,
-	// "internal" uses internal Kubernetes service URLs.
-	//+kubebuilder:default:=external
+	// "external"  resolves URLs from Ingress routes,
+	// "internal" (default) uses internal Kubernetes service URLs.
+	//+kubebuilder:default:=internal
```
osmman: we should use enum

internal/controller/tuf/utils/service_urls.go
```diff
@@ -110,3 +121,14 @@ func ResolveOIDCIssuers(ctx context.Context, c client.Client, namespace string)
...
+
+func isTlsService(ctx context.Context, c client.Client, serviceNamespacedName types.NamespacedName) (bool, error) {
+	service := &corev1.Service{}
+	if err := c.Get(ctx, serviceNamespacedName, service); err != nil {
+		return false, err
+	}
+	if _, ok := service.Annotations[annotations.TLS]; ok {
+		return true, nil
+	}
+	return false, nil
```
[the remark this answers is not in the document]
osmman: @bouskaJ please find different method than TLS annotation. This will not work if deployed on EKS for example or with own TLS certificates configured on services.

internal/controller/tuf/actions/tuf_init_job.go
```diff
@@ -88,14 +88,32 @@ func (i initJobAction) jobPresent(ctx context.Context, job *v2.Job, instance *rh
...
-		return i.Error(ctx, fmt.Errorf("fail to resolve service url: %w", err), instance)
+		err = fmt.Errorf("fail to resolve service url: %w", err)
+		return i.Error(ctx, err, instance,
+			metav1.Condition{
+				Type:    tufConstants.RepositoryCondition,
+				Status:  metav1.ConditionFalse,
+				Reason:  state.Pending.String(),
+				Message: err.Error(),
+			},
+		)
```
osmman, line 116: **Inconsistent error condition between init job and migration job**

This path sets `tufConstants.RepositoryCondition` with `Reason: state.Pending`, while the equivalent path in `migration_job.go:128-135` sets `constants.ReadyCondition` with `Reason: state.Initialize`.

Both handle the same failure (service URL resolution), but:
- **Init job** → only sets `RepositoryCondition` (internal sub-condition, not directly visible via `kubectl get`)
- **Migration job** → sets `ReadyCondition` (user-visible top-level condition)

A user hitting a URL resolution failure during initial install won't see any change on the `Ready` condition — only the `RepositoryCondition` is updated, which is harder to discover.

**Recommendation**: Align both paths to use the same condition type — either `ReadyCondition` or `RepositoryCondition` — so the behavior is consistent regardless of which code path is hit.

test/e2e/support/tas/cosign/cosign_local.go
```diff
@@ -40,36 +48,45 @@ func tryCosignSign(ctx context.Context, targetImageName, fulcioUrl, rekorUrl, ts
...
 	Eventually(func(ctx context.Context) error {
-		return clients.Execute("cosign", "initialize", "--mirror="+tufUrl, "--root="+tufUrl+"/root.json")
+		return clients.Execute("cosign", "initialize", "--mirror="+c.tufUrl, "--root="+c.tufUrl+"/root.json")
 	}).WithContext(ctx).WithPolling(2 * time.Second).Should(Succeed())
 	Eventually(func(ctx context.Context) error {
-		return tryCosignSign(ctx, targetImageName, fulcioUrl, rekorUrl, tsaUrl)
+		return c.Sign(ctx, targetImageName)
 	}).WithContext(ctx).WithPolling(2 * time.Second).Should(Succeed())
-	CosignVerify(ctx, targetImageName, rekorUrl, tsaUrl)
+	Expect(c.Verify(ctx, targetImageName)).To(Succeed())
```
osmman: **`Verify` lost retry for TSA certificate chain fetch**

The old `CosignVerify` wrapped `GetCertificateChain` in `Eventually` to tolerate transient TSA unavailability:
```go
// OLD
Eventually(func(ctx context.Context) error {
    return tsa.GetCertificateChain(ctx, tsaUrl)
}).WithContext(ctx).Should(Succeed())
```

The new `Verify()` (line 67) does a single-shot call, and this `VerifyByCosign` invokes it without retry.

The upgrade test at `upgrade_test.go:229` also calls `cosign.Verify()` directly after an operator upgrade — if TSA is briefly unavailable post-upgrade, the test fails without retry.

**Recommendation**: Either wrap this `Verify` call in `Eventually` here, or wrap the call site in `upgrade_test.go:229` with `Eventually`.

osmman (commented):
Code review — second pass (2 inline comments below)

internal/controller/tuf/actions/tuf_init_job.go
bouskaJ: The `RepositoryCondition` is our legacy code - the InitJob always reported status to `RepositoryCondition` only - this is not changed.

AFAIR We agreed to use `ReadyCondition` on migration because this is just a one shot execution - it is overkill to introduce new condition that becomes part of the API.

I can reuse the `RepositoryCondition` also for migration job if you want so.

osmman (approved):

PR #1779, opened 2026-05-06
refactor(action): separate status persistence from flow control

osmman:
## Summary

- Replace `StatusUpdate(ctx, obj) *Result` with `PersistStatus(ctx, obj) (bool, error)` — status persistence is now a pure side-effect, flow control is always an explicit separate choice
- `PersistStatus` skips the API call when status is unchanged (no-op optimization) and returns a `changed` flag so callers can react: `changed=true` → watch event will fire → `Return()`; `changed=false` → no event → `Requeue()` as safety net
- Add `RequeueAfter(delay)` for explicit polling delays; change `Requeue()` default from 5s to 100ms (safety-net, not polling); polling sites now use `RequeueAfter(5 * time.Second)`
- Migrate all ~124 call sites across 60+ action files
- Fix pre-existing bug: missing `return` before `Error()` in `rekor/server/sharding_config.go`
- Fix `tuf/actions/resolve_keys.go` to persist status before `Continue()` (action independence)
- Add `internal/action/doc.go` with patterns and anti-patterns
- Add `PersistStatus` test coverage including conflict retry
- Add godoc to all `BaseAction` methods and `Action` interface

## Motivation

The old `StatusUpdate` conflated two concerns: persisting status to the API server AND controlling reconciler flow. It always returned `*Result`, making it impossible to distinguish a successful write from a no-op. With the no-op optimization (skip API call when status unchanged), a `return i.StatusUpdate(ctx, instance)` call could silently stall reconciliation — no API call means no watch event means no re-trigger.

## Migration patterns

**State transition** (set final state, let watch event drive next action):
```go
changed, err := i.PersistStatus(ctx, instance)
if err != nil {
    return i.Error(ctx, err, instance)
}
if !changed {
    return i.Requeue()
}
return i.Return()
```

**Polling loop** (wait for external state):
```go
if _, err := i.PersistStatus(ctx, instance); err != nil {
    return i.Error(ctx, err, instance)
}
return i.RequeueAfter(5 * time.Second)
```

**Persist then continue** (intermediate checkpoint):
```go
if _, err := i.PersistStatus(ctx, instance); err != nil {
    return i.Error(ctx, err, instance)
}
return i.Continue()
```

## Test plan

- [x] Unit tests pass (`go test ./internal/...`)
- [x] PersistStatus tests: changed, unchanged (no-op), not-found, conflict retry
- [x] E2e test "Securesign install with certificate generation" — 13/13 passed
- [x] Full CI suite

### DIFF
--- a/internal/action/pvc/action.go
+++ b/internal/action/pvc/action.go
@@ -80,7 +80,7 @@ func (i pvcAction[T]) Handle(ctx context.Context, instance T) *action.Result {
 			Reason:             ReasonSpecified,
 			ObservedGeneration: instance.GetGeneration(),
 		})
-		return i.StatusUpdate(ctx, instance)
+		return i.ReturnOnChange(i.PersistStatus)(ctx, instance)
 	}
 
 	var name string
@@ -107,7 +107,7 @@ func (i pvcAction[T]) Handle(ctx context.Context, instance T) *action.Result {
 			Message:            fmt.Sprintf("Discovered and using  existing default PVC `%s`.", pvc.GetName()),
 			ObservedGeneration: instance.GetGeneration(),
 		})
-		return i.StatusUpdate(ctx, instance)
+		return i.ReturnOnChange(i.PersistStatus)(ctx, instance)
 	}
 
 	if pvcSpec.Size == nil {
@@ -163,5 +163,5 @@ func (i pvcAction[T]) Handle(ctx context.Context, instance T) *action.Result {
 	}
 
 	wrapped.SetStatusPVCName(pvc.Name)
-	return i.StatusUpdate(ctx, instance)
+	return i.ReturnOnChange(i.PersistStatus)(ctx, instance)
 }
--- a/internal/controller/tuf/actions/tuf_init_job.go
+++ b/internal/controller/tuf/actions/tuf_init_job.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"maps"
 	"slices"
+	"time"
 
 	rhtasv1alpha1 "github.com/securesign/operator/api/v1alpha1"
 	"github.com/securesign/operator/internal/action"
@@ -76,7 +77,7 @@ func (i initJobAction) jobPresent(ctx context.Context, job *v2.Job, instance *rh
 					return i.Error(ctx, err, instance)
 				}
 			}
-			return i.StatusUpdate(ctx, instance)
+			return i.ReturnOnChange(i.PersistStatus)(ctx, instance)
 		} else {
 			err := fmt.Errorf("tuf-repository-init job failed")
 			meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
@@ -94,12 +95,10 @@ func (i initJobAction) jobPresent(ctx context.Context, job *v2.Job, instance *rh
 			Reason:  state.Initialize.String(),
 			Message: "waiting for init-repository job to complete",
 		})
-		result := i.StatusUpdate(ctx, instance)
-		if result.Err != nil {
-			return result
+		if _, err := i.PersistStatus(ctx, instance); err != nil {
+			return i.Error(ctx, err, instance)
 		}
-		// ensure that new requeue iteration is triggered even if no status update happened
-		return i.Requeue()
+		return i.RequeueAfter(5 * time.Second)
 	}
 }
 
@@ -147,5 +146,5 @@ func (i initJobAction) ensureInitJob(ctx context.Context, labels map[string]stri
 		Reason:  state.Initialize.String(),
 		Message: msg,
 	})
-	return i.StatusUpdate(ctx, instance)
+	return i.ReturnOnChange(i.PersistStatus)(ctx, instance)
 }
--- a/api/v1alpha1/suite_test.go
+++ b/api/v1alpha1/suite_test.go
@@ -1,11 +1,10 @@
 package v1alpha1
 
 import (
-	"fmt"
 	"path/filepath"
-	"runtime"
 	"testing"
 
+	testenvhelper "github.com/securesign/operator/internal/testing/envtest"
 	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/test"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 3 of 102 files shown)


internal/action/pvc/action.go
```diff
@@ -80,7 +80,14 @@ func (i pvcAction[T]) Handle(ctx context.Context, instance T) *action.Result {
 			Reason:             ReasonSpecified,
 			ObservedGeneration: instance.GetGeneration(),
 		})
-		return i.StatusUpdate(ctx, instance)
+		changed, err := i.PersistStatus(ctx, instance)
+		if err != nil {
+			return i.Error(ctx, err, instance)
+		}
+		if !changed {
```
bouskaJ: Maybe we can implement a helper function RequeueIfNotChanged(changed) that would be less verbose and easier to read.

bouskaJ:
> ### Code Review by Qodo
> `🐞 Bugs (1)` `📘 Rule violations (0)`
> 
> Grey Divider
> 
> 
> Remediation recommended
> 
> 1. Envtest version mismatch `🐞 Bug` `☼ Reliability`
> 
> > Description
> > 
> > The new internal/action TestMain hardcodes envtest binaries to Kubernetes 1.32.0, while an existing
> > envtest suite hardcodes 1.29.1; if the test environment only provides one version, one suite will
> > fail to start. Because TestMain is in package internal/action, all tests in that package (including
> > pure unit tests that don’t use k8sClient) now pay envtest startup cost.
> 
> > Code
> > 
> > `[internal/action/base_action_test.go[R26-33]](https://github.com/securesign/secure-sign-operator/pull/1779/files#diff-40e5b221bda0b3229ab78c12cc0864182f2e601c7864e99d2c9993c7dee3d3cbR26-R33)`
> > ```diff
> > +func TestMain(m *testing.M) {
> > +	testEnv = &envtest.Environment{
> > +		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "config", "crd", "bases")},
> > +		ErrorIfCRDPathMissing: true,
> > +		BinaryAssetsDirectory: filepath.Join("..", "..", "bin", "k8s",
> > +			fmt.Sprintf("1.32.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
> > +	}
> > +
> > ```
> 
> > Evidence
> > 
> > internal/action/base_action_test.go introduces a package-level TestMain that starts envtest and pins
> > assets to 1.32.0. api/v1alpha1/suite_test.go pins envtest assets to 1.29.1, demonstrating the repo
> > now expects multiple envtest versions. internal/action/result_test.go contains unit-only tests (no
> > envtest client usage), but will still run under the internal/action TestMain, increasing runtime and
> > flake surface for that package’s unit tests.
> > 
> > `[internal/action/base_action_test.go[26-54]](https://github.com/securesign/secure-sign-operator/blob/bb67c17aadced982de88bf034eb33294f0346488/internal/action/base_action_test.go/#L26-L54)`
> > `[api/v1alpha1/suite_test.go[42-60]](https://github.com/securesign/secure-sign-operator/blob/bb67c17aadced982de88bf034eb33294f0346488/api/v1alpha1/suite_test.go/#L42-L60)`
> > `[internal/action/result_test.go[1-10]](https://github.com/securesign/secure-sign-operator/blob/bb67c17aadced982de88bf034eb33294f0346488/internal/action/result_test.go/#L1-L10)`
> 
> > Agent prompt
> > 
> > ```
> > The issue below was found during a code review. Follow the provided context and guidance below and implement a solution
> > 
> > ### Issue description
> > `internal/action/base_action_test.go` starts envtest in a package-level `TestMain` and hardcodes a specific envtest asset version (1.32.0). Elsewhere (`api/v1alpha1/suite_test.go`) envtest is hardcoded to a different asset version (1.29.1). This creates a repo-wide expectation that multiple envtest binary versions are available, and it also makes all tests in `internal/action` pay the envtest startup overhead.
> > 
> > ### Issue Context
> > - Adding `TestMain` to `internal/action` affects all tests in that package, including unit-only tests that don’t interact with envtest.
> > - Hardcoding versions in multiple places makes CI/tooling brittle unless it provisions every pinned version.
> > 
> > ### Fix Focus Areas
> > - internal/action/base_action_test.go[26-54]
> > - api/v1alpha1/suite_test.go[42-60]
> > - internal/action/result_test.go[1-10]
> > 
> > ### Suggested fix
> > 1. **Unify envtest asset selection**:
> >   - Prefer using `KUBEBUILDER_ASSETS` (or controller-runtime’s default asset discovery) rather than hardcoding versions in multiple packages; or
> >   - Centralize the version/path computation in a shared test helper package and reuse it everywhere.
> > 2. **Reduce blast radius of envtest startup**:
> >   - Move the envtest-based tests out of `internal/action` into a dedicated subpackage (e.g., `internal/action/envtest` or `internal/action/integration`) so unit tests in `internal/action` aren’t forced to start envtest.
> >   - Alternatively, guard envtest tests behind a build tag so default unit test runs remain fast.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `ⓘ Copy this prompt and use it to remediate the issue with your preferred AI generation tools`
> 
> Grey Divider
> 
> [Qodo Logo](https://www.qodo.ai)

@osmman fix

osmman:
@bouskaJ Fixed by https://github.com/securesign/secure-sign-operator/pull/1779/changes/a7f6593acef4c4bf825a7fe510417640e7f527b2

internal/action/pvc/action.go
osmman: Looking at this again, the if statement should be switched to
```
if changed {
   return i.Return()
}
...
...
return i.Requeue() # or i.Continue()
```

If the status changes, as that update will trigger a new watch event and a subsequent reconciliation. If the status hasn't changed, we can continue execution or finish the action with a `Requeue` or `Continue` result.

osmman: @bouskaJ I added ReturnOnChange helper function

internal/controller/tuf/actions/tuf_init_job.go
```diff
@@ -94,11 +94,9 @@ func (i initJobAction) jobPresent(ctx context.Context, job *v2.Job, instance *rh
...
 			Message: "waiting for init-repository job to complete",
 		})
-		result := i.StatusUpdate(ctx, instance)
-		if result.Err != nil {
-			return result
+		if _, err := i.PersistStatus(ctx, instance); err != nil {
+			return i.Error(ctx, err, instance)
 		}
-		// ensure that new requeue iteration is triggered even if no status update happened
 		return i.Requeue()
```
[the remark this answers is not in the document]
bouskaJ: @osmman use RequeueAfter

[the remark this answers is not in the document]
osmman: fixed

bouskaJ (approved):

PR #1784, opened 2026-05-07
fix: SECURESIGN-4401, add retry logic to the platform discovery to ha…

knrc:
…ndle openshift API Server blackouts

### DIFF
--- a/internal/utils/flag_or_env.go
+++ b/internal/utils/flag_or_env.go
@@ -4,9 +4,22 @@ import (
 	"flag"
 	"os"
 	"strconv"
+	"time"
 
 	"github.com/securesign/operator/internal/images"
 )
+
+// DurationFlagOrEnv defines a string flag which can be set by an environment variable.
+// Precedence: flag > env var > default value.
+func DurationFlagOrEnv(p *time.Duration, name string, envName string, defaultValue time.Duration, usage string) {
+	envValue := os.Getenv(envName)
+	if envValue != "" {
+		if value, err := time.ParseDuration(envValue); err == nil {
+			defaultValue = value
+		}
+	}
+	flag.DurationVar(p, name, defaultValue, usage)
+}
 
 // StringFlagOrEnv defines a string flag which can be set by an environment variable.
 // Precedence: flag > env var > default value.
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"flag"
 	"os"
+	"time"
 
 	appconfig "github.com/securesign/operator/internal/config"
 	"github.com/securesign/operator/internal/constants"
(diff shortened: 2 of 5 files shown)


knrc:
> 1. Timeout hides root cause `🐞 Bug` `◔ Observability`

Nit-picky, but we might as well be thorough

knrc:
> ```
> - Consider tightening `isTransientError` for `net.Error` by checking `Timeout()`/`Temporary()` where applicable, to avoid retrying clearly non-transient conditions.
> ```

This advice is incorrect in our scenario, as we are considering all network issues to be transient during this period.
- `Temporary()` is deprecated and should not be used
- `Timeout()` would be too restrictive

knrc:
> > 1. Timeout hides root cause `🐞 Bug` `◔ Observability`
> 
> Nit-picky, but we might as well be thorough

I modified the code to keep track of the last error, and it now gets reported as
```
E0507 23:42:50.404293       1 main.go:141] "Platform auto-detection failed, exiting" err="timed out waiting for API server during OpenShift platform detection: failed to get server groups: Get \"https://10.96.0.1:443/api\": dial tcp 10.96.0.1:443: connect: connection refused" logger="setup"
```

internal/utils/flag_or_env.go
```diff
@@ -4,10 +4,29 @@ import (
...
+
+// DurationFlagOrEnv defines a string flag which can be set by an environment variable.
+// Precedence: flag > env var > default value.
+func DurationFlagOrEnv(p *time.Duration, name string, envName string, defaultValue time.Duration, usage string) {
+	envValue := os.Getenv(envName)
+	if envValue != "" {
+		if value, err := time.ParseDuration(envValue); err == nil {
+			defaultValue = value
+		} else {
+			setupLog.Error(err, "Unable to parse duration from environment, falling back to default")
```
miyunari: Did you check if that log line appears?
Claude:
```
This is evaluated at package init time, but ctrl.SetLogger(klog.NewKlogr()) isn't called until cmd/main.go:136 — after DurationFlagOrEnv is invoked at line 113. If time.ParseDuration fails on the env var value, the error log (setupLog.Error(...)) goes through the default no-op logger and is silently swallowed. The user would get the default value with no indication that their env var was rejected.
```

miyunari (commented):
otherwise lgtm :)

internal/utils/flag_or_env.go
knrc: Good catch, I didn't test that part.  Will do that now

knrc: I took the easier way out and just handled it silently, with an additional log to show what is being used within the detection logic.  This is similar to the `BoolFlagOrEnv` behaviour, the one difference being that function will ignore the `defaultValue` on failure (always false) whereas the `DurationFlagOrEnv` function will honour it.

miyunari (approved):

PR #1789, opened 2026-05-11
feat(lint): custom action framework linter

osmman:
## Summary
- Adds custom golangci-lint v2 module plugin with two analyzers for the action framework:
  - **persiststatuserr**: flags `PersistStatus()` calls where the error return is discarded
  - **specmutation**: flags `instance.Spec` mutations inside action types (only Status is persisted)
- The specmutation analyzer detects **direct** and **indirect** Spec mutations:
  - Direct: `instance.Spec.Field = value` in action methods
  - Indirect via Spec pointer: `helper(&instance.Spec)` where helper writes through the pointer
  - Indirect via whole object: `helper(instance)` where helper writes `instance.Spec.Field`
  - Cross-package: uses `go/analysis` facts to track parameter mutations across package boundaries
- Linter is a separate Go module (`pkg/analyzer/`) to keep operator deps clean
- CI workflow updated to build and run the custom binary via `golangci-lint-action` with `install-mode: none` for inline PR annotations
- Fixes all existing spec mutation violations in ctlog, fulcio, and tsa actions
- Adds unit test coverage for TSA `alignStatusFields` (default signer, user-provided keys, partial refs, external chain) and KMS/Tink signer Handle paths

## Test plan
- [x] Analyzer unit tests pass — intra-package and cross-package (`go test ./pkg/analyzer/...`)
- [x] Controller tests pass for modified ctlog, fulcio, and tsa packages
- [x] Custom linter reports 0 new issues after fixes
- [x] CI passes on GitHub

### DIFF
--- a/.custom-gcl.yml
+++ b/.custom-gcl.yml
@@ -1,0 +1,5 @@
+version: v2.12.2
+plugins:
+  - module: 'github.com/securesign/operator/pkg/analyzer'
+    import: 'github.com/securesign/operator/pkg/analyzer'
+    path: ./pkg/analyzer
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -13,9 +13,14 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: 'go.mod'
+      - name: Build custom golangci-lint
+        run: |
+          make custom-golangci-lint
+          install ./bin/custom-gcl /usr/local/bin/golangci-lint
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v9
         with:
+          install-mode: none
           version: v2.12.2
           args: --verbose --timeout=15m
           only-new-issues: true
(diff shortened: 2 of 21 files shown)


bouskaJ (approved):
Not introduced by this PR, but worth noting. generate_signer.go:376:

  if instance.Spec.Signer.File == nil && instance.Spec.Signer.CertificateChain.CertificateChainRef ==
  nil {
      instance.Status.Signer.File = new(v1alpha1.File)
  }

  A KMS or Tink signer with no CertificateChainRef also satisfies this — File is nil,
  CertificateChainRef is nil. Status ends up with both Kms and File populated, with
  PrivateKeyRef/PasswordRef pointing at nonexistent keys.

  Unlikely in practice (KMS/Tink normally provide an external chain), but the guard could be:

  if instance.Spec.Signer.File == nil && instance.Spec.Signer.Kms == nil && instance.Spec.Signer.Tink ==
  nil && instance.Spec.Signer.CertificateChain.CertificateChainRef == nil {

PR #1797, opened 2026-05-12
DO NOT MERGE - TEST

SequeI:


bouskaJ:
@SequeI Try to invoke the TUF migration job. I do expect a failure here.

PR #1896, opened 2026-05-29
feat: add Coverport instrumentation and parallel CI jobs for E2E coverage

kdacosta0:
## Summary

- Instruments the operator binary with [Coverport](https://github.com/konflux-ci/coverport) to extract runtime E2E coverage from a live Kind cluster via HTTP port-forwarding
- Adds two new parallel CI jobs (`build-operator-coverage`, `test-kind-coverage`) that run alongside the existing `test-kind` job without affecting it
- Adds `unit` and `e2e` flag definitions to `codecov.yml` with carryforward semantics

## Two-Build Approach

| Build | Dockerfile | CGO | Purpose | Affected by this PR? |
|---|---|---|---|---|
| **Production** | `Dockerfile.rhtas-operator.rh` | CGO=1 (FIPS) | Konflux/Tekton release pipeline | No |
| **Community** | `Dockerfile` | CGO=0 | GitHub Actions dev CI | No |
| **Coverage** | `Dockerfile.coverage` (new) | CGO=0 + `-cover` | E2E coverage collection | Yes (new) |

Production images are safe by design: both production Dockerfiles use `COPY cmd/main.go cmd/main.go` (not `COPY cmd/`), so `cmd/coverage.go` is physically excluded. The `//go:build coverage` tag provides a second layer of protection.

## New CI Jobs

- **`build-operator-coverage`**: Builds the instrumented image using `Dockerfile.coverage` and pushes to GHCR with a `-coverage` tag suffix. Runs in parallel with `build-operator`.
- **`test-kind-coverage`**: Deploys the instrumented image to a Kind cluster, runs the existing E2E suite (`make test-e2e`), then uses `coverport collect` to scrape coverage from the live operator pod and uploads to Codecov under the `e2e` flag. Mirrors the existing `test-kind` job step-for-step.

Both jobs are gated to `main` branch only (pushes + PRs targeting main).

## Test plan

- [ ] `build-operator-coverage` job succeeds and pushes the coverage image to GHCR
- [ ] `test-kind-coverage` job deploys the instrumented operator, E2E tests pass, and Coverport collects coverage
- [ ] Codecov receives E2E coverage upload under the `e2e` flag
- [ ] Existing `build-operator`, `test-kind`, and all other jobs are completely unaffected
- [ ] Production Dockerfiles (`Dockerfile`, `Dockerfile.rhtas-operator.rh`) are not modified

Ref: [COVERPORT-197](https://redhat.atlassian.net/browse/COVERPORT-197)

[COVERPORT-197]: https://redhat.atlassian.net/browse/COVERPORT-197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -645,3 +645,159 @@ jobs:
         run: |
           kubectl logs -n openshift-rhtas-operator deployment/rhtas-operator-controller-manager
         if: failure()
+
+  build-operator-coverage:
+    name: Build-operator-coverage
+    runs-on: ubuntu-24.04
+    if: >-
+      (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
+      (github.event_name == 'pull_request' && github.base_ref == 'main')
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
+      - name: Log in to GitHub Container Registry
+        uses: redhat-actions/podman-login@v1
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          auth_file_path: /tmp/config.json
+
+      - name: Replace images
+        run: make dev-images && cat config/default/images.env
+
+      - name: Build coverage-instrumented operator container
+        run: make docker-build-coverage
+
+      - name: Push coverage image
+        run: podman push ${{ env.IMG }}-coverage
+
+  test-kind-coverage:
+    name: Test E2E with coverage
+    runs-on: ubuntu-24.04
+    if: >-
+      (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
+      (github.event_name == 'pull_request' && github.base_ref == 'main')
+    permissions:
+      contents: read
+      packages: read
+      id-token: write
+    needs:
+      - build-operator-coverage
+      - build-tuftool
+    steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -645,3 +645,159 @@ jobs:
...
+      - name: Replace images
+        run: make dev-images && cat config/default/images.env
+
+      - name: Build coverage-instrumented operator container
+        run: make docker-build-coverage
+
+      - name: Push coverage image
+        run: podman push ${{ env.IMG }}-coverage
+
+  test-kind-coverage:
```
osmman, line 679: Is there some reason why you did not modify existin `test-*` jobs and need to create new one?

kdacosta0: Yes, it's because the coverage binary is a different build artifact - different Dockerfile compiled with `-cover -covermode=atomic -tags=coverage` and includes the Coverport instrumentation import. I wanted to keep it separate so we still run E2E against the clean community build too, since this new build could introduce or mask subtle differences

kdacosta0: We could merge it, but:

`+ the coverage would grow `
`+ less jobs`

`- we would lose the clean build validity`

osmman: We can keep it currently like that. We will have to really fix e2e flaky tests.

osmman (approved):

PR #1917, opened 2026-06-10
feat: vanilla-Kubernetes support, monitoring fail-fast, startup detection, docs

anithapriyanatarajan:
This PR improves support for running the operator on vanilla Kubernetes by
making monitoring integration robust and clearly documented.

## Kubernetes platform support and documentation

- Added detailed guidance in `docs/kubernetes.md` and `README.md` for running on vanilla Kubernetes, covering prerequisites (cert-manager, server-side apply, image pull access), differences from OpenShift, and
  external access / hostname configuration.
- Updated the kubernetes overlay namespace from `sigstore-operator` to `rhtas-operator` to be consistent with the `openshift-rhtas-operator` naming convention.

## Monitoring integration

- Added `--monitoring` / `MONITORING` flag with auto-detection of the Prometheus Operator API (`monitoring.coreos.com/v1`) at startup using the discovery client. Handles both `apierrors.IsNotFound` and
  `*discovery.ErrGroupDiscoveryFailed` for reliable absence detection across K8s versions.
- **Fail-fast on missing CRD**: if `monitoring.enabled: true` but the ServiceMonitor CRD is not installed, `Handle()` now returns a clear error directing the user to install Prometheus Operator or set `monitoring.enabled: false`. This matches the project position established in PR #1440 — the CR spec must never be silently ignored.
- Moves the `kubernetes.IsMonitoringAvailable()` check from `CanHandle()` into `Handle()` for all 8 monitoring actions (ctlog, ctlog/monitor, fulcio, rekor/server, rekor/monitor, trillian/logserver, trillian/logsigner, tsa).
- Updated `monitoring_test.go` to assert the correct behavior: `CanHandle` returns `true` when `monitoring.enabled=true` regardless of CRD presence, so `Handle` always runs and produces a meaningful error.

Co-authored-by: [Claude Code](https://claude.com/claude-code)


knrc:
This was also identified as part of the EKS work, see https://github.com/securesign/secure-sign-operator/pull/1440

internal/controller/ctlog/actions/monitoring.go
```diff
@@ -29,7 +29,7 @@ func (i monitoringAction) Name() string {
 }
 
 func (i monitoringAction) CanHandle(_ context.Context, instance *rhtasv1alpha1.CTlog) bool {
-	return instance.Spec.Monitoring.Enabled && state.FromInstance(instance, constants.ReadyCondition) >= state.Creating
+	return instance.Spec.Monitoring.Enabled && kubernetes.IsMonitoringAvailable() && state.FromInstance(instance, constants.ReadyCondition) >= state.Creating
```
osmman: Adding `kubernetes.IsMonitoringAvailable()` to `CanHandle()` means the action is never entered when the monitoring CRD is absent — `Handle()` never runs, so no error, no status condition, no event is produced. The user's `monitoring.enabled: true` is silently ignored.

This same approach was rejected in PR #1440 (CHANGES_REQUESTED). The established project position is: if `monitoring.enabled` is `true` but the CRD is missing, reconciliation should FAIL with a clear status condition. See also SECURESIGN-4584 for the planned v1 API solution.

The check should be moved into `Handle()` and return an error:
```go
if !kubernetes.IsMonitoringAvailable() {
    return i.Error(ctx, fmt.Errorf("monitoring.enabled is true but ServiceMonitor CRD is not installed; install Prometheus Operator or set monitoring.enabled=false"), instance)
}
```

Same issue applies to all 8 monitoring `CanHandle()` methods: ctlog, ctlog/monitor, fulcio, rekor/server, rekor/monitor, trillian/logserver, trillian/logsigner, tsa.

osmman (changes requested):
Thank you for the effort on vanilla-Kubernetes support, @anithapriyanatarajan. The documentation in `docs/kubernetes.md` and `README.md` is genuinely valuable and should be preserved. The `DetectMonitoringAvailable()` detection logic is also solid on its own.

However, there are blocking issues:

**1. Shadow disabling of monitoring (BLOCKING)**

Adding `kubernetes.IsMonitoringAvailable()` to every `CanHandle()` silently skips monitoring when the ServiceMonitor CRD is absent, even when `monitoring.enabled: true`. The CR says monitoring is on, status shows Ready, but no ServiceMonitor was created.

This exact approach was rejected in PR #1440 (@osmman, CHANGES_REQUESTED):

> "Status shows `Ready: true`, Spec declares `monitoring.enabled: true`, But monitoring is NOT actually enabled. Expected behavior: reconciliation should **fail** with an error."

Correct behavior: fail reconciliation with a clear condition (e.g., `reason: ServiceMonitorCRDNotAvailable`), emit an event, direct the user to install Prometheus Operator or set `monitoring.enabled: false`.

There is also planned work (SECURESIGN-4584) to properly split ServiceMonitor creation from metrics exposure in the v1 API.

**2. Duplicate image source of truth (BLOCKING)**

`config/overlays/kubernetes/images.env` duplicates all 17 image references from `config/default/images.env`. Every image bump now requires updating two files. The existing `RELATED_IMAGE_*` env vars and `--<component>-image` flags already solve image overriding without duplication.

For clusters without `registry.redhat.io` credentials, there are existing options:
- Configure registry mirroring from Red Hat registry to quay.io at the cluster level
- For local development, use `make dev-images` as documented in `DEVELOPMENT.md`

**3. v1alpha1 sample changes (BLOCKING)**

Do not modify `config/samples/rhtas_v1alpha1_securesign.yaml`. The project has begun v1 API work; sample changes should target v1.

**4. Commit structure**

Please split into logical commits (docs, detection logic, controller changes, kustomize overlay).

config/overlays/kubernetes/images.env
```diff
@@ -0,0 +1,74 @@
+# Public-registry image overrides for vanilla Kubernetes.
```
osmman: This file should not exist — it creates a second source of truth for container images alongside `config/default/images.env`. Every image bump now requires updating two files.

The existing `RELATED_IMAGE_*` env var / CLI flag mechanism already provides per-image overriding without duplication. For clusters without `registry.redhat.io` credentials:
- Configure registry mirroring from Red Hat registry to quay.io at the cluster level
- For local development, use `make dev-images` as documented in `DEVELOPMENT.md`

Also note that `RELATED_IMAGE_TRILLIAN_NETCAT` still points to `registry.redhat.io` here, which breaks the file's own stated purpose.

config/samples/rhtas_v1alpha1_securesign.yaml
```diff
@@ -11,12 +11,12 @@ spec:
     externalAccess:
       enabled: true
     monitoring:
-      enabled: true
+      enabled: false
```
osmman: v1alpha1 is frozen — the project has begun v1 API work. Do not modify v1alpha1 samples. If a vanilla-Kubernetes-specific sample is needed, create a separate file or target the v1 samples.

internal/utils/kubernetes/platform.go
```diff
@@ -100,6 +105,41 @@ func detectOpenShiftWithRetry(ctx context.Context, log logr.Logger, cl client.Cl
 	return found, nil
 }
 
+// DetectMonitoringAvailable detects whether the Prometheus Operator's
+// monitoring.coreos.com/v1 API (and the ServiceMonitor kind in particular) is
```
osmman, line 109: `ServerResourcesForGroupVersion` does not reliably return errors that pass `apierrors.IsNotFound()` when a group/version is not registered. Depending on K8s version you may get a generic error string or `*discovery.ErrGroupDiscoveryFailed`, causing the "auto-detection failed" error log in `cmd/main.go` instead of clean absence detection.

cmd/main.go
```diff
@@ -113,6 +113,7 @@ func main() {
 		"Directory containing tls.crt and tls.key for metrics server TLS. If empty, a self-signed certificate is generated.")
 	flag.Int64Var(&appconfig.CreateTreeDeadline, "create-tree-deadline", appconfig.CreateTreeDeadline, "The time allowance (in seconds) for the create tree job to run before failing.")
 	utils.BoolFlagOrEnv(&appconfig.Openshift, "openshift", "OPENSHIFT", false, "Enable to ensures the operator applies OpenShift specific configurations.")
+	utils.BoolFlagOrEnv(&appconfig.MonitoringAvailable, "monitoring", "MONITORING", false, "Enable ServiceMonitor (monitoring.coreos.com) creation. When not set explicitly, the operator auto-detects the Prometheus Operator API.")
```
osmman, line 118: Detection runs once at startup. If Prometheus Operator is installed after the operator starts, it won't be detected until restart. With the skip-based approach this is especially problematic — the operator silently won't create ServiceMonitors even after Prometheus Operator is available. If the approach changes to fail-on-missing, this resolves naturally via reconcile retries.

internal/controller/ctlog/actions/monitoring_test.go
```diff
@@ -0,0 +1,85 @@
...
+			monitoringAvailable: true,
+			phase:               state.Creating,
+			want:                true,
+		},
+		{
+			name:                "enabled but ServiceMonitor API unavailable",
+			monitoringEnabled:   true,
+			monitoringAvailable: false,
+			phase:               state.Creating,
+			want:                false,
```
osmman: This test validates the wrong behavior. When `monitoring.enabled: true` but ServiceMonitor API is unavailable, `CanHandle()` should return `true` so `Handle()` can run and produce a proper error — not `false`.

config/overlays/kubernetes/kustomization.yaml
```diff
@@ -1,6 +1,13 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
+# Override the base namespace (config/default sets "openshift-rhtas-operator")
+# with a platform-neutral name for vanilla Kubernetes installs. This retargets
+# every namespaced resource, including the cert-manager Certificate whose
+# namespace is propagated into the webhook configs/CRDs via the replacements
+# below, so they stay consistent.
+namespace: sigstore-operator
```
osmman: This operator is not part of the Sigstore organization — if renaming from `openshift-rhtas-operator`, the namespace should be `rhtas-operator`. This is also a breaking change for existing users of the kubernetes overlay.

osmman:
Hello @anithapriyanatarajan, you can find design of new changes for ServiceMonitor in [SECURESIGN-4584](https://redhat.atlassian.net/browse/SECURESIGN-4584?focusedCommentId=17363390) issue.

[SECURESIGN-4584]: https://redhat.atlassian.net/browse/SECURESIGN-4584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

anithapriyanatarajan:
@osmman Thank you for the detailed review. Attempted to address most of them. Tested with a local kind cluster to make sure the changes are working fine.

config/overlays/kubernetes/kustomization.yaml
```diff
@@ -104,3 +111,5 @@ replacements:
       delimiter: '.'
       index: 1
       create: true
+
+
```
fghanmi: extra newline

anithapriyanatarajan:
closing in favor of https://github.com/securesign/secure-sign-operator/pull/1990. Github Actions CI does not work on PR from forks

PR #1928, opened 2026-06-16
feat: Reduce usage of deprecated EncryptPEMBlock and DecryptPEMBlock

JasonPowr:
# Summary
- Remove x509.EncryptPEMBlock from all production code — auto-generated private keys (Fulcio, CTLog, TSA) are now stored as unencrypted PEM
- Retain x509.DecryptPEMBlock/x509.IsEncryptedPEMBlock in three backward-compatibility code paths so existing encrypted keys continue to work without migration
- Users can still provide both encrypted and unencrypted keys — fully backward compatible
- Add E2E test coverage for user-provided unencrypted keys alongside existing encrypted key tests

https://redhat.atlassian.net/browse/SECURESIGN-3617

### DIFF
--- a/internal/controller/fulcio/utils/fulcio_secrets.go
+++ b/internal/controller/fulcio/utils/fulcio_secrets.go
@@ -43,19 +43,17 @@ func (c FulcioCertConfig) ToData() map[string][]byte {
 	return result
 }
 
-func CreateCAKey(key *ecdsa.PrivateKey, password []byte) ([]byte, error) {
+func CreateCAKey(key *ecdsa.PrivateKey) ([]byte, error) {
 	mKey, err := x509.MarshalECPrivateKey(key)
 	if err != nil {
 		return nil, err
 	}
 
-	block, err := x509.EncryptPEMBlock(rand.Reader, "EC PRIVATE KEY", mKey, password, x509.PEMCipherAES256) //nolint:staticcheck
-	if err != nil {
-		return nil, err
-	}
-
 	var pemData bytes.Buffer
-	if err := pem.Encode(&pemData, block); err != nil {
+	if err := pem.Encode(&pemData, &pem.Block{
+		Type:  "EC PRIVATE KEY",
+		Bytes: mKey,
+	}); err != nil {
 		return nil, err
 	}
 
@@ -89,6 +87,7 @@ func CreateFulcioCA(config *FulcioCertConfig) ([]byte, error) {
 
 	block, _ := pem.Decode(config.PrivateKey)
 	keyBytes := block.Bytes
+	// Deprecated: kept for backward compatibility with existing encrypted keys from pre-FIPS operator versions.
 	if x509.IsEncryptedPEMBlock(block) { //nolint:staticcheck
 		keyBytes, err = x509.DecryptPEMBlock(block, config.PrivateKeyPassword) //nolint:staticcheck
 		if err != nil {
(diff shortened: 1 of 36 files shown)


osmman:
@JasonPowr should we mark `*PasswordRef` fields in api as deprecated?

JasonPowr:
> @JasonPowr should we mark `*PasswordRef` fields in api as deprecated?

I don't think we should, the operator just no longer generates encrypted PEM internally it still supports reading user-provided encrypted keys, so the PasswordRef will still be needed. If we do deprecate it we would also have to require all customers to rotate their keys to use unencrypted materials and that could be messy

osmman (commented):
## Code Review — FIPS: Remove deprecated EncryptPEMBlock / DecryptPEMBlock

**Looks good.** Clean separation between "stop generating encrypted keys" and "keep reading existing ones." Consistent pattern across Fulcio, CTLog, and TSA. New E2E test for the unencrypted path is solid.

### Deprecating `*PasswordRef` fields

For marking them deprecated in the CRD — add kubebuilder doc comments on the struct fields. Controller-gen picks these up into the OpenAPI description. Keep the fields functional but document them as legacy. Example following Go's `x509.EncryptPEMBlock` deprecation style:

```go
// Deprecated: Legacy PEM encryption as specified in RFC 1423 is insecure by design
// and not FIPS-compliant. Auto-generated keys are no longer password-encrypted;
// this field is retained only for backward compatibility with existing user-provided
// encrypted keys. Kubernetes Secrets provide encryption-at-rest.
// +optional
PrivateKeyPasswordRef *SecretKeySelector `json:"privateKeyPasswordRef,omitempty"`
```

internal/controller/fulcio/utils/fulcio_secrets.go
```diff
@@ -89,6 +87,7 @@ func CreateFulcioCA(config *FulcioCertConfig) ([]byte, error) {
 
 	block, _ := pem.Decode(config.PrivateKey)
 	keyBytes := block.Bytes
```
osmman, line 92: Pre-existing: `pem.Decode` can return nil `block` → panic on `block.Bytes`. TSA's `parsePrivateKey` handles this correctly. Worth adding a nil check since you're touching this path.

osmman:
> > @JasonPowr should we mark `*PasswordRef` fields in api as deprecated?
> 
> I don't think we should, the operator just no longer generates encrypted PEM internally it still supports reading user-provided encrypted keys, so the PasswordRef will still be needed. If we do deprecate it we would also have to require all customers to rotate their keys to use unencrypted materials and that could be messy

I though to keep the field forever but mark them to user that it is preferred to not use them.

internal/controller/tsa/actions/generate_signer.go
```diff
@@ -393,15 +389,6 @@ func (g generateSigner) alignStatusFields(secretName string, instance *rhtasv1.T
 				},
 			}
 		}
```
osmman, line 391: After upgrade, `alignStatusFields` no longer populates `Status.Signer.File.PasswordRef`. Is this upgrade path tested?

JasonPowr:
> > > @JasonPowr should we mark `*PasswordRef` fields in api as deprecated?
> > 
> > 
> > I don't think we should, the operator just no longer generates encrypted PEM internally it still supports reading user-provided encrypted keys, so the PasswordRef will still be needed. If we do deprecate it we would also have to require all customers to rotate their keys to use unencrypted materials and that could be messy
> 
> I though to keep the field forever but mark them to user that it is preferred to not use them.

Ok yea that makes a lot of sense :), I saw deprecated and thought removal, will add :)

internal/controller/tsa/actions/generate_signer.go
JasonPowr: The upgrade e2e seems to pass fine, so that shows it comes up fine after, I also added a unit test to make sure its dropped correctly

osmman (approved):

PR #1930, opened 2026-06-17
fix: remove TerminalError from transient API server errors

osmman:
## Summary

- Remove `reconcile.TerminalError` wrapper from 10 call sites where transient Kubernetes API errors were incorrectly treated as permanent failures, preventing controller-runtime from retrying
- Audit all 24 `TerminalError` call sites — keep 14 that are truly unrecoverable (invalid config, exhausted job retries, data corruption)
- Add unit tests verifying that API server errors are retryable (not terminal) using client interceptors

Fixes: [SECURESIGN-4664](https://issues.redhat.com/browse/SECURESIGN-4664)

## Root Cause

On fresh OLM installs, the API server's CRD REST mapping may not be fully propagated when the operator starts reconciling (~9s after CRD creation). When `ensure.ControllerReference` sets `blockOwnerDeletion: true` on an ownerReference, the API server validates the owner's GVK via REST mapping — and rejects the request if the CRD hasn't propagated yet.

These errors were wrapped in `reconcile.TerminalError`, which tells controller-runtime **not to requeue**. The operator gave up permanently after the first failure.

### Production code (10 TerminalError removals)

**`internal/action/rbac/action.go`** — SA, Role, RoleBinding creation (3 sites)
**`internal/action/tree/action.go`** — SA, Role, RoleBinding creation + ConfigMap Get (6 sites)

Plain errors are returned instead, letting controller-runtime retry with exponential backoff.

### TerminalError sites kept (14 — truly unrecoverable)

| Category | Examples |
|----------|----------|
| Invalid user config | missing private key, invalid cron schedule, Trillian port not specified, PVC size not set, root key secret ref not specified |
| Exhausted job retries | TUF init/migration job failed, createtree job failed |
| Data integrity | multiple jobs present, unparseable tree ID, config marshal failure |

### Tests added/updated

- `TestRbac_CreationFailure_ReturnsRetryableError` — table-driven, covers SA/Role/RoleBinding via Create interceptor
- `TestResolveTree_ConfigMapGetFailure_ReturnsRetryableError` — covers handleJob/handleJobFinished/handleExtractJobResult via Get interceptor
- `TestResolveTree_RbacCreationFailure_ReturnsRetryableError` — covers SA creation via Create interceptor

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/action/rbac/...` — all pass including new error-path tests
- [x] `go test ./internal/action/tree/...` — all pass including new error-path tests
- [x] `go test ./internal/controller/tuf/...` — all pass including updated migration test
- [x] `go test ./internal/...` — full internal test suite passes (48 packages, 0 failures)
- [ ] E2E: fresh install on a cluster confirms the operator recovers from transient CRD propagation delays

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

[SECURESIGN-4664]: https://redhat.atlassian.net/browse/SECURESIGN-4664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/action/rbac/action.go
+++ b/internal/action/rbac/action.go
@@ -17,7 +17,6 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 func WithRule[T apis.ConditionsAwareObject](rule rbacv1.PolicyRule) func(action2 *rbacAction[T]) {
@@ -100,7 +99,7 @@ func (i rbacAction[T]) handleServiceAccount(ctx context.Context, instance T) *ac
 		ensure.ControllerReference[*v1.ServiceAccount](instance, i.Client),
 		ensure.Labels[*v1.ServiceAccount](slices.Collect(maps.Keys(l)), l),
 	); err != nil {
-		return i.Error(ctx, reconcile.TerminalError(fmt.Errorf("could not create SA: %w", err)), instance)
+		return i.Error(ctx, fmt.Errorf("could not create SA: %w", err), instance)
 	}
 
 	return i.Continue()
@@ -132,7 +131,7 @@ func (i rbacAction[T]) handleRole(ctx context.Context, instance T) *action.Resul
 		ensure.Labels[*rbacv1.Role](slices.Collect(maps.Keys(l)), l),
 		kubernetes.EnsureRoleRules(i.rules...),
 	); err != nil {
-		return i.Error(ctx, reconcile.TerminalError(fmt.Errorf("could not create Role: %w", err)), instance)
+		return i.Error(ctx, fmt.Errorf("could not create Role: %w", err), instance)
 	}
 
 	return i.Continue()
@@ -171,7 +170,7 @@ func (i rbacAction[T]) handleRoleBinding(ctx context.Context, instance T) *actio
 			rbacv1.Subject{Kind: "ServiceAccount", Name: i.rbacName, Namespace: instance.GetNamespace()}, //nolint:goconst
 		),
 	); err != nil {
-		return i.Error(ctx, reconcile.TerminalError(fmt.Errorf("could not create RoleBinding: %w", err)), instance)
+		return i.Error(ctx, fmt.Errorf("could not create RoleBinding: %w", err), instance)
 	}
 
 	return i.Continue()
--- a/internal/action/rbac/action_test.go
+++ b/internal/action/rbac/action_test.go
@@ -2,6 +2,8 @@ package rbac
 
 import (
 	"context"
+	stderrors "errors"
+	"fmt"
 	"reflect"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


bouskaJ (commented):
The other 9 removals are correct — those are transient API server
  errors that should retry.
  
  However, the migration_job.go change (root key secret NotFound) is
  different: this is a user config problem, not a transient API
  error. If the secret referenced by RootKeySecretRef doesn't exist,
  retrying with backoff won't make it appear — the user needs to
  fix their CR. The TerminalError here intentionally sets the TUF
  resource to Failure for users who don't have their keys in the
  cluster — the doc procedure starts from this state.

  Suggest reverting this one site and the corresponding test change,
  keeping the other 9 removals as-is.

osmman:
> The other 9 removals are correct — those are transient API server errors that should retry.
> 
> However, the migration_job.go change (root key secret NotFound) is different: this is a user config problem, not a transient API error. If the secret referenced by RootKeySecretRef doesn't exist, retrying with backoff won't make it appear — the user needs to fix their CR. The TerminalError here intentionally sets the TUF resource to Failure for users who don't have their keys in the cluster — the doc procedure starts from this state.
> 
> Suggest reverting this one site and the corresponding test change, keeping the other 9 removals as-is.

There are three scenarios when this secret NotFound happens:

1. **Misconfiguration in the CR** — user referenced a wrong secret name, they need to fix the CR
2. **Correct CR config, but user hasn't created the secret yet** — they need to create it, operator should pick it up automatically once they do
3. **Everything is correct, but the API server hasn't propagated it yet** — cache delays, etcd propagation, informer sync gaps — resolves on its own

We can't distinguish between these at the API level — all three return the same NotFound error. With TerminalError, scenario 3 permanently bricks the operator. In scenario 2, once the user creates the secret, the operator won't pick it up — they have to also touch the CR to restart reconciliation, which isn't obvious. Every other controller returns a plain retryable error for the same pattern ([Fulcio generate_cert.go:110](https://github.com/securesign/secure-sign-operator/blob/3d40daf1/internal/controller/fulcio/actions/generate_cert.go#L110), [CTlog server_config.go:212](https://github.com/securesign/secure-sign-operator/blob/3d40daf1/internal/controller/ctlog/actions/server_config.go#L212), [CTlog server_config.go:285](https://github.com/securesign/secure-sign-operator/blob/3d40daf1/internal/controller/ctlog/actions/server_config.go#L285)) — [`migration_job.go:55`](https://github.com/securesign/secure-sign-operator/blob/3d40daf1/internal/controller/tuf/actions/migration_job.go#L55) is the only place wrapping secret NotFound in TerminalError. Retryable error with backoff covers all three cases — transient issues self-heal, config errors still surface in status while recovering automatically once fixed.

Is there a specific case where the terminal behavior is needed here? If there's a scenario I'm missing, it would help us make the right call — otherwise aligning with the rest of the codebase seems safer.

bouskaJ:
All the secrets you are referring to are generated by the action itself, so it makes sense to keep trying. However, this one is different.
Please keep in mind that the migration job runs only during upgrades and is not triggered on fresh deployments. This eliminates most scenarios that fall under Case 3.

 Regarding Case 1, we issue a terminal error for user misconfigurations, so the 'Failure' status is correct in this context. 

Case 2 applies to users who **intentionally** do not store their keys on the cluster (e.g., storing them on a USB drive instead). These users will need to follow the manual TUF upgrade process documented at: https://docs.redhat.com/en/documentation/red_hat_trusted_artifact_signer/1/html/administration_guide/rhtas-ocp-admin#manually-updating-the-update-framework-repository-after-upgrading_admin"

osmman:
> Case 2 applies to users who intentionally do not store their keys on the cluster (e.g., storing them on a USB drive instead). These users will need to follow the manual TUF upgrade process documented at:

Thank you for explanation, commit d07ac913 removed from PR.

bouskaJ (approved):

PR #1927, opened 2026-06-16
feat: add imagePullSecrets support for all components

osmman:
## Summary
- Add `ServiceAccountConfig` with `ImagePullSecrets` field to all component specs (CTlog, Fulcio, Rekor, Trillian, TSA, TUF)
- Add `WithImagePullSecrets` option to RBAC action to propagate secrets to ServiceAccounts
- Add conversion field restoration for `ImagePullSecrets` in all v1alpha1 spoke handlers

## Dependencies
- Depends on #1934 (conversion infrastructure)

## Test plan
- [x] `go build ./...` passes
- [x] Conversion unit tests pass
- [x] Fuzz roundtrip tests pass
- [ ] E2E: verify imagePullSecrets propagated to ServiceAccounts

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1934, opened 2026-06-17
refactor: vendor cluster-api conversion utilities and adopt conversion-gen

osmman:
## Summary
- Vendor `MarshalData`, `UnmarshalData`, and `FuzzTestFunc` from `sigs.k8s.io/cluster-api/util/conversion` into `internal/conversion/`, removing the direct cluster-api module dependency
- Use project-scoped annotation key `rhtas.redhat.com/conversion-data`
- Replace JSON-based `marshalConvert` with `conversion-gen` generated field-by-field conversion functions
- Add `conversion-gen` Makefile target (`make generate-conversion`)
- Move conversion tests from `api/v1` to `api/v1alpha1` (spoke owns conversion)

## Test plan
- [x] `internal/conversion` unit tests pass (MarshalData, UnmarshalData, MarshalDataUnsafeNoCopy)
- [x] All 7 conversion unit tests pass
- [x] All 7 fuzz roundtrip tests pass (10,000 iterations × 2 directions each)
- [x] `go build ./...` passes
- [x] Linter passes (no new issues)


bouskaJ (commented):
One concern worth discussing: sigs.k8s.io/cluster-api is a monolithic module (~160 dependencies including etcd, CEL, viper, prometheus, gRPC, etc.) and we're importing it for two symbols — MarshalData (~15
  lines) and the DataAnnotation constant. The go.sum already shows ~20 bumped indirect deps as a side effect. Beyond the dependency weight, this increases our exposure to CVEs in transitive dependencies we don't
  actually use — any vulnerability in etcd, gRPC, or the other indirect imports would show up in scans and require attention even though none of that code is relevant to us.

  Since this operator isn't a cluster-api provider, there's no interop reason to depend on the upstream module or match their annotation key. We could vendor just the functions we need (MarshalData,
  UnmarshalData, and the constant — roughly 40 lines) into an internal package and avoid the transitive dependency surface entirely.

osmman:
> One concern worth discussing: sigs.k8s.io/cluster-api is a monolithic module (~160 dependencies including etcd, CEL, viper, prometheus, gRPC, etc.) and we're importing it for two symbols — MarshalData (~15 lines) and the DataAnnotation constant. The go.sum already shows ~20 bumped indirect deps as a side effect. Beyond the dependency weight, this increases our exposure to CVEs in transitive dependencies we don't actually use — any vulnerability in etcd, gRPC, or the other indirect imports would show up in scans and require attention even though none of that code is relevant to us.
> 
> Since this operator isn't a cluster-api provider, there's no interop reason to depend on the upstream module or match their annotation key. We could vendor just the functions we need (MarshalData, UnmarshalData, and the constant — roughly 40 lines) into an internal package and avoid the transitive dependency surface entirely.

Cloned code to project and I have to move conversion-gen. There has been issues with original json based migration which caused loses nil-vs-empty distinction. It happen on `[]v1.Toleration{}`, `*metav1.Time{}` and `map[string]string{}`. Fuzzer from cluster-api discover these.

bouskaJ:
> > One concern worth discussing: sigs.k8s.io/cluster-api is a monolithic module (~160 dependencies including etcd, CEL, viper, prometheus, gRPC, etc.) and we're importing it for two symbols — MarshalData (~15 lines) and the DataAnnotation constant. The go.sum already shows ~20 bumped indirect deps as a side effect. Beyond the dependency weight, this increases our exposure to CVEs in transitive dependencies we don't actually use — any vulnerability in etcd, gRPC, or the other indirect imports would show up in scans and require attention even though none of that code is relevant to us.
> > Since this operator isn't a cluster-api provider, there's no interop reason to depend on the upstream module or match their annotation key. We could vendor just the functions we need (MarshalData, UnmarshalData, and the constant — roughly 40 lines) into an internal package and avoid the transitive dependency surface entirely.
> 
> Cloned code to project and I have to move conversion-gen. There has been issues with original json based migration which caused loses nil-vs-empty distinction. It happen on `[]v1.Toleration{}`, `*metav1.Time{}` and `map[string]string{}`. Fuzzer from cluster-api discover these.

yes, I think that conversion-gen is the right choice. Vast majority of mature operators use this tool.

internal/conversion/conversion.go
```diff
@@ -0,0 +1,276 @@
...
+	if err := json.Unmarshal([]byte(data), to); err != nil {
+		return false, err
+	}
+	delete(annotations, DataAnnotation)
+	from.SetAnnotations(annotations)
+	return true, nil
+}
+
+// GetFuzzer returns a new fuzzer to be used for testing.
+func GetFuzzer(scheme *runtime.Scheme, funcs ...fuzzer.FuzzerFuncs) *randfill.Filler {
```
bouskaJ, line 128: This is a test utility. Move it to test the `internal/testing` package (apply to all Fuzzers in this file)

internal/conversion/conversion.go
```diff
@@ -0,0 +1,276 @@
...
+
+const (
+	// DataAnnotation is the annotation that conversion webhooks
+	// use to retain the data in case of down-conversion from the hub.
+	DataAnnotation = "rhtas.redhat.com/conversion-data"
+)
+
+// MarshalData stores the source object as json data in the destination object annotations map.
+// It ignores the metadata of the source object.
+func MarshalData(src metav1.Object, dst metav1.Object) error {
```
bouskaJ, line 51: MarshalData / UnmarshalData
  
  With conversion-gen handling the actual field-by-field conversion, these functions serve a different
  role than they do in cluster-api. In cluster-api, marshaling the full hub object is the conversion
  mechanism — the name fits. Here, they're just storing and restoring a supplementary annotation for
  hub-only fields that can't survive the spoke round-trip. MarshalData reads like it marshals and
  returns data, but it actually mutates dst by setting an annotation. Since the project already
  decoupled from cluster-api, names like StoreHubState/RestoreHubState would better describe what they
  actually do.

bouskaJ (changes requested):

internal/conversion/conversion.go
osmman: This file is a vendored copy of `sigs.k8s.io/cluster-api/util/conversion` (see the copyright header and package doc comment). The intent is to keep it as close to upstream as possible so that future upgrades/syncs are straightforward.

Moving the fuzzer helpers to a separate package would diverge the structure from upstream and make future syncs harder to maintain. Every difference is something someone has to remember and resolve during each sync, and in this case that someone is me. Less divergence = easier maintenance.

internal/conversion/conversion.go
osmman: Keeping the names as-is for the same reason as above — this is a vendored copy of cluster-api's `util/conversion` and we need to periodically sync changes from upstream. Every rename is a difference someone has to remember and resolve during each sync, and in this case that someone is me. Less divergence = easier maintenance.

api/v1alpha1/securesign_conversion.go
```diff
@@ -1,14 +1,24 @@
...
 )
 
 func (src *Securesign) ConvertTo(dstRaw conversion.Hub) error {
-	return marshalConvert(src, dstRaw.(runtime.Object))
+	dst := dstRaw.(*rhtasv1.Securesign)
+	if err := marshalConvert(src, dst); err != nil {
+		return err
+	}
+	removeDataAnnotation(dst)
+	return nil
```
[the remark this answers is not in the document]
osmman: This is side effect of splitting PR from https://github.com/securesign/secure-sign-operator/pull/1927. Thre is not usage becasue there are no migration compare to planed changes see #1927 PR

bouskaJ:
FIPS error:
```
All timeout after 300s waiting for Fulcio. Root cause — Fulcio pod fatals immediately:
  fileca: encrypted private keys are not supported in FIPS mode, use an unencrypted key

  The FIPS-enabled Fulcio image rejects encrypted PEM keys. This is a FIPS environment issue, not related to this PR. The PR doesn't touch Fulcio controller code, cert handling, or test fixtures.

  2. CTlog FIPS mode assertion (1 failure)

  Test: Securesign FIPS Install — Verify ctlog is running in FIPS mode

  Expected ctlog binary to be linked against libcrypto.so but the string wasn't found. This is a ctlog image build issue (FIPS binary wasn't compiled with the expected crypto backend), unrelated to conversion
  infrastructure.
```
This does not seems to be related to this PR. @JasonPowr are we OK to merge it?

bouskaJ (approved):

PR #1938, opened 2026-06-18
feat: use dedicated struct for Trillian status

bouskaJ:


bouskaJ:
crdify failure is expected, FIPS is a known issue, please ignore

osmman (changes requested):
**Add a deterministic unit test for status conversion**

The existing `TestTrillianConversionUnit` in `conversion_unit_test.go` only exercises spec fields — the `Status` is left as zero value in both hub and spoke test cases. The new `TrillianDBStatus ↔ TrillianDB` status conversion (particularly the `PvcName ↔ Pvc.Name` flattening and `DatabaseSecretRef` nil-guard) is only covered by the fuzz roundtrip test.

Consider adding a test case that explicitly documents the status field mapping with concrete values — e.g., a `TrillianStatus` with a non-empty `PvcName`, a non-nil `DatabaseSecretRef`, and TLS fields set on all three sub-structs. A deterministic test makes the expected mapping immediately obvious to future readers and catches regressions without relying on fuzz coverage.

osmman (approved):

PR #1939, opened 2026-06-18
feat: use dedicated struct for Fulcio status

bouskaJ:


bouskaJ:
crdify failure is expected, FIPS is a known issue, please ignore

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -62,6 +65,24 @@ func TestFulcioConversion(t *testing.T) {
 		Scheme: rhtasScheme(),
 		Hub:    &rhtasv1.Fulcio{},
 		Spoke:  &Fulcio{},
+		// Only fill fields that survive roundtrip — v1 status types omit spec-only fields.
+		FuzzerFuncs: []fuzzer.FuzzerFuncs{
+			func(_ runtimeserializer.CodecFactory) []interface{} {
+				return []interface{}{
+					func(s *FulcioStatus, c randfill.Continue) {
```
osmman, line 72: The custom fuzzer overrides the entire `FulcioStatus` but doesn't fill `ServerConfigRef` — it'll be nil in all fuzz iterations. Add `c.FillNoCustom(&s.ServerConfigRef)` alongside the other fields.

internal/controller/fulcio/actions/generate_cert_test.go
osmman, line 1: Tests don't exercise the `ObservedGeneration`-based change detection introduced in `CanHandle`. All test fixtures leave both `Generation` and `ObservedGeneration` at their zero defaults, so the generation check (`instance.Generation != cc.ObservedGeneration`) is never meaningfully tested. Existing test cases that set conditions with `ObservedGeneration` should use explicit non-zero values matching the instance's `Generation`.

New scenarios should cover generation mismatches — e.g., `Generation: 2` with `ObservedGeneration: 1` expecting re-reconciliation, verifying that spec changes are detected through the generation mechanism.

osmman (changes requested):

osmman (approved):

PR #1940, opened 2026-06-18
feat: switch to Go native FIPS, propagate GODEBUG to managed workloads, run integration test suite in strict fips mode 

JasonPowr:
## Summary
- Switch from OpenSSL-based FIPS to Go native FIPS (`GOFIPS140=v1.0.0`, `CGO_ENABLED=0`, `-tags=no_openssl`)
- Propagate operator's `GODEBUG` env var to all managed workloads (Deployments, StatefulSets, Jobs, CronJobs)
- Auto-detect FIPS clusters in e2e tests and adapt: swap MariaDB for PostgreSQL, use unencrypted certs, skip incompatible tests

## Test plan
- [ ] Unit tests pass for `SetGodebugEnv`
- [ ] e2e suite: 104 passed, 9 skipped (FIPS-incompatible),
- [ ] Update suite: 59/59 passed
- [ ] Non-FIPS clusters unaffected — detection returns false, all tests run as before


osmman (changes requested):
Good direction overall a few things need to be addressed before merging.

### Critical

**1. `fips140=auto` should be committed to `go.mod`**

`RUN go mod edit -godebug=fips140=auto` in `Dockerfile.rhtas-operator.rh` mutates `go.mod` only inside the build container. The checked-in `go.mod` doesn't match what's compiled, and local `make build` / `make run` produce binaries without FIPS auto-detection. Commit the directive to `go.mod` directly so builds are reproducible and local development matches container builds. This also means the Makefile `build` and `run` targets should set `GOFIPS140=v1.0.0` for full parity.

**2. Remove the upstream `Dockerfile`**

The Makefile now exclusively uses `Dockerfile.rhtas-operator.rh` for all docker targets (`docker-build`, `docker-build-skip-test`, `docker-buildx`). The upstream `Dockerfile` is dead code — no target references it. Remove it to avoid confusion.

### Major

**3. Add `rhtas.redhat.com/godebug` annotation for per-component override**

`SetGodebugEnv` unconditionally overwrites any existing GODEBUG value on all managed containers. Users have no way to override or disable this per-component for debugging or workarounds. Introduce a `rhtas.redhat.com/godebug` annotation following the same pattern as `rhtas.redhat.com/log-type`:
- Annotation set → use its value (override)
- Annotation not set → fall back to operator's GODEBUG env var (current behavior)
- Annotation set to empty → disable GODEBUG propagation for that component

Add it to `InheritableAnnotations` so it can be set on the Securesign CR for all components or on individual component CRs.

**4. Update FIPS metadata — Dockerfile label and bundle**

`features.operators.openshift.io/fips-compliant` is still `"false"` in `Dockerfile.rhtas-operator.rh`. The bundle metadata should also reflect FIPS support.

**5. Several test suites will fail on FIPS clusters**

These test files hardcode `VerifyAllComponents(ctx, cli, s, true, true)` without FIPS detection and will fail on FIPS clusters (they expect MariaDB which uses SHA1 auth):
- `test/e2e/benchmark/install_test.go`
- `test/e2e/custom_install/proxy_test.go`
- `test/e2e/custom_install/rolling_upgrade_test.go`
- `test/e2e/high_avalability/common_ha_install_test.go`
- `test/e2e/high_avalability/non_ha_to_ha_test.go`
- `test/e2e/upgrade_test.go`

Add `DetectAndConfigureFIPS` with PostgreSQL setup, or skip on FIPS clusters.

**6. Document that init containers are intentionally excluded from GODEBUG propagation**

`SetGodebugEnv` only operates on `Spec.Containers`, not `Spec.InitContainers`. Current init containers are shell/netcat so this is safe, but it should be a documented decision (code comment) so future contributors don't accidentally add a Go init container without GODEBUG.

JasonPowr:
Thanks for the review :) 

**1. fips140=auto should be committed to go.mod**
  - I intentionally keep this out of go.mod and apply it only in the container build. The value fips140=auto is specific to the [golang-fips/go fork](https://github.com/golang-fips/go) upstream Go only recognizes off, on, and only. Committing it to go.mod would cause a hard build error for anyone using the upstream Go toolchain (it's treated as an [unrecognized godebug setting](https://go.dev/doc/godebug)), which would break make build, make run, local tests, and CI for contributors not using the Red Hat Go fork. The Dockerfile approach ensures FIPS auto-detection is enabled in the shipped binary while keeping the codebase buildable with standard Go.
  
**4. Update FIPS metadata — Dockerfile label and bundle**
- I'd prefer to hold off on flipping features.operators.openshift.io/fips-compliant to "true" until the full stack is ready. We're still waiting on RHTAS Console and go-tuf CLI to land.

Everything else I am happy to make the changes :)

test/e2e/support/kubernetes/deployment.go
```diff
@@ -0,0 +1,74 @@
...
+		g.Expect(cli.Update(ctx, dep)).To(Succeed())
+	}).WithContext(ctx).Should(Succeed())
+
+	WaitForDeploymentRollout(ctx, cli, namespace, deploymentName, containerName, envs...)
+}
+
+func WaitForDeploymentRollout(ctx context.Context, cli ctrlclient.Client, namespace, deploymentName, containerName string, envs ...v1.EnvVar) {
+	Eventually(func(g Gomega) {
+		dep := &appsv1.Deployment{}
+		g.Expect(cli.Get(ctx, types.NamespacedName{
```
osmman, line 49: You need to use context from function param. The reason is that eventually is encapsulate it `.WithContext(ctx)` and it should properly propagate terminal signal.

test/e2e/support/kubernetes/olm/csv.go
```diff
@@ -0,0 +1,87 @@
...
+	// Check if any CSV owns this deployment. If not, fall back to direct patch.
+	csvList := &v1alpha1.ClusterServiceVersionList{}
+	if err := cli.List(ctx, csvList, client.InNamespace(namespace)); err != nil || !csvOwnsDeployment(csvList, deploymentName, containerName) {
+		k8ssupport.PatchDeploymentEnv(ctx, cli, namespace, deploymentName, containerName, envs...)
+		return
+	}
+
+	Eventually(func(g Gomega) {
+		list := &v1alpha1.ClusterServiceVersionList{}
+		g.Expect(cli.List(ctx, list, client.InNamespace(namespace))).To(Succeed())
```
osmman, line 30: same problem with context

test/e2e/support/kubernetes/deployment.go
JasonPowr: done :)

test/e2e/support/kubernetes/olm/csv.go
JasonPowr: done :)

osmman (approved):

PR #1963, opened 2026-06-25
feat: generic signer secret action replacing per-component implementations

osmman:
## Summary

Introduces `internal/action/generateSigner/` — a generic action for managing operator-managed signer secrets. Replaces 4 per-component implementations (Rekor, Fulcio, TSA, CTlog) with thin wrappers calling a single generic action.

**Core guarantees:**
- **Deterministic secret names** — eliminates the duplicate-secret bug class caused by `CreateOrUpdate` + `GenerateName`
- **Kubernetes immutable secrets** (`Immutable: true`) — API server enforces data immutability after creation
- **`kubernetes.CreateOrUpdate` with ensure functions** — follows codebase conventions for secret creation
- **TerminalError on mutations** — signer keys cannot be safely regenerated; most changes require manual intervention

**Config struct (5 callbacks):**
- `ResolveRef` — should use (user-provided refs, pre-existing secrets from previous operator versions). Returns nil when no pre-existing secret applies and the action hould generate new keys.
- `GenerateData` — produces key/cert material for fresh installation
- `AlignStatus` — copies secret refs into component-specific status fields
- `IsEnabled` (optional) — false for KMS/Tink paths
- `MutateSecret` (optional) — modify secret before creation (e.g., add TUF autodiscovery labels)

SECURESIGN-4794


internal/action/generateSigner/action.go
```diff
@@ -0,0 +1,235 @@
...
+func (i signerAction[T]) CanHandle(_ context.Context, instance T) bool {
+	w := i.wrapper(instance)
+	c := meta.FindStatusCondition(instance.GetConditions(), constants.ReadyCondition)
+
+	switch {
+	case c == nil:
+		return false
+	case state.FromCondition(c) < state.Pending:
+		return false
+	case !w.IsEnabled():
```
bouskaJ: The old code handled the secret→KMS transition inline: when KMS was not "secret", it cleared Status.Signer, nulled PublicKeyRef, and set ServerCondition=False. The new generic action simply returns false
    from CanHandle when IsEnabled() is false, so stale Status.Signer.KeyRef is never cleared.


Failure scenario: user switches Rekor from KMS: "" to KMS: "awskms://key" → stale secret ref stays in status → deployment mounts both the old secret and the KMS arg → Rekor starts with conflicting signer 
  config.

internal/action/generateSigner/types.go
```diff
@@ -0,0 +1,83 @@
...
+	"context"
+
+	rhtasv1 "github.com/securesign/operator/api/v1"
+	"github.com/securesign/operator/internal/apis"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Config holds the component-specific callbacks for the generic signer secret action.
+type Config[T apis.ConditionsAwareObject] struct {
```
bouskaJ: I would prefer using Interface rather than callbacks
'''
  type  interface {
      // Is secret-based signing active? (false = KMS/Tink)
      IsEnabled(instance T) bool

      // Find existing secret (user-provided ref, status migration)
      // (nil, nil) = not found, generate new one
      ResolveRef(ctx context.Context, instance T, c client.Client) (*SecretKeySelector, error)

      // Produce key/cert material for a new secret
      GenerateData(ctx context.Context, instance T, c client.Client) (map[string][]byte, error)

      // Write secret ref into component-specific status fields
      AlignStatus(instance T, ref SecretKeySelector)

      // Clean up stale status when transitioning to disabled
      OnDisabled(instance T)
  }
'''

  Why this is better than callbacks:

  1. Compiler enforces completeness — forget a method and it won't compile. With callbacks, a nil OnDisabled is a runtime panic waiting to happen.
  2. Testable in isolation — mock the strategy interface, test the generic action's lifecycle logic. Test each component's strategy against its own types. With callbacks, you test the whole bundle or nothing.
  3. Adding a 6th component = implement 5 methods, compiler tells you what's missing. With callbacks, you copy-paste a Config{} literal from another component and hope you got all the fields right.

internal/action/generateSigner/action.go
osmman: fixed

internal/action/generateSigner/types.go
osmman: fixed

internal/action/generateSigner/resolve.go
```diff
@@ -0,0 +1,49 @@
...
+
+	rhtasv1 "github.com/securesign/operator/api/v1"
+	"github.com/securesign/operator/internal/utils/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// RequireSecret verifies that the secret referenced by ref exists.
+// Returns nil if found, ErrSecretNotFound if the API confirms it does not exist,
+// or the raw API error for transient/permission failures.
+func RequireSecret(ctx context.Context, c client.Client, namespace string, ref *rhtasv1.SecretKeySelector) error {
```
bouskaJ: We may need this in other actions. I would move it to the `utils.kubernetes.secret`

internal/action/generateSigner/resolve.go
```diff
@@ -0,0 +1,49 @@
...
+
+// ResolveStatusSecret checks whether a status-referenced secret still exists.
+// It skips the check when the ref is nil, empty, or matches the deterministic
+// name (meaning the secret was created by the current operator version).
+//
+// Returns:
+//   - (name, nil)  — secret exists, caller should reuse it
+//   - ("", nil)    — ref is nil/empty, matches deterministic name, or secret not found
+//   - ("", err)    — transient/permission error, wraps ErrStatusSecretRead
+func ResolveStatusSecret(ctx context.Context, c client.Client, ref *rhtasv1.SecretKeySelector, namespace, deterministicName string) (*rhtasv1.SecretKeySelector, error) {
```
bouskaJ: Parameter order inconsistency — RequireSecret takes (ctx, c, namespace, ref) but  ResolveStatusSecret takes (ctx, c, ref, namespace, ...). Should be consistent

bouskaJ (approved):
Left a comment - minor issue. LGTM otherwise

internal/action/generateSigner/resolve.go
osmman: fixed

internal/action/generateSigner/resolve.go
osmman: fixed

PR #1966, opened 2026-06-26
refactor!: remove CommonName from FulcioCertStatus

osmman:
## Summary

Removes `CommonName` from `FulcioCertStatus` in the v1 API. This field was cached in status to survive cluster restore, but the cert is generated once and stored in an immutable secret — re-resolving CommonName has no effect since the cert is never regenerated.

All usage was already removed from the signer action in #1963.

Depends on #1963 (generic signer secret action).

SECURESIGN-4794


osmman:
CRDFY failing linter is expected due v1 API change.

bouskaJ (approved):

PR #1980, opened 2026-06-30
feat: resolve trust material from component CR status for TUF and CTlog autodiscovery

osmman:
## Summary

Moves TUF and CTlog autodiscovery from labeled secrets to component CR status fields, and adds drift detection so a component's cached trust material is never silently overwritten.

- **resolvePubKey action** (`internal/action/resolvePubKey`) — generic action with a `Resolver[T]` interface. Components implement `ComponentName()`, `CanHandle()`, `GetTrustMaterial()`, `SetTrustMaterial()`, `Resolve()`. Fetches and compares trust material on every reconcile — there is no time/generation-based throttle, since Rekor, Fulcio, TSA, and CTlog (including KMS-backed signers) only ever change their served key/cert via a process restart, which always produces a Deployment status change the owning controller already watches.

- **HTTP utils** (`internal/utils/http`) — shared `FetchFromAPI` (1 MiB response cap), `LoadTrustedCAs`, and a TLS-aware HTTP client builder used by all resolvers.

- **CTlog autodiscovery** — `handle_fulcio_root` reads `Fulcio.Status.CertificateChain` instead of searching labeled secrets, extracts the signing cert, and stores it in a CTlog-owned secret (`ctlog-fulcio-root-<name>`). CTlog controller now `Watches(&Fulcio{})` for rotation triggers.

- **TUF autodiscovery** — `resolve_keys` reads `Rekor.Status.PublicKey`, `CTlog.Status.PublicKey`, `Fulcio.Status.CertificateChain`, `TSA.Status.CertificateChain` into a TUF-owned secret (`tuf-keys-<name>`) with an owner reference. User-provided `SecretRef`s pass through unchanged.

- **Drift detection and manual acknowledgement** — a fetched value that differs from the cached one is flagged (`TrustMaterialAvailable` condition, Reason=Drifted) rather than silently accepted; Ready flips to `False` until acknowledged. Set the `rhtas.redhat.com/refresh-trust-material: "true"` annotation to accept the new value; the operator clears the annotation once applied. While unacknowledged, drift is re-verified on every reconcile: self-heals silently if the value reverts, re-affirms without duplicating the warning event otherwise.

- **TUF key rotation remains fully manual** — this PR does not add any automation for updating already-resolved TUF trust material. Rotating a component's key still requires following the documented `docs/*-key-rotation.md` procedure by hand.

Refs: SECURESIGN-4883


api/v1alpha1/conversion_overrides.go
```diff
@@ -32,3 +32,19 @@ func Convert_v1_TrillianSpec_To_v1alpha1_TrillianSpec(in *v1.TrillianSpec, out *
 func Convert_v1_TufSpec_To_v1alpha1_TufSpec(in *v1.TufSpec, out *TufSpec, s apiconversion.Scope) error {
 	return autoConvert_v1_TufSpec_To_v1alpha1_TufSpec(in, out, s)
 }
+
+// Manual conversion functions for Status types where v1 has fields
+// that don't exist in v1alpha1 (e.g. PublicKey, CertificateChain).
+// These fields are preserved via MarshalData/UnmarshalData annotation.
+
+func Convert_v1_CTlogStatus_To_v1alpha1_CTlogStatus(in *v1.CTlogStatus, out *CTlogStatus, s apiconversion.Scope) error {
```
bouskaJ: We should agree on where to place conversion functions. Currently half of them live in
  conversion_overrides.go and the other half in per-component conversion files (e.g.
  fulcio_conversion.go). I don't like conversion_overrides.go because it can grow unwieldy as we add
  more overrides. On the other hand, keeping them in the per-component files makes it immediately
  obvious what has been overridden.

internal/controller/rekor/actions/server/resolve_pub_key.go
```diff
@@ -1,173 +1,49 @@
...
-	return "resolve public key"
+func (r rekorTrustMaterialResolver) CanHandle(_ context.Context, instance *rhtasv1.Rekor) bool {
+	return state.FromInstance(instance, actions.ServerCondition) >= state.Initialize
 }
 
-func (i resolvePubKeyAction) CanHandle(_ context.Context, instance *rhtasv1.Rekor) bool {
-	return meta.IsStatusConditionTrue(instance.Status.Conditions, actions.ServerCondition) &&
-		instance.Status.PublicKeyRef == nil
+func (r rekorTrustMaterialResolver) GetTrustMaterial(instance *rhtasv1.Rekor) string {
+	return instance.Status.PublicKey
```
bouskaJ, line 24: Old CanHandle: IsStatusConditionTrue(ServerCondition) && PublicKeyRef == nil — stopped running once the key was resolved. New CanHandle: state >= Initialize — always true after initialization. Handle calls
  Resolve(), which does HTTP GET /api/v1/log/publicKey on every reconcile.

  A KMS signer's backing key can rotate externally, so actively re-fetching is correct in principle. However, it's worth being aware of this API call on every reconcile and limiting unnecessary executions where
  possible. Since the action won't run without a user spec change anyway, we can't rely on that to pick up KMS rotations automatically. It would be wise to give the user an explicit option to trigger key re-resolution (e.g.,
  an annotation or spec field) rather than polling on every reconcile.
(this apply for all components)

internal/utils/http/client.go
```diff
@@ -0,0 +1,113 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var serviceCAFiles = []string{
+	"/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt",
+	"/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+}
+
+var (
+	builderMu sync.RWMutex
```
bouskaJ, line 25: SetClientBuilder/GetClientBuilder introduce a global mutable singleton protected by a mutex, but the lock exists only to guard against concurrent test swaps — production code never mutates the builder.
  Consider eliminating the global state by passing the client builder into the resolver struct:
```
  type rekorTrustMaterialResolver struct {
      clientBuilder func(...[]byte) *http.Client
  }

  func NewResolvePubKeyAction() action.Action[*rhtasv1.Rekor] {
      return resolvePubKey.NewAction[*rhtasv1.Rekor](rekorTrustMaterialResolver{
          clientBuilder: httputils.DefaultClientBuilder,
      })
  }
```
  Tests construct the resolver with a mock builder directly — no global state, no mutex, no SetClientBuilder/ResetClientBuilder ceremony. The var builder, sync.RWMutex, and all three accessor functions can be  deleted.

bouskaJ (changes requested):

api/v1alpha1/ctlog_conversion.go
```diff
@@ -17,6 +17,7 @@ func (src *CTlog) ConvertTo(dstRaw conversion.Hub) error {
 	}
 	dst.Spec.ImagePullSecrets = restored.Spec.ImagePullSecrets
 	dst.Spec.TrustedCA = restored.Spec.TrustedCA
+	dst.Status.PublicKey = restored.Status.PublicKey
```
bouskaJ, line 25: Storing PEM trust material directly in CR status fields rather than secret references is intentional — it avoids requiring cross-namespace secret read RBAC for TUF when components are deployed in separate
  namespaces. The double storage (status field + tuf-keys-* / ctlog-fulcio-root-* secrets) is unavoidable: status serves as the cross-namespace transport, the secret is the pod-mountable runtime artifact.
  
  The etcd size impact is negligible (public keys ~400 B, cert chains ~4-8 KB against a 1.5 MiB per-object limit). Kubernetes itself uses the same pattern — CertificateSigningRequest.status.certificate stores
  full PEM inline. Worth adding a brief doc comment on the status fields noting the cross-namespace rationale so future readers don't question why data lives in status instead of a secret reference.

bouskaJ (changes requested):

api/v1alpha1/conversion_overrides.go
osmman: Sure lets standardise it per-component files.

internal/controller/rekor/actions/server/resolve_pub_key.go
osmman: @bouskaJ 

The reason for fetching on every reconcile is KMS/TINK key rotation. When a signer uses KMS-backed keys, rotation happens externally — the operator has zero access to KMS and no way to know when the backing key changes. The only source of truth is the running service API (e.g. `GET /api/v1/log/publicKey`), which always reflects the current active key.

An annotation-based trigger won't work here because there is no user action to trigger — KMS rotation is invisible to the operator.

To avoid unnecessary API calls, we can throttle with a dedicated `TrustMaterialAvailable` condition and a configurable refresh interval (operator-level default, e.g. 5 minutes):

- **True** — trust material resolved and fresh. If `LastTransitionTime` is within the refresh interval → skip fetch.
- **Unknown** (reason: NeedsRefresh) — interval elapsed, set condition to Unknown → requeue → next reconcile re-fetches.
- **False** — resolution failed, retry.

This way the HTTP call happens at most once per interval instead of every reconcile, and `LastTransitionTime` naturally tracks the last successful verification without extra status fields.

Do you see any issues with this approach?

api/v1alpha1/ctlog_conversion.go
osmman: The cross-namespace rationale is already documented in `internal/action/resolvePubKey/doc.go` (Motivation section). API field comments describe what the field contains, the design decision why is in the package doc.

internal/utils/http/client.go
osmman: @bouskaJ 

The struct field approach works for unit tests (construct the action with a mock builder directly), but doesn't solve envtests. Envtests set up one controller via `SetupWithManager` and swap mock HTTP responses per test case in `BeforeEach`. With a builder baked in at construction time, there's no way to vary mock behavior per test without relocating the same shared mutable state elsewhere.

The global `SetClientBuilder`/`GetClientBuilder` exists solely for this envtest constraint — production code never mutates the builder.

Happy to refactor if you see an approach that avoids this.

internal/controller/rekor/actions/server/resolve_pub_key.go
osmman: I added rate limiting based on time. Commit https://github.com/securesign/secure-sign-operator/pull/1980/changes/ad63c825aac95bba6d7e7cfcdac343398cf520ac

osmman: Dug into this more and it turns out Rekor/Fulcio/TSA/CTlog (KMS/Tink-backed included) all read their signing key/cert once at startup and never reload it, so a rotation on the KMS side alone never actually reaches the running pod. It only takes effect after a restart, and that's already a Deployment status change the controller watches. So the throttle wasn't protecting against anything real. I removed the refresh interval and just fetch/compare every reconcile now (cheap, since PersistStatus already does a Get+diff before writing).

internal/action/resolvePubKey/helpers.go
```diff
@@ -0,0 +1,175 @@
...
+		return false
+	}
+	b, _ := strconv.ParseBool(v)
+	return b
+}
+
+// patchAnnotations fetches a fresh copy of instance and applies mutate to its
+// annotation map, retrying on write conflicts. Used for bookkeeping that lives
+// in metadata rather than status (which is written via [action.BaseAction.PersistStatus]).
+func patchAnnotations(ctx context.Context, cli client.Client, instance client.Object, mutate func(map[string]string) map[string]string) error {
```
bouskaJ: Isn't it possible to use https://github.com/securesign/secure-sign-operator/blob/main/internal/utils/kubernetes/ensure/common.go#L29? with regular Ensure workflow?

internal/action/trustmaterial/helpers.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// ResolveBaseURL returns the base URL for in-cluster HTTP calls to a component's service.
+// When the operator runs inside a pod (ContainerMode), it uses the internal Kubernetes DNS name
+// which is always reachable. When running outside (local dev), it falls back to statusUrl.
+//
+// TODO: The http:// protocol is hardcoded. When internal TLS between ingress and pods is
+// implemented (planned for OCP), the protocol must be resolved dynamically from the
+// component's TLS configuration (e.g. service-serving-cert annotation or TLS status).
+func ResolveBaseURL(deploymentName, namespace, statusUrl string, port ...int) string {
```
bouskaJ, line 40: I would tend to always use statusUrl (it may refer to internal service if ingress is disabled) + resolve tls from component's Trust bundle. It's more transparent and easy to debug.

internal/utils/http/client.go
```diff
@@ -0,0 +1,117 @@
...
+				RootCAs:    pool,
+				MinVersion: tls.VersionTLS12,
+			},
+		},
+	}
+}
+
+// FetchFromAPI performs an HTTP GET request to the given URL and returns the response body.
+func FetchFromAPI(client *http.Client, url string) ([]byte, error) {
+	resp, err := client.Get(url) //nolint:gosec // URL is constructed by the caller from operator-controlled sources
```
bouskaJ: use `http.NewRequestWithContext` and allow request cancel on shutdown

internal/action/resolvePubKey/helpers.go
```diff
@@ -0,0 +1,175 @@
...
+			return condAware, nil
+		}
+	}
+	return nil, ErrNoReadyInstance
+}
+
+// ValidatePEM checks that data contains a PEM block that parses as either a
+// certificate or a public key.
+func ValidatePEM(data []byte) error {
+	block, _ := pem.Decode(data)
```
bouskaJ: https://pkg.go.dev/encoding/pem#Decode `block` is only the first pem block. If you pass the cert-chain (Fulcio,TSA) only the first block is validated. The rest is dropped with `_` as the second argument.

internal/action/resolvePubKey/action.go
```diff
@@ -0,0 +1,184 @@
+package resolvePubKey
```
bouskaJ: camelCase violates Go convention. Suggest trustmaterial since it handles certs and chains, not just public keys.

internal/controller/tuf/actions/resolve_keys.go
```diff
@@ -82,51 +149,41 @@ func (i resolveKeysAction) Handle(ctx context.Context, instance *rhtasv1.Tuf) *a
...
+	getTrustMaterial func(runtime.Object) string
 }
 
-func (i resolveKeysAction) discoverSecret(ctx context.Context, namespace string, key *rhtasv1.TufKey) (*rhtasv1.SecretKeySelector, error) {
-	labelName := labels.LabelNamespace + "/" + key.Name
-	s, err := k8sutils.FindSecret(ctx, i.Client, namespace, labelName)
-	if err != nil {
-		return nil, err
+var keyToSource = map[string]trustMaterialSource{
+	rekorKeyName:  {list: &rhtasv1.RekorList{}, getTrustMaterial: func(obj runtime.Object) string { return obj.(*rhtasv1.Rekor).Status.PublicKey }},
```
bouskaJ, line 163: unchecked type assertions can panic but this will be refactored with the service discovery impl so we may keep it for now.

internal/action/trustmaterial/helpers.go
osmman: Already tried simply use `statusUrl `. It broke Kind e2e tests. Kind ingress hostnames resolve via nip.io straight to `127.0.0.1`, which only reaches the ingress controller from the test runner's side (via Docker's port mapping). From inside the operator pod, that same `127.0.0.1` is just its own loopback so the operator ends up calling itself instead of the ingress controller. That's why we take the internal-service-DNS path in-container and only fall back to `statusUrl` outside the cluster.

This area needs another pass anyway once internal TLS between ingress and pods lands (see the TODO above this function). I'd rather fold this into that rework than patch it now. Keeping the dual-path logic as-is."

internal/controller/tuf/actions/resolve_keys.go
osmman: Lets keep it for now, it will be rewrited with service discovery.

internal/action/trustmaterial/helpers.go
bouskaJ: Let's keep it for now but I would rather see another TODO to remove the dual-path logic. Using a different URL than is visible to user is dangerous.

internal/utils/http/client.go
osmman: fixed

internal/action/resolvePubKey/action.go
osmman: fixed

internal/action/resolvePubKey/helpers.go
osmman: fixed

internal/action/trustmaterial/helpers.go
osmman: Will keep current dual-path for dev and CI environment.

internal/action/resolvePubKey/helpers.go
osmman: fixed

bouskaJ (approved):

PR #1982, opened 2026-07-01
feat: defaulting webhook

bouskaJ:
Assisted-by: Claude Code (claude-opus-4-6)

### DIFF
--- a/api/v1/ctlog_types_test.go
+++ b/api/v1/ctlog_types_test.go
@@ -3,34 +3,35 @@ package v1
 import (
 	"context"
 	"math"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	_ "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var _ = Describe("CTlog", func() {
 
 	Context("CTlogSpec", func() {
 		It("can be created", func() {
-			created := generateCTlogObject("ctlog-create")
+			created := generateMinimalCTlog("ctlog-create")
 			Expect(k8sClient.Create(context.Background(), created)).To(Succeed())
 
 			fetched := &CTlog{}
-			Expect(k8sClient.Get(context.Background(), getKey(created), fetched)).To(Succeed())
+			Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(created), fetched)).To(Succeed())
 			Expect(fetched).To(Equal(created))
 		})
 
 		It("can be updated", func() {
-			created := generateCTlogObject("ctlog-update")
+			created := generateMinimalCTlog("ctlog-update")
 			Expect(k8sClient.Create(context.Background(), created)).To(Succeed())
 
 			fetched := &CTlog{}
-			Expect(k8sClient.Get(context.Background(), getKey(created), fetched)).To(Succeed())
+			Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(created), fetched)).To(Succeed())
 			Expect(fetched).To(Equal(created))
 
 			var id int64 = 1234567890123456789
@@ -39,16 +40,30 @@ var _ = Describe("CTlog", func() {
 		})
 
 		It("can be deleted", func() {
-			created := generateCTlogObject("ctlog-delete")
+			created := generateMinimalCTlog("ctlog-delete")
 			Expect(k8sClient.Create(context.Background(), created)).To(Succeed())
 
 			Expect(k8sClient.Delete(context.Background(), created)).To(Succeed())
-			Expect(k8sClient.Get(context.Background(), getKey(created), created)).ToNot(Succeed())
+			Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(created), created)).ToNot(Succeed())
+		})
+
+		It("default constants are correct", func() {
+			created := generateMinimalCTlog("ctlog-literals")
+			Expect(k8sClient.Create(context.Background(), created)).To(Succeed())
+
+			fetched := &CTlog{}
+			Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(created), fetched)).To(Succeed())
+			Expect(fetched.Spec.MaxCertChainSize).To(Equal(ptr.To(int64(153600))))
+			Expect(fetched.Spec.Trillian.Port).To(Equal(ptr.To(int32(8091))))
+			Expect(fetched.Spec.Replicas).To(Equal(ptr.To(int32(1))))
+			Expect(fetched.Spec.Monitoring.Enabled).To(Equal(ptr.To(true)))
+			Expect(fetched.Spec.Monitoring.TLog.Enabled).To(Equal(ptr.To(false)))
+			Expect(fetched.Spec.Monitoring.TLog.Interval).To(Equal(&metav1.Duration{Duration: 10 * time.Minute}))
 		})
 
 		Context("is validated", func() {
 			It("public key", func() {
-				invalidObject := generateCTlogObject("public-key-invalid")
+				invalidObject := generateMinimalCTlog("public-key-invalid")
 				invalidObject.Spec.PublicKeyRef = &SecretKeySelector{
 					Key:                  "key",
 					LocalObjectReference: LocalObjectReference{Name: "name"},
@@ -60,7 +75,7 @@ var _ = Describe("CTlog", func() {
 			})
 
 			It("private key password", func() {
-				invalidObject := generateCTlogObject("private-key-password-invalid")
+				invalidObject := generateMinimalCTlog("private-key-password-invalid")
 				invalidObject.Spec.PublicKeyRef = &SecretKeySelector{
 					Key:                  "key",
 					LocalObjectReference: LocalObjectReference{Name: "name"},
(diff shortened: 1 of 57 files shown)


api/v1/securesign_webhook.go
```diff
@@ -9,18 +9,18 @@ import (
...
 func SetupSecuresignWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr, &Securesign{}).
 		WithDefaulter(&SecuresignDefaulter{}).
 		Complete()
 }
 
-func (d *SecuresignDefaulter) Default(_ context.Context, obj *Securesign) error {
+func (d *SecuresignDefaulter) Default(ctx context.Context, obj *Securesign) error {
 	securesignlog.Info("default", "name", obj.Name)
+	obj.Spec.SetDefaults()
```
osmman, line 22: ### M1 — Missing Securesign-level integration test (Medium)

Every other CRD type (CTlog, Fulcio, Rekor, TSA, Trillian, Tuf) has a "webhook defaults match SetDefaults" integration test, but the umbrella `Securesign` CR does not.

**Why it matters:** `SecuresignSpec.SetDefaults()` delegates to all sub-component `SetDefaults()` methods. Without a Securesign-level test, there is no integration-level proof that a minimal `Securesign` CR created via the API server receives correct defaults for all nested components. If someone later adds a sub-component and forgets to wire its `SetDefaults()` call, no test catches it.

**Suggested fix:**
```go
It("webhook defaults match SetDefaults", func() {
    created := generateMinimalSecuresign("securesign-defaults")
    expected := generateMinimalSecuresign("securesign-defaults")
    expected.Spec.SetDefaults()
    Expect(k8sClient.Create(context.Background(), created)).To(Succeed())
    fetched := &Securesign{}
    Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(created), fetched)).To(Succeed())
    Expect(fetched.Spec).To(Equal(expected.Spec))
})
```

api/v1/rekor_types.go
```diff
@@ -119,8 +108,7 @@ type RekorSigner struct {
 	//   - azurekms://keyname
 	//   - gcpkms://keyname
 	//   - hashivault://keyname
-	// +kubebuilder:validation:XValidation:rule="self == '' || self == 'secret' || self == 'memory' || self.matches('^awskms://.+$') || self.matches('^gcpkms://.+$') || self.matches('^azurekms://.+$') || self.matches('^hashivault://.+$')",message="KMS must be '', 'secret', 'memory', or a valid URI with a key path (e.g., awskms:///key-id)"
-	// +kubebuilder:default:=secret
+	// +kubebuilder:validation:XValidation:rule="self == 'secret' || self == 'memory' || self.matches('^awskms://.+$') || self.matches('^gcpkms://.+$') || self.matches('^azurekms://.+$') || self.matches('^hashivault://.+$')",message="KMS must be 'secret', 'memory', or a valid URI with a key path (e.g., awskms:///key-id)"
```
osmman, line 111: ### M2 — CEL validation tightened: empty string no longer accepted (Medium)

The old CEL rule accepted `self == ''` (empty string) as a valid value for KMS. This PR removes that option.

**Why it was safe before:** The old `+kubebuilder:default:=secret` filled in `"secret"` before validation ran, so users never actually submitted `""`.

**Why it's mostly safe now:** The new webhook also fills in `"secret"` before validation runs (mutating webhook -> validating webhook ordering is correct).

**What could go wrong:** If the webhook is temporarily unavailable and someone submits `kms: ""` (e.g., GitOps tool bypassing webhooks, server-side apply), the CEL rule now rejects it. Before, it was accepted. This is an edge case but could surprise users during upgrades.

**Suggested fix:** Either keep `self == '' || ...` in the CEL rule as a safety net, or explicitly document this as an intentional breaking change in release notes.

api/v1/default_utils.go
```diff
@@ -0,0 +1,14 @@
+package v1
+
+func setDefault[T comparable](dst *T, val T) {
+	var zero T
+	if *dst == zero {
```
osmman, line 8: ### M3 — `setDefault` panics on nil pointer (Medium)

This line dereferences `*dst` without checking if `dst` is nil. All **current** callers are safe (they pass `&s.Field` where `s` is a valid receiver), but a future caller passing a nil `*T` gets a panic with no useful error message.

**Suggested fix — add a nil guard (zero cost when `dst` is non-nil):**
```go
func setDefault[T comparable](dst *T, val T) {
    if dst == nil {
        return
    }
    var zero T
    if *dst == zero {
        *dst = val
    }
}
```

api/v1/ctlog_webhook.go
```diff
@@ -9,18 +9,18 @@ import (
...
 
 func SetupCTlogWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr, &CTlog{}).
 		WithDefaulter(&CTlogDefaulter{}).
 		Complete()
 }
 
-func (d *CTlogDefaulter) Default(_ context.Context, obj *CTlog) error {
+func (d *CTlogDefaulter) Default(ctx context.Context, obj *CTlog) error {
 	ctloglog.Info("default", "name", obj.Name)
```
osmman: ### L1 — Package-level logger loses request context (Low)

`ctloglog` is a package-level variable. When the API server calls this webhook, the `ctx` parameter carries request metadata (user identity, request ID, etc.). Using `logf.FromContext(ctx)` instead includes that metadata in every log line, which makes production debugging much easier when multiple concurrent requests hit the webhook.

This pattern applies to all `*_webhook.go` files in this PR (ctlog, fulcio, rekor, securesign, tsa, trillian, tuf).

**Suggested fix:**
```go
func (d *CTlogDefaulter) Default(ctx context.Context, obj *CTlog) error {
    log := logf.FromContext(ctx).WithName("ctlog-resource")
    log.Info("default", "name", obj.Name)
    obj.Spec.SetDefaults()
    return nil
}
```

api/v1/tuf_defaults.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	k8sresource "k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/utils/ptr"
+)
+
+func (s *TufSpec) SetDefaults() {
+	s.PodRequirements.SetDefaults()
+	s.ExternalAccess.SetDefaults()
+	setDefault((*string)(&s.SigningConfigURLMode), "external")
```
osmman: ### L3 — Type cast reduces readability (Low)

The cast `(*string)(&s.SigningConfigURLMode)` works but obscures the type. Since `TufSigningConfigURLMode` is a named `string` type and is `comparable`, a more readable alternative:

```go
setDefault(&s.SigningConfigURLMode, TufSigningConfigURLMode("external"))
```

Functionally identical — purely a readability preference.

osmman (changes requested):

api/v1/ctlog_types_test.go
```diff
@@ -39,16 +39,28 @@ var _ = Describe("CTlog", func() {
...
-			created := generateCTlogObject("ctlog-delete")
+			created := generateMinimalCTlog("ctlog-delete")
 			Expect(k8sClient.Create(context.Background(), created)).To(Succeed())
 
 			Expect(k8sClient.Delete(context.Background(), created)).To(Succeed())
-			Expect(k8sClient.Get(context.Background(), getKey(created), created)).ToNot(Succeed())
+			Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(created), created)).ToNot(Succeed())
+		})
+
+		It("webhook defaults match SetDefaults", func() {
```
osmman: ### new default-parity tests are tautological

**Severity: Medium.** Every component's old test suite asserted **literal** expected values (e.g. `Expect(fetched.Spec.Trillian.Port).To(Equal(int32(8091)))`). The PR replaces these with a single `"webhook defaults match SetDefaults"` test per component that does:

```go
created := generateMinimalX(...)
expected := generateMinimalX(...)
expected.Spec.SetDefaults()          // same function...
k8sClient.Create(ctx, created)        // ...that the webhook also calls
Expect(fetched.Spec).To(Equal(expected.Spec))
```

Both sides of the equality call the *same* `SetDefaults()` implementation. This correctly proves the webhook is **wired up** (a real improvement over the old no-op scaffolds), but it can no longer catch a wrong literal typo'd into `*_defaults.go` (e.g. `8091` → `8190`, `"10m"` → `"1m"`) — both sides would be wrong identically and the test would still pass.

**Affected:** `common` (no dedicated test file at all for the 10 shared defaults), `ctlog`, `rekor`, `tsa`, `trillian`, `tuf` all lost their literal-value assertions this way. `securesign_types_test.go` (new, 160 lines) only asserts literals for explicitly-overridden fields, not the propagation of defaults into child CRs. Contrast: the untouched `api/v1alpha1/*_types_test.go` files still hardcode literals for the frozen legacy API.

**Recommendation:** add back at least one literal-value assertion per component (e.g. `Expect(fetched.Spec.Trillian.Port).To(Equal(int32(8091)))`) alongside the new wiring test, so a regressed constant in `*_defaults.go` actually fails CI.

api/v1/tuf_types.go
```diff
@@ -159,7 +150,7 @@ func (s TufStatus) MatchesKeys(specKeys []TufKey) bool {
 //+kubebuilder:printcolumn:name="URL",type=string,JSONPath=`.status.url`,description="The component url"
 
 // Tuf is the Schema for the tufs API
-// +kubebuilder:validation:XValidation:rule="!(self.spec.replicas > 1) || ('ReadWriteMany' in self.spec.pvc.accessModes)",message="For deployments with more than 1 replica, pvc.accessModes must include 'ReadWriteMany'."
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.replicas) || !(self.spec.replicas > 1) || (has(self.spec.pvc.accessModes) && 'ReadWriteMany' in self.spec.pvc.accessModes)",message="For deployments with more than 1 replica, pvc.accessModes must include 'ReadWriteMany'."
```
osmman, line 158: ### Rekor/Tuf “replicas>1 needs ReadWriteMany” CEL rules 

Relocated from SecuresignSpec down to the standalone Rekor/Tuf CRDs (with a correct !has(self.spec.replicas) guard added). Enforcement isn’t lost, but a Securesign object with a bad combination now passes admission and only fails later, async, when the controller tries to create the child CR — fail-fast UX at kubectl apply time is gone.

osmman (approved):

PR #1985, opened 2026-07-01
fix: use pointer for TlogMonitoring.Interval 

bouskaJ:
bare Duration with no ommitempty was never defaulted

TlogMonitoring.Interval was a bare metav1.Duration without omitempty, so its zero value "0s" was always serialized into the JSON payload. The CRD default ("10m") never applied because the field was always present. This caused spec drift between Securesign and Rekor CRs in the CreateOrUpdate loop, preventing convergence.

Switch to *metav1.Duration with omitempty so the field is omitted when unset, allowing the CRD default to apply. Add CEL validation (>= 10s), manual v1alpha1 conversion (zero → nil for clean roundtrip), and controller-side nil+zero guard to protect existing stored objects from time.NewTicker(0) panic on upgrade.


fghanmi (approved):
LGTM! (updated)

bouskaJ:
> LGTM, pending lint fixes.

CRDIFY linter failure is expected - we are changing unreleased v1 API

osmman (approved):

PR #1990, opened 2026-07-02
feat: vanilla-Kubernetes support, monitoring fail-fast, startup detection, docs

anithapriyanatarajan:
Duplicate of https://github.com/securesign/secure-sign-operator/pull/1917

Raised from direct branch within the repo. The PR from fork had CI permission issue, hence raised this duplicate with exact same changes as in https://github.com/securesign/secure-sign-operator/pull/1917

This PR improves support for running the operator on vanilla Kubernetes by
making monitoring integration robust and clearly documented.

## Kubernetes platform support and documentation

- Added detailed guidance in `docs/kubernetes.md` and `README.md` for running on vanilla Kubernetes, covering prerequisites (cert-manager, server-side apply, image pull access), differences from OpenShift, and
  external access / hostname configuration.
- Updated the kubernetes overlay namespace from `sigstore-operator` to `rhtas-operator` to be consistent with the `openshift-rhtas-operator` naming convention.

## Monitoring integration

- Added `--monitoring` / `MONITORING` flag with auto-detection of the Prometheus Operator API (`monitoring.coreos.com/v1`) at startup using the discovery client. Handles both `apierrors.IsNotFound` and
  `*discovery.ErrGroupDiscoveryFailed` for reliable absence detection across K8s versions.
- **Fail-fast on missing CRD**: if `monitoring.enabled: true` but the ServiceMonitor CRD is not installed, `Handle()` now returns a clear error directing the user to install Prometheus Operator or set `monitoring.enabled: false`. This matches the project position established in PR #1440 — the CR spec must never be silently ignored.
- Moves the `kubernetes.IsMonitoringAvailable()` check from `CanHandle()` into `Handle()` for all 8 monitoring actions (ctlog, ctlog/monitor, fulcio, rekor/server, rekor/monitor, trillian/logserver, trillian/logsigner, tsa).
- Updated `monitoring_test.go` to assert the correct behavior: `CanHandle` returns `true` when `monitoring.enabled=true` regardless of CRD presence, so `Handle` always runs and produces a meaningful error.

Co-authored-by: [Claude Code](https://claude.com/claude-code)


cmd/main.go
```diff
@@ -156,6 +157,23 @@ func main() {
 		setupLog.Info("Platform explicitly configured via flag/env", "openshift", appconfig.Openshift)
 	}
 
+	if !utils.IsFlagProvided("monitoring", "MONITORING") {
```
osmman, line 160: This entire startup detection block should be removed — see the top-level review comment §2 for the full rationale.

TL;DR: single-shot discovery with no retry/backoff (compare with `DetectOpenShiftPlatform` right above which has 5-step exponential backoff and transient error handling). Any API hiccup during startup permanently disables monitoring until operator restart. The detection result is also not self-healing — installing Prometheus Operator after startup won't fix it without a restart.

Recommended: remove all startup detection and check at reconciliation time instead using `meta.IsNoMatchError` on the existing `CreateOrUpdate` error.

internal/utils/kubernetes/platform.go
```diff
@@ -100,6 +105,47 @@ func detectOpenShiftWithRetry(ctx context.Context, log logr.Logger, cl client.Cl
 	return found, nil
 }
 
+// DetectMonitoringAvailable detects whether the Prometheus Operator's
+// monitoring.coreos.com/v1 API (and the ServiceMonitor kind in particular) is
+// available on the cluster. It is used to gate ServiceMonitor creation so the
+// operator degrades gracefully on clusters without the Prometheus Operator
+// installed (e.g. vanilla Kubernetes) instead of failing the reconcile with
+// "no matches for kind ServiceMonitor". The result is memoized in
```
osmman, line 113: This function uses `discovery.NewDiscoveryClientForConfig` — an uncached, single-shot HTTP call to the API server with no retries. Compare with `detectOpenShiftWithRetry` above which handles transient errors (503, ECONNREFUSED, EOF, net.Error) with exponential backoff.

Recommend removing this entirely. The reconciliation-time approach (catching `meta.IsNoMatchError` from the existing `CreateOrUpdate` call) uses the manager's cached `DynamicRESTMapper` instead — zero extra API pressure and self-healing when CRDs are added/removed.

Also: [SECURESIGN-4584](https://redhat.atlassian.net/browse/SECURESIGN-4584) plans to split `MonitoringConfig`, making this flag/global obsolete.

internal/controller/fulcio/actions/monitoring.go
```diff
@@ -38,6 +38,10 @@ func (i monitoringAction) Handle(ctx context.Context, instance *rhtasv1.Fulcio)
 		err error
 	)
 
+	if !kubernetes.IsMonitoringAvailable() {
+		return i.Error(ctx, fmt.Errorf("monitoring.enabled is true but ServiceMonitor CRD is not installed; install Prometheus Operator or set monitoring.enabled=false"), instance)
```
osmman, line 42: Both suggestions in this error message are misleading with the current implementation:

1. **"install Prometheus Operator"** — with startup-only detection, installing it won't help until the operator is restarted (the detection already ran and set `config.MonitoringAvailable = false`)
2. **"set monitoring.enabled=false"** — the operator doesn't support ServiceMonitor cleanup, so the `true → false` transition is blocked by validation rules

If switched to reconciliation-time checking (see §2), suggestion 1 becomes accurate (next reconcile self-heals). Suggestion 2 should say: *"recreate the resource with monitoring.enabled=false"* since direct transition isn't supported.

docs/kubernetes.md
```diff
@@ -1,32 +1,119 @@
...
 
-## Metrics
+## Image overrides
+
+All component images default to `registry.redhat.io` (the OpenShift/product
+registry), which requires credentials. On vanilla Kubernetes you must override the
+images that have no unauthenticated pull path. Set the matching `RELATED_IMAGE_*`
+environment variable on the operator Deployment. Precedence is flag > env var >
+built-in default.
+
```
osmman, line 56: `quay.io/securesign` is **not a supported distribution channel**. While these images carry the same sha256 digests as `registry.redhat.io` product images, they are intermediate CI/build artifacts that can be pruned, deleted, or moved to a different registry at any time without notice. They are never supported for production use on any platform.

**The only supported production images come from `registry.redhat.io`.**

Instead of suggesting quay.io mirrors, this section should document how to authenticate with `registry.redhat.io` on vanilla Kubernetes:

1. **Per-component `imagePullSecrets` (recommended)** — the operator already supports `spec.imagePullSecrets` on every component CRD via the inlined `ServiceAccountConfig`. These propagate to the component's ServiceAccount automatically:
```yaml
spec:
  fulcio:
    imagePullSecrets:
      - name: redhat-pull-secret
  rekor:
    imagePullSecrets:
      - name: redhat-pull-secret
  # ... same for ctlog, trillian, tuf, tsa
```

2. **Global pull secret** on the default ServiceAccount or node-level config.

See [Red Hat Container Registry Authentication](https://access.redhat.com/articles/RegistryAuthentication) and [Registry Service Account management](https://access.redhat.com/terms-based-registry/) for creating non-expiring tokens.

If quay.io info is useful for contributors, it belongs in a development guide with clear "not for production" warnings.

README.md
```diff
@@ -11,6 +11,22 @@ Red Hat Trusted Artifact Signer enhances software supply chain security by simpl
...
 - [Kubernetes](docs/kubernetes.md) — Install via kustomize (EKS included)
 
+> **Vanilla Kubernetes at a glance:** the operator is OpenShift-first but runs on
+> generic Kubernetes (kind, EKS, GKE, ...). Note the prerequisites and differences,
+> all covered in the [Kubernetes guide](docs/kubernetes.md):
+> - **Server-side apply is required** (`kubectl apply --server-side -k ...`) — the
+>   `securesigns` CRD exceeds the 256 KB client-side annotation limit.
+> - **cert-manager is required** (operator webhooks); an **Ingress controller** is
+>   required only for external access; the **Prometheus Operator** is optional
+>   (if absent and `monitoring.enabled: true`, reconciliation fails with a clear
```
osmman, line 21: This is misleading — `quay.io/securesign` images are not a guaranteed or supported distribution channel (they're intermediate CI builds that can be pruned at any time). The only supported production images come from `registry.redhat.io`, regardless of platform. See the detailed comment on `docs/kubernetes.md`.

config/overlays/kubernetes/kustomization.yaml
```diff
@@ -1,6 +1,13 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
+# Override the base namespace (config/default sets "openshift-rhtas-operator")
+# with a platform-neutral name for vanilla Kubernetes installs. This retargets
+# every namespaced resource, including the cert-manager Certificate whose
+# namespace is propagated into the webhook configs/CRDs via the replacements
+# below, so they stay consistent.
```
osmman, line 8: This namespace rename (`openshift-rhtas-operator` → `rhtas-operator`) is a breaking change for anyone currently running the operator on Kubernetes. Please:
- Create a Jira Bug issue to track it
- Split into its own PR with migration guidance and release notes
- Include the corresponding `.github/workflows/main.yml` changes in that PR

osmman (changes requested):
## 1. Please split this PR

This PR bundles 4 independent concerns that should be reviewed and merged separately:

| Separate PR | Why |
|---|---|
| **Namespace rename** (`openshift-rhtas-operator` → `rhtas-operator`) + CI workflow updates | Breaking change — needs its own Jira Bug issue, migration docs, and release notes |
| **Kustomize: move `../prometheus` from `config/default` to `config/overlays/openshift`** | No technical objections, but it's an independent structural change |
| **Monitoring runtime fail-fast** (the `Handle()` checks in 8 actions) | Core behavior change — needs focused review (see §2 below for rework needed) |
| **Documentation** (`kubernetes.md`, `README.md`) | Can land independently, and has factual issues to fix first (see §3) |

---

## 2. Monitoring detection: remove the startup approach, check at reconciliation time instead

The startup detection (`DetectMonitoringAvailable`, `config.MonitoringAvailable`, `--monitoring`/`MONITORING` flag) introduces a lot of moving parts but still leaves users in a broken state. The core problem: **it's overengineered yet still underengineered**.

### What's wrong with startup detection

**No retry/backoff.** Compare with `DetectOpenShiftPlatform`, which has exponential backoff (5 steps), transient error classification (503, ECONNREFUSED, EOF, net.Error), and configurable timeout. `DetectMonitoringAvailable` is a single-shot discovery call — any API hiccup during startup silently and permanently disables monitoring until the operator is restarted. This has a high probability of happening on already-deployed instances during cluster upgrades or restarts.

| | `DetectOpenShiftPlatform` | `DetectMonitoringAvailable` (this PR) |
|---|---|---|
| Retry/backoff | Exponential: 2s→32s, 5 steps | None — single shot |
| Transient error handling | 503, timeout, 429, ECONNREFUSED, EOF | None |
| On failure | `os.Exit(1)` | Logs error, disables monitoring, continues |

**Not self-healing.** Unlike the OpenShift platform (a fixed property), monitoring CRD availability can change — Prometheus Operator may be installed after the operator starts. With startup-only detection, users who follow the error message advice ("install Prometheus Operator") will find it doesn't work without an operator restart that the message doesn't mention.

**The error message becomes misleading.** Both suggestions are wrong:
- *"install Prometheus Operator"* — won't help without restarting the operator (detection already ran)
- *"set `monitoring.enabled=false`"* — the operator doesn't support ServiceMonitor cleanup, so the `true → false` transition is blocked by validation rules

**Short-lived.** [SECURESIGN-4584](https://redhat.atlassian.net/browse/SECURESIGN-4584) plans to split `MonitoringConfig.Enabled` into `metricsEnabled` and `prometheusEnabled`. The `--monitoring` flag and `config.MonitoringAvailable` global will need to be removed when that lands.

### Recommended approach

Remove all of:
- `config.MonitoringAvailable` global
- `--monitoring` / `MONITORING` flag and env var
- `DetectMonitoringAvailable()` function
- `IsMonitoringAvailable()` helper
- The startup detection block in `cmd/main.go`

Instead, let the monitoring `Handle()` **fail at reconciliation time** when the ServiceMonitor CRD is absent. Either catch the "no matches for kind ServiceMonitor" error from `kubernetes.CreateOrUpdate`, or use the manager's `RESTMapper` to check CRD presence (it uses an in-memory cache — no extra API calls per reconciliation).

This gives you:
- **Self-healing** — install Prometheus Operator → next reconcile succeeds, no restart needed
- **No startup flakiness** — no startup-time API calls that can silently disable features
- **No global state** — no mutable config that drifts from cluster reality
- **Future-proof** — nothing to rip out when SECURESIGN-4584 lands
- **Simpler code** — less to write, less to test, less to maintain

**Example** — the existing `CreateOrUpdate` call already fails with `*meta.NoKindMatchError` when the CRD is missing (the controller-runtime client resolves the GVK via the RESTMapper before making any API call). You just need to catch it:

```go
import "k8s.io/apimachinery/pkg/api/meta"

// in Handle(), wrap the existing CreateOrUpdate error handling:
if _, err = kubernetes.CreateOrUpdate(ctx, i.Client,
    kubernetes.CreateServiceMonitor(instance.Namespace, DeploymentName), /* ... */); err != nil {
    if meta.IsNoMatchError(err) {
        return i.Error(ctx, fmt.Errorf(
            "monitoring.enabled is true but ServiceMonitor CRD is not installed; "+
            "install the Prometheus Operator or recreate the resource with "+
            "monitoring.enabled=false"), instance)
    }
    return i.Error(ctx, fmt.Errorf("could not create ServiceMonitor: %w", err), instance)
}
```

`meta.IsNoMatchError` is from `k8s.io/apimachinery/pkg/api/meta` — the same pattern controller-runtime uses internally. The dynamic RESTMapper caches discovery in memory, so this adds zero API pressure per reconciliation and auto-invalidates when CRDs are added/removed.

---

## 3. Image registry guidance: `quay.io/securesign` is not a supported distribution channel

See inline comment on `docs/kubernetes.md`.

---

## 4. Namespace rename needs a Jira Bug issue

Changing the Kubernetes overlay namespace from `openshift-rhtas-operator` to `rhtas-operator` is a breaking change for existing deployments. Please create a Jira Bug issue to track it, and include migration guidance and release notes in the separate PR.

---

## 5. Minor: test coverage

The new test file (`monitoring_test.go`) tests `CanHandle()` behavior that didn't change in this PR. The actual new logic — failing in `Handle()` when the CRD is missing — is untested. Please add test coverage for the fail-fast path. (Note: if the startup detection is removed per §2, the test approach simplifies significantly — no global state mutation needed.)

anithapriyanatarajan:
Thank you very much @osmman. I shall mark this PR as well to DRaft status and proceed as per the comments.

anithapriyanatarajan:
This draft shall be closed once a new PR for https://redhat.atlassian.net/browse/SECURESIGN-4912 is done

PR #1997, opened 2026-07-03
fix: resolve infinite reconcile loop on Securesign upgrade due to mis…

bouskaJ:
…sing CRD defaults

After upgrading from TAS 1.1.0, the Securesign CR lacks CRD-defaulted fields (trillian server/signer) causing ensure_trillian to detect a spec diff on every reconcile, permanently stuck on "Creating".


bouskaJ:
@osmman crdify complains but I believe, we should bypass the linter to keep upgrade working. WDYT?

fghanmi (approved):
LGTM! Thanks Jan!

tommyd450 (approved):

PR #1998, opened 2026-07-06
fix: move operator metrics ServiceMonitor to OpenShift overlay only

anithapriyanatarajan:
Issue summary: [kustomization.yaml ](https://github.com/securesign/secure-sign-operator/blob/main/config/default/kustomization.yaml#L23)included ../prometheus, which propagated the operator's own metrics ServiceMonitor into the kubernetes overlay. On vanilla Kubernetes (no Prometheus Operator), this causes kubectl apply -k config/overlays/kubernetes to fail with no matches for kind "ServiceMonitor".

Fix: Move ../../prometheus from default to openshift exclusively. The kubernetes overlay no longer includes any ServiceMonitor, making it deployable on vanilla Kubernetes. OpenShift behavior is unchanged.

Testing: Verified with kustomize build config/overlays/kubernetes (no ServiceMonitor) and kustomize build config/overlays/openshift (ServiceMonitor present).

Co-authored-by: Claude

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -20,7 +20,6 @@ resources:
 - ../rbac
 - ../manager
 - ../webhook
-- ../prometheus
 
 patches:
 - path: manager_metrics_patch.yaml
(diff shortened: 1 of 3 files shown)


anithapriyanatarajan:
Split PR in response to review comments from https://github.com/securesign/secure-sign-operator/pull/1990

osmman (approved):

PR #2006, opened 2026-07-09
fix: propagate cluster proxy config to in-cluster cosign e2e Job

kdacosta0:
## Summary
- E2e test `Sign image with cosign cli` was consistently failing on proxy clusters — the in-cluster cosign Job couldn't reach `ttl.sh` (`dial tcp: i/o timeout`)
- Root cause: Job container had no `HTTP_PROXY`/`HTTPS_PROXY` env vars, so it tried to connect directly instead of through the cluster proxy
- Fix reads proxy settings from the OpenShift `Proxy/cluster` object and sets them on the cosign Job container
- Also registers `configv1` in the test client scheme so it can read the Proxy object

Not a prod issue — the operator propagates proxy correctly via OLM/`ensure.SetProxyEnvs()`. This only affected the test which creates Jobs directly.

## Test plan
- [x] Verified fix on proxy cluster — test passes (149s vs previous 30min timeout)
- [x] Full e2e suite regression check

### DIFF
--- a/test/e2e/support/common.go
+++ b/test/e2e/support/common.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"strings"
 
+	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	v12 "k8s.io/api/apps/v1"
 	v13 "k8s.io/api/batch/v1"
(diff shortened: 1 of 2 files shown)


osmman:
this fix will be needed in `main` branch too

osmman:
You need https://github.com/securesign/secure-sign-operator/pull/2011 to pass FBC build step

osmman (approved):

PR #2013, opened 2026-07-10
feat: add FIPS mode enforcement for signing keys and algorithms

JasonPowr:
## Summary
Reject password-protected private keys in FIPS mode and restrict client signing algorithms to FIPS-approved set for Fulcio and Rekor.
- Add PasswordRef callback to generateSigner framework that returns a TerminalError when a password reference is set in FIPS mode
- Pass --client-signing-algorithms flag with FIPS-approved ECDSA and RSA algorithms to Fulcio and Rekor deployments when FIPS is enabled
- Add unit and e2e tests for both features across all components

### DIFF
--- a/internal/action/generateSigner/action_test.go
+++ b/internal/action/generateSigner/action_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/securesign/operator/internal/constants"
 	"github.com/securesign/operator/internal/state"
 	testAction "github.com/securesign/operator/internal/testing/action"
+	"github.com/securesign/operator/internal/utils/fips"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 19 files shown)


internal/action/generateSigner/action_test.go
```diff
@@ -445,3 +446,88 @@ func TestHandle(t *testing.T) {
...
 }
+
+func TestHandle_FIPSPasswordRefGuard(t *testing.T) {
+	original := fips.Enabled
+	fips.Enabled = func() bool { return true }
+	t.Cleanup(func() { fips.Enabled = original })
+
+	t.Run("rejects PasswordRef as terminal error", func(t *testing.T) {
+		g := NewWithT(t)
+		ctx := context.TODO()
```
osmman: get `context` from `t.Context()`

internal/action/generateSigner/types.go
```diff
@@ -55,6 +63,11 @@ type wrapper[T apis.ConditionsAwareObject] struct {
 }
 
 func (w *wrapper[T]) ResolveRef(ctx context.Context, c client.Client) (*rhtasv1.SecretKeySelector, error) {
+	if w.cfg.PasswordRef != nil && fips.Enabled() {
```
osmman: Please move it to `action.Handle` function and keep ResolveRef just to call ResolveRef function 

https://github.com/securesign/secure-sign-operator/blob/main/internal/action/generateSigner/action.go#L76

PR #2027, opened 2026-07-14
feat: validate user-provided crypto material for FIPS compliance

JasonPowr:
## Summary

Adds FIPS 140 compliance validation for user-provided cryptographic material across all components (CTlog, Fulcio, Rekor, TSA, Trillian, TUF).

A dedicated generic `fips.NewAction[T]` runs before the signer action in each controller's pipeline, validating private keys, public keys, and certificates against FIPS-approved algorithms and key sizes. Validation failures are terminal errors; transient failures (e.g. secret not yet available) allow retry.

- **`internal/action/fips/`** — Generic FIPS validation action with per-component `Config` callbacks
- **`internal/utils/fips/`** — PEM/DER validation functions, FIPS rules, and `ValidationError` type
- **6 `fips_validation.go` files** — Per-component wiring mapping spec fields to validation functions
- **`AppendSecretRef` / `AppendFIPSCondition` helpers** — Reduce boilerplate across controllers

### DIFF
--- a/internal/action/fips/action.go
+++ b/internal/action/fips/action.go
@@ -1,0 +1,115 @@
+package fips
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/securesign/operator/internal/action"
+	"github.com/securesign/operator/internal/apis"
+	"github.com/securesign/operator/internal/constants"
+	"github.com/securesign/operator/internal/state"
+	fipsutil "github.com/securesign/operator/internal/utils/fips"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// NewAction creates a generic FIPS validation action that checks crypto
+// material and password refs before the signer action runs.
+func NewAction[T apis.ConditionsAwareObject](
+	conditionType string,
+	component string,
+	wrapper func(T) *wrapper[T],
+) action.Action[T] {
+	return &fipsAction[T]{
+		conditionType: conditionType,
+		component:     component,
+		wrapper:       wrapper,
+	}
+}
+
+type fipsAction[T apis.ConditionsAwareObject] struct {
+	action.BaseAction
+	conditionType string
+	component     string
+	wrapper       func(T) *wrapper[T]
+}
+
+func (i fipsAction[T]) Name() string {
+	return fmt.Sprintf("validate %s FIPS compliance", i.component)
+}
+
+func (i fipsAction[T]) CanHandle(_ context.Context, instance T) bool {
+	if !fipsutil.Enabled() {
+		return false
+	}
+
+	w := i.wrapper(instance)
+
+	c := meta.FindStatusCondition(instance.GetConditions(), constants.ReadyCondition)
+	switch {
+	case c == nil:
+		return false
+	case state.FromCondition(c) < state.Pending:
+		return false
+	case !w.IsEnabled():
+		return false
+	default:
+		cc := meta.FindStatusCondition(instance.GetConditions(), i.conditionType)
+		return cc == nil || cc.Status != metav1.ConditionTrue || instance.GetGeneration() != cc.ObservedGeneration
+	}
+}
+
+func (i fipsAction[T]) Handle(ctx context.Context, instance T) *action.Result {
+	w := i.wrapper(instance)
+
+	if w.PasswordRef() != nil {
+		err := reconcile.TerminalError(fipsutil.ErrPasswordRefInFIPS)
+		return i.Error(ctx, err, instance,
+			metav1.Condition{
+				Type:               i.conditionType,
+				Status:             metav1.ConditionFalse,
+				Reason:             state.Failure.String(),
+				Message:            err.Error(),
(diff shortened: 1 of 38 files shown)


internal/action/generateSigner/action.go
```diff
@@ -117,6 +117,23 @@ func (i signerAction[T]) Handle(ctx context.Context, instance T) *action.Result
 		)
 	}
 	if resolvedRef != nil {
+		if fips.Enabled() {
+			if err := i.validateCryptoMaterial(ctx, instance, w); err != nil {
```
osmman: Why do you not simply pass `resolvedRef`?

internal/action/generateSigner/action.go
```diff
@@ -154,6 +171,30 @@ func (i signerAction[T]) Handle(ctx context.Context, instance T) *action.Result
 	}
 }
 
+func (i signerAction[T]) validateCryptoMaterial(ctx context.Context, instance T, w *wrapper[T]) error {
```
osmman: `instance` parameter is not used

internal/action/generateSigner/types.go
```diff
@@ -44,6 +60,12 @@ type Config[T apis.ConditionsAwareObject] struct {
 	// if the selector is non-nil (password-protected keys are forbidden in FIPS).
 	// Nil disables the FIPS password-ref guard.
 	PasswordRef func(T) *rhtasv1.SecretKeySelector
+
+	// CryptoMaterial extracts user-provided crypto material from Kubernetes secrets
+	// for FIPS validation. Called only in FIPS mode, only when ResolveRef returns
+	// a non-nil ref (user-provided material). Returns a slice of CryptoRef values
+	// to validate. Nil disables FIPS crypto validation.
+	CryptoMaterial func(context.Context, T, client.Client) ([]CryptoRef, error)
```
osmman: Will need to update package doc in `doc.go`

osmman (changes requested):
I think it'd be better to have a new, separate generic `fips` action instead of validating inside `generateSigner`. `CanHandle` would be `fips.Enabled()`. Its only job is verifying the crypto material and settings, nothing else.

I'd put it in the `Pending` phase, since it only needs the raw user-provided refs and doesn't depend on anything ResolveRef produces.

**Main benefit:** single responsibility `generateSigner` goes back to just resolving/creating/aligning secrets (no FIPS awareness) and the terminal-vs-retriable error handling lives in one place instead of being split across multiple actions. We could add for example integrate check for usage of `PrivateKeyPassword` or other future FIPS specific settings.

internal/controller/rekor/actions/server/fips_validation.go
```diff
@@ -0,0 +1,76 @@
...
+				if i.Spec.Signer.KeyRef != nil {
+					return i.Spec.Signer.PasswordRef //nolint:staticcheck
+				}
+				return nil
+			},
+			CryptoMaterial: rekorCryptoMaterial,
+			IsEnabled: func(i *rhtasv1.Rekor) bool {
+				return i.Spec.Signer.KMS == signerKMSSecret || i.Spec.Signer.KMS == ""
+			},
+		}),
```
osmman: `IsEnabled` skips the *whole* action for external KMS, so Redis TLS cert/key and sharding public key validation never run — even though that material is always local/user-provided regardless of signer backend. Move the KMS check into `CryptoMaterial` so only the private-key check is skipped.

internal/controller/tsa/actions/fips_validation.go
```diff
@@ -0,0 +1,44 @@
...
+				}
+				if err := fipsAction.AppendSecretRef(ctx, c, i.Namespace, i.Spec.Signer.CertificateChain.CertificateChainRef,
+					"spec.signer.certificateChain.certificateChainRef", fipsutil.ValidateCertificateChainPEM, &refs); err != nil {
+					return nil, err
+				}
+				return refs, nil
+			},
+			IsEnabled: func(i *rhtasv1.TimestampAuthority) bool {
+				return tsaUtils.IsFileType(i)
+			},
```
osmman: `IsEnabled` = `IsFileType`, so Kms/Tink signers skip the whole action — but `certificateChain.certificateChainRef` is required and unvalidated for every signer type. Same fix as Rekor: gate only the private-key check inside `CryptoMaterial`, not the whole action.

internal/action/fips/action.go
```diff
@@ -0,0 +1,115 @@
...
+	case !w.IsEnabled():
+		return false
+	default:
+		cc := meta.FindStatusCondition(instance.GetConditions(), i.conditionType)
+		return cc == nil || cc.Status != metav1.ConditionTrue || instance.GetGeneration() != cc.ObservedGeneration
+	}
+}
+
+func (i fipsAction[T]) Handle(ctx context.Context, instance T) *action.Result {
+	w := i.wrapper(instance)
```
osmman, line 64: `FIPSCompliant` condition is always created but when `IsEnabled` is false (see above), this action never runs to resolve it — CR reaches `Ready` with a stuck `Unknown` condition. Fixing the `IsEnabled` gating above should fix this too.

osmman (changes requested):

JasonPowr:
@osmman Thanks for the review :), should be good for another round

osmman (approved):

PR #2037, opened 2026-07-15
chore: CTLog API cleanup

bouskaJ:


api/v1/ctlog_types.go
```diff
@@ -72,6 +74,7 @@ type CTlogSpec struct {
 
 	// Max certificate chain size in bytes. Passed as --max_cert_chain_size.
 	//+optional
+	//+kubebuilder:validation:Minimum=1
```
osmman, line 77: I think that 0 or -1 value disables size limit

bouskaJ: I don't think so. The size is defaulted by https://github.com/securesign/certificate-transparency-go/blob/main/trillian/ctfe/ct_server/main.go#L91, then passed on https://github.com/securesign/certificate-transparency-go/blob/main/trillian/ctfe/ct_server/main.go#L453. See https://github.com/golang/go/issues/45101, https://go.dev/src/net/http/request.go?h=Reader and https://pkg.go.dev/io#LimitedReader

PR #2038, opened 2026-07-15
chore: Trillian API cleanup

bouskaJ:


api/v1/trillian_types.go
```diff
@@ -39,6 +39,7 @@ type TrillianSpec struct {
 
 	// MaxRecvMessageSize sets the maximum size in bytes for incoming gRPC messages handled by the Trillian logserver and logsigner
 	//+optional
+	//+kubebuilder:validation:Minimum=1
 	MaxRecvMessageSize *int64 `json:"maxRecvMessageSize,omitempty"`
```
osmman, line 43: Please check if 0 or -1 values disables message limits

bouskaJ: The default value here is `0` (https://github.com/google/trillian/blob/master/cmd/trillian_log_server/main.go#L79) so you are technically right. It does not disable it but set it to 4MB as https://pkg.go.dev/google.golang.org/grpc#MaxRecvMsgSize says but still a valid option :+1:

PR #2067, opened 2026-07-17
feat: make CTlog log prefix configurable

bouskaJ:
## Summary
- Adds `prefix` field to `CTlogSpec` with default `"trusted-artifact-signer"`, replacing the hardcoded value in server config generation and monitor statefulset
- Wires the field through `CreateCtlogConfig`, conversion roundtrip (both standalone CTlog and embedded Securesign), and CRD manifests

## Test plan
- [x] All CTlog controller tests pass
- [x] CTlog conversion roundtrip fuzz test passes (hub-spoke-hub)
- [x] Securesign conversion roundtrip fuzz test passes
- [x] Verified removing the conversion restore line causes test failure
- [ ] Deploy on cluster and verify CTlog starts with default prefix
- [ ] Deploy with custom prefix and verify it propagates to server config and monitor

### DIFF
--- a/internal/controller/ctlog/actions/server_config.go
+++ b/internal/controller/ctlog/actions/server_config.go
@@ -40,6 +40,7 @@ var serverConfigAnnotations = []string{
 	labels.LabelNamespace + "/trillianUrl",
 	labels.LabelNamespace + "/rootCertificatesHash",
 	labels.LabelNamespace + "/privateKeyRef",
+	labels.LabelNamespace + "/logPrefix",
 }
 
 func NewServerConfigAction() action.Action[*rhtasv1.CTlog] {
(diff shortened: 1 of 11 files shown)


internal/controller/ctlog/actions/server_config.go
```diff
@@ -151,7 +151,7 @@ func (i serverConfig) Handle(ctx context.Context, instance *rhtasv1.CTlog) *acti
 	}
 
 	var cfg map[string][]byte
-	if cfg, err = ctlogUtils.CreateCtlogConfig(trillianUrl, *instance.Status.TreeID, rootCerts, certConfig); err != nil {
+	if cfg, err = ctlogUtils.CreateCtlogConfig(trillianUrl, *instance.Status.TreeID, rootCerts, certConfig, instance.Spec.Prefix); err != nil {
```
osmman, line 155: 🔴 **Critical: changing `Spec.Prefix` on an existing CTlog doesn't regenerate this config**

`configMatchingAnnotations()` (used by `validateExistingSecret`, ~L343-347 in this file) only tracks `treeID`, `trillianUrl`, `rootCertificatesHash`, and `privateKeyRef` — not `Prefix`. So when `Prefix` changes on an already-reconciled CTlog:
- `Generation` bumps, but no tracked annotation changes, so `validateExistingSecret()` still reports the secret as valid.
- The `isSpecChange` branch just persists `ObservedGeneration` and returns early — this `CreateCtlogConfig` call (which now bakes `instance.Spec.Prefix` into the config) never re-runs, so the secret keeps serving the **old** prefix.

Meanwhile `internal/controller/ctlog/actions/monitor/statefulset.go` reads `Spec.Prefix` directly and reconfigures the monitor immediately — so the monitor starts querying the new prefix path while the CT log server still only serves the old one, breaking monitoring silently.

Suggested fix: add the prefix to `serverConfigAnnotations` and `configMatchingAnnotations` so a prefix change is detected as config drift and triggers regeneration.

osmman (approved):

PR #2089, opened 2026-07-21
feat: full CTLog path in status.url

bouskaJ:


bouskaJ:
```
 2. TUF init/migration: user-provided address loses prefix

  tuf_init_job.go and tuf_migration_job.go removed the manual fmt.Sprintf("%s/%s", url, instance.Spec.Ctlog.Prefix) append. This is correct for auto-resolved addresses (from Status.Url which now includes the
  prefix). But when the user provides a manual Spec.Ctlog.Address without the path, ResolveServiceAddress skips resolution (continue at line 40), and ServiceAsUrl returns the address as-is — no prefix appended.
  The ErrCtlogPrefixNotSpecified guard was also removed.

  The Fulcio deployment (deployment.go:91-112) still does the prefix append manually for user-provided addresses — this inconsistency suggests the TUF removal may be premature unless the intent is that
  user-provided addresses must include the full path.
```
will be fixed in another PR - I am trying to avoid a huge PR that contians CTLog status changes + ServiceRef binding changes

osmman (approved):

PR #2106, opened 2026-07-21
feat: replace tuftool with tufcli

fghanmi:
[SECURESIGN-2771]

[SECURESIGN-2771]: https://redhat.atlassian.net/browse/SECURESIGN-2771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR should be ready to review once tufcli is added to cli-stacks.

### DIFF
--- a/ci/registries.conf
+++ b/ci/registries.conf
@@ -77,9 +77,9 @@ location = "registry.redhat.io/rhtas/rekor-monitor-rhel9"
 location = "quay.io/securesign/rekor-monitor"
 
 [[registry]]
-location = "registry.redhat.io/rhtas/tuffer-rhel9"
+location = "registry.redhat.io/rhtas/tufcli-init-rhel9"
 [[registry.mirror]]
-location = "quay.io/securesign/tuffer"
+location = "quay.io/securesign/tufcli-init"
 
 [[registry]]
 location = "registry.redhat.io/rhtas/ctlog-monitor-rhel9"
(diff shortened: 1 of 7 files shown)


JasonPowr (commented):
Now that we have the go tooling, can you add a new test case in fips_test.go, https://github.com/securesign/secure-sign-operator/blob/main/test/e2e/fips/fips_test.go#L342, thanks in advance :)

ci/registries.conf
```diff
@@ -77,9 +77,9 @@ location = "registry.redhat.io/rhtas/rekor-monitor-rhel9"
 location = "quay.io/securesign/rekor-monitor"
 
 [[registry]]
-location = "registry.redhat.io/rhtas/tuffer-rhel9"
+location = "registry.redhat.io/rhtas/tufcli-init-rhel9"
```
osmman, line 80: Why there is init in image name?

fghanmi: this is not the cli itself, it's the initialization of the TUF repo (this was the purpose of tuffer, as is the current purpose of tufcli-init)

osmman: ok so we have two images one tufcli-init-rhel9 (some extra sauce) and secound tufcli-rhel9 (only tufcli)?

fghanmi: exactly (same as before)

osmman (approved):

PR #2111, opened 2026-07-22
feat: rename ExternalAccess API to Ingress in v1

osmman:
## Why

The v1 API's `ExternalAccess` struct is misnamed: the operator only ever creates a Kubernetes `Ingress` object — it never creates an OpenShift `Route` directly (OpenShift's platform auto-derives a Route from the Ingress). The `RouteSelectorLabels` field name also leaked OpenShift-Route terminology onto what is just generic Ingress labels.

## What

- Renamed `ExternalAccess` to `Ingress` in `api/v1`, and the embedded field (json tag `externalAccess` -> `ingress`) in `RekorSpec`, `FulcioSpec`, `TufSpec`, `TimestampAuthoritySpec`, and `ConsoleSpec.UI`.
- Renamed `RouteSelectorLabels` to `Labels` in `Ingress` and in the duplicate field on `RekorSearchUI`.
- Reworded doc comments that incorrectly implied the operator creates a `Route` directly.
- `v1alpha1` stays byte-identical and keeps using `externalAccess`/`routeSelectorLabels` — the conversion webhook transparently bridges it to/from the renamed `v1` fields, so existing `v1alpha1` consumers are unaffected. Added/extended manual conversion overrides (`Convert_v1alpha1_ExternalAccess_To_v1_Ingress` and friends) for a lossless round-trip.
- Moved CRD-specific conversion overrides that previously lived in `conversion_overrides.go` into their own CRD's conversion file (`rekor_conversion.go`, `fulcio_conversion.go`, `tuf_conversion.go`, `timestampauthority_conversion.go`, `ctlog_conversion.go`, `trillian_conversion.go`). `conversion_overrides.go` now holds only conversions genuinely shared across more than one CRD (`MonitoringConfig`, `TlogMonitoring`, `Ingress`/`ExternalAccess`).
- Fixed a pre-existing bug in the Rekor Search UI ingress action: it was reading the parent Rekor server's `Labels` instead of its own `RekorSearchUI.Labels`.
- Removed `CreateIngress` from `internal/utils/kubernetes/ingress.go` (dead code, no callers).
- Updated all controller/action call sites, CLI-download ingress, tests, and `docs/kubernetes.md`.


osmman:
crdify is expected to fail due API changes in v1

test/e2e/support/tas/securesign/securesign.go
```diff
@@ -85,11 +85,11 @@ func ChooseDefaults(fipsEnabled bool, namespace string) Opts {
 
 func WithExternalAccess() Opts {
```
bouskaJ: Consider renaming the function too.

bouskaJ (approved):
Just one note - WithExternalAccess() in test/e2e/support/tas/securesign/securesign.go — this function's name still says "ExternalAccess" but it now sets .Ingress.Enabled. Consider renaming to WithIngress() for consistency.
  Minor, but it'll confuse someone grepping later.

test/e2e/support/tas/securesign/securesign.go
osmman: fixed

PR #2112, opened 2026-07-22
Disable cluster-wide caching of Secret and ConfigMap data

osmman:
## Why

The manager's cached client (`mgr.GetClient()`) caches reads by default via controller-runtime's informer cache. `cmd/main.go` already restricts this cache for high-cardinality types the operator exclusively owns (Deployment, ReplicaSet, StatefulSet, Pod, Service, Ingress, CronJob, Job), but `Secret` and `ConfigMap` were missing from that list.

Because of that, the first typed `Get`/`List` against either type starts an unfiltered, cluster-wide informer that lists and watches every Secret/ConfigMap in every namespace the operator's RBAC allows (a cluster-scoped `ClusterRole`/`ClusterRoleBinding` with no `resourceNames`), retaining full object data in the shared cache for the life of the operator process.

- For `Secret`, this means decrypted key material for objects the operator was never asked to read can sit resident in memory indefinitely.
- For `ConfigMap`, the payloads aren't sensitive, but ConfigMaps are typically far more numerous cluster-wide than Secrets, reintroducing the same class of OOM risk that the existing `ByObject` restriction already addresses for the other types.

A label-filtered `ByObject` entry (mirroring the existing Deployment/Pod/Service pattern) isn't viable for these two types: this operator's CRDs support user-provided ("bring your own") Secret/ConfigMap refs — signer keys, TLS certs, CA refs, `TrustedCA` bundles — that are never labeled by the operator. A `Get()` on an object outside a `ByObject` selector returns a permanent `NotFound` with no live fallback, which would permanently break resolution of those user-provided refs.

## What

- Add `Client.Cache.DisableFor` for `corev1.Secret` and `corev1.ConfigMap` to the manager's `ctrl.Options`, so reads for these types go straight to the API server instead of the shared informer cache.
- Use the existing metadata-only `ExistsSecret` helper for the TUF migration job's root-key existence check, which previously used the full content-capable `GetSecret` but only needed to know whether the secret exists.
- Add a `GetSecretMetadata` helper (metadata-only `GET`, alongside the existing `ExistsSecret`/`FindSecret`/`ListSecrets`) and use it for CTlog's `validateExistingSecret`, which only ever reads the existing config secret's annotations and never its data.

### DIFF
--- a/internal/utils/kubernetes/common.go
+++ b/internal/utils/kubernetes/common.go
@@ -101,6 +101,17 @@ func FindByLabelSelector(ctx context.Context, c client.Client, list client.Objec
 	return c.List(ctx, list, client.InNamespace(namespace), listOptions)
 }
 
+func Create[T client.Object](ctx context.Context, cli client.Client, obj T, fn ...func(object T) error) error {
+	var err error
+	for _, f := range fn {
+		err = errors.Join(err, f(obj))
+	}
+	if err != nil {
+		return err
+	}
+	return cli.Create(ctx, obj)
+}
+
 func CreateOrUpdate[T client.Object](ctx context.Context, cli client.Client, obj T, fn ...func(object T) error) (result controllerutil.OperationResult, err error) {
 	err = retry.OnError(retry.DefaultRetry, func(err error) bool {
 		return apiErrors.IsConflict(err) || apiErrors.IsAlreadyExists(err)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -315,6 +315,18 @@ func main() {
 			RecoverPanic: ptr.To(true),
 		},
 		Cache: cacheOpts,
+		// Secret/ConfigMap are excluded from cacheOpts.ByObject: a label/field selector on
+		// these types would return a permanent NotFound for any object outside the selector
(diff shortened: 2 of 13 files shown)


internal/utils/kubernetes/common.go
```diff
@@ -101,6 +101,19 @@ func FindByLabelSelector(ctx context.Context, c client.Client, list client.Objec
 	return c.List(ctx, list, client.InNamespace(namespace), listOptions)
 }
 
+func Create[T client.Object](ctx context.Context, cli client.Client, obj T, fn ...func(object T) error) error {
+	return retry.OnError(retry.DefaultRetry, apiErrors.IsAlreadyExists, func() error {
```
bouskaJ: Retrying on `AlreadyExists` doesn't make sense for a create-only helper.  With a static Name, retries will all fail the same way until exhaustion.

bouskaJ (changes requested):

internal/utils/kubernetes/common.go
osmman: fixed

bouskaJ (approved):

PR #2114, opened 2026-07-22
feat: service ref for trillian service

bouskaJ:
Assisted-by: Claude Code (claude-opus-4-6)


internal/action/tree/action.go
```diff
@@ -217,14 +218,7 @@ func (i resolveTree[T]) handleJob(ctx context.Context, instance T) *action.Resul
...
 
-	switch {
-	case trillianService.Port == nil:
-		err = fmt.Errorf("%s: %v", i.Name(), ErrTrillianPortNotSpecified)
-	case trillianService.Address == "":
-		trillUrl = fmt.Sprintf("%s.%s.svc:%d", logserverDeploymentName, instance.GetNamespace(), *trillianService.Port)
-	default:
-		trillUrl = fmt.Sprintf("%s:%d", trillianService.Address, *trillianService.Port)
-	}
+	trillUrl, err = utils.ResolveInternalServiceUrl(ctx, i.Client, *trillianService, instance.GetNamespace(), &rhtasv1.Trillian{})
```
osmman, line 221: The new `ResolveInternalServiceUrl` call can return transient errors (K8s API blips, CR not yet created), but they're still wrapped in `TerminalError`, which blocks controller-runtime retries and can permanently strand the CR in Failure state. The same PR correctly avoids `TerminalError` for the identical call in the CTlog and Rekor deployment actions.

api/v1alpha1/conversion_overrides.go
```diff
@@ -1,13 +1,18 @@
...
 	"net/url"
+	"regexp"
+	"strconv"
 
 	v1 "github.com/securesign/operator/api/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apiconversion "k8s.io/apimachinery/pkg/conversion"
 )
 
+var portRe = regexp.MustCompile(`:(\d+)(?:/|$)`)
```
osmman, line 14: Both `portRe` regexes (here and in `api/v1alpha1/conversion_overrides.go`) duplicate host:port splitting. The stdlib already does this — [`net.SplitHostPort`](https://pkg.go.dev/net#SplitHostPort) handles it (including IPv6), so no regex is needed. Just strip the scheme prefix (e.g. `dns:///`) first, then pass the remainder to `net.SplitHostPort`. Worth consolidating both call sites onto this in `internal/utils/service_ref_resolver.go`.

internal/utils/service_ref_resolver.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func ResolveInternalServiceUrl(ctx context.Context, cl client.Client, serviceRef v1.ServiceReference, instanceNamespace string, instance client.Object) (string, error) {
+	if serviceRef.URL != "" {
+		return serviceRef.URL, nil
+	}
+	if serviceRef.Ref != nil && serviceRef.Ref.Name != "" {
+		ns := serviceRef.Ref.Namespace
+		if ns == "" {
+			ns = instanceNamespace
```
osmman: Should not be `serviceRef.Ref.Namespace` be configured by webhook defaulter? If for some reason passed empty value here than return error.

api/v1/common.go
```diff
@@ -152,6 +152,29 @@ type TsaService struct {
...
+// ServiceReference identifies a component service either by in-cluster CR
+// reference or by an external URL.
+// +kubebuilder:validation:XValidation:rule="!(has(self.ref) && has(self.url) && size(self.url) > 0)",message="ref and url are mutually exclusive"
+type ServiceReference struct {
+	// In-cluster reference to a component CR.
+	//+optional
+	Ref *ServiceReferenceRef `json:"ref,omitempty"`
+	// Direct URL for an external or cross-namespace service.
+	//+optional
+	URL string `json:"url,omitempty"`
```
osmman, line 167: When we will add validation webhook we should add check on url format (url.Parse). Currently we should verify it's value during reconciliation look.

osmman (changes requested):
- CTlog/Rekor controllers now depend on Trillian CRs via ResolveInternalServiceUrl but don't watch them — no event-driven re-reconcile if Trillian is created/updated later.
- Stale `doc.go` still references removed `*Service` types.

api/v1alpha1/conversion_overrides.go
bouskaJ: I know about [net.SplitHostPort](https://pkg.go.dev/net#SplitHostPort) The problem is with the `dns:///` (note 3 `/`) That means that the `host:port` is not the `host` part but the `path` SplitHostPort does not work on that.

I wanted to create something universal that will work even for _normal_ urls and even for `host:port` that is technically not valid but it is highly used.

Implementing all workarounds coming from ^^ lead me to the simple regexp solution.

internal/utils/service_ref_resolver.go
bouskaJ: yes, it is. I actually had that but I changed that on AI review comment :smiling_face_with_tear: I will change it back

api/v1/common.go
bouskaJ: yes, we are yet back at url.Parse() - AFAIK this is the only tool that parse URL. Problem is that has its own head:
```
host:80                                                  => Scheme="host" Host="" Path="" Opaque="80"
  host:port                                                => Scheme="host" Host="" Path="" Opaque="port"
  //host:80                                                => Scheme="" Host="host:80" Path="" Opaque=""
  //host:port                                              => ERROR: invalid port ":port" after host
  dns:///host:80                                           => Scheme="dns" Host="" Path="/host:80" Opaque=""
  dns:///host:port                                         => Scheme="dns" Host="" Path="/host:port" Opaque=""
  http://host:80/path                                      => Scheme="http" Host="host:80" Path="/path" Opaque=""
```
I may try to use regexp that will validate that `//` are always present - ensure the scheme does not misinterprets as host.

bouskaJ:
```
CTlog/Rekor controllers now depend on Trillian CRs via ResolveInternalServiceUrl but don't watch them — no event-driven re-reconcile if Trillian is created/updated later.
```
Trillian never changes its url so no need to watch it for now.

api/v1alpha1/conversion_overrides.go
bouskaJ: as you can see on https://github.com/grpc/grpc-go/issues/2403 there was ask to deliver grpc parser but it was not accepted.
```
I discussed this with the other grpc language leads, and they don't offer anything similar. Target URIs aren't intended to be parsed by users. We are investigating some different options to configure connection settings based on the scheme, but haven't made any concrete plans yet.
```
https://github.com/grpc/grpc-go/issues/2403#issuecomment-443028444

The regexp is the easiest option unless we do not want to re-implement the grpc parser.

bouskaJ:
> ```
> CTlog/Rekor controllers now depend on Trillian CRs via ResolveInternalServiceUrl but don't watch them — no event-driven re-reconcile if Trillian is created/updated later.
> ```
> 
> Trillian never changes its url so no need to watch it for now.

Added watcher for consistency and possible future usage. It may also push Services into error when trillian gets removed that is better than wait till the pod dies..

osmman (approved):

PR #2126, opened 2026-07-23
Tuf service ref

bouskaJ:


api/v1alpha1/conversion_overrides.go
```diff
@@ -93,30 +69,64 @@ func Convert_v1_Ingress_To_v1alpha1_ExternalAccess(in *v1.Ingress, out *External
...
+}
+
+func Convert_v1alpha1_CtlogService_To_v1_ServiceReference(in *CtlogService, out *v1.ServiceReference, _ apiconversion.Scope) error {
+	addressPortToServiceReference(in.Address, in.Port, out)
+	if out.URL != "" && in.Prefix != "" {
+		var err error
+		if out.URL, err = urlWithPath(out.URL, in.Prefix); err != nil {
+			return err
+		}
 	}
```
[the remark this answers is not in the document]
bouskaJ: legacy address have to had protocol too. It was just not validated.

api/v1/common.go
```diff
@@ -167,6 +167,8 @@ type ServiceReference struct {
 	URL string `json:"url,omitempty"`
 }
 
+func (s ServiceReference) GetServiceRef() ServiceReference { return s }
```
osmman, line 170: Why do you need method to return itself? Should not it be `return s.ServiceReferenceRef` based on name?

bouskaJ: It is an interface implementation.

api/v1/tuf_types.go
```diff
@@ -29,6 +29,15 @@ const (
 	TufKeyTSA    = "tsa.certchain.pem"
 )
 
+type ServiceRefWithOIDC struct {
+	ServiceReference `json:",inline"`
+	// OIDC Configuration
+	// +optional
+	// +listType=set
+	//+kubebuilder:validation:items:Pattern=`^[a-zA-Z][a-zA-Z0-9+.-]*://.+$`
+	OIDCIssuers []string `json:"oidcIssuers,omitempty"`
```
osmman, line 39: Why it is not pulling ODIC information from Fulcio object?

bouskaJ: this is for static (user provided) configuration. If serviceRef is used it loads from Fulcio

osmman: Can you extend description for that. Like that can override existing configuration or similiar.

api/v1/tuf_types.go
```diff
@@ -50,16 +59,20 @@ type TufSpec struct {
 	Pvc Pvc `json:"pvc,omitempty"`
 	// Ctlog service configuration
 	//+optional
-	Ctlog CtlogService `json:"ctlog,omitempty"`
+	//+kubebuilder:validation:XValidation:rule="!has(self.url) || size(self.url) == 0 || self.url.matches('^[a-zA-Z][a-zA-Z0-9+.-]*://[^/]+/.+$')",message="url must follow the pattern scheme://host[:port]/path"
+	Ctlog ServiceReference `json:"ctlog,omitempty"`
```
osmman, line 64: ## Service References Cannot Support TUF Metadata Structure

**Problem**: Current implementation uses a single `ServiceReference` per service, but TUF metadata (`trusted_root.json`, `signing_config.v0.2.json`) requires **lists** of services with validity ranges.

**Why**: TUF metadata has multiple entries per service type to support:
- Key rotation (same service, different keys over time)
- Log sharding (e.g., CTLog with different URL paths: `/2024/shard1`, `/2025/shard2`)

Reference: Sigstore TrustedRoot spec has `repeated` (lists) for `tlogs`, `certificate_authorities`, `ctlogs`, `timestamp_authorities`, each with `valid_for` time ranges.

**Current (broken)**:
```yaml
spec:
  ctlog:
    ref:
      name: ctlog
```
Only supports ONE CTLog reference.

**Needed**:
```yaml
spec:
  ctlog:
    - ref:
        name: ctlog
        namespace: trusted-artifact-signer
    - validFrom: "2024-01-01T00:00:00Z"
      validTo: "2025-01-01T00:00:00Z"
      url: https://ctlog.example.com/2024/shard1
      publicKey: <PEM bytes>
```

**Possible solution**:
```go
type TufServiceEntry struct {
    ServiceReference `json:",inline"`
    ValidFrom *metav1.Time `json:"validFrom,omitempty"`
    ValidTo   *metav1.Time `json:"validTo,omitempty"`
    PublicKey []byte       `json:"publicKey,omitempty"`
}

type TufSpec struct {
    // ...
    Rekor []TufServiceEntry `json:"rekor,omitempty"`
    Ctlog []TufServiceEntry `json:"ctlog,omitempty"`
    // ...
}
```

Each service field must be a **list** supporting refs AND explicit entries with validity ranges.

osmman: Example for 2 valid rekor instances from Sigstore:
- https://github.com/sigstore/root-signing/blob/main/targets/signing_config_rekor_v2.v0.2.json 
- https://github.com/sigstore/root-signing/blob/main/targets/trusted_root.json

bouskaJ: This is not the final shape of the TUF API — it will evolve further (e.g., publicKey option and auto-loading trust material from ServiceReference). However, for now we are intentionally keeping the static form
  with a single active entry per service type.

  Moving to a list-based model would require solving two significant problems first:

  - Split-view avoidance: We would need a mechanism to parse TUF repository history and reconcile it back into the CR, otherwise the CR and the TUF repo drift apart as competing sources of truth.
  - Live repository management: List changes would necessitate re-signing TUF metadata (automatic key rotation), including an opt-out path for users who manage signing keys outside the cluster.

  These are substantial design challenges on top of other TUF work (KMS support, object storage backends). A comprehensive TUF refactoring was explicitly scoped as out of bounds in the [ADR](https://docs.google.com/document/d/19yoFpzqJyE9RBhgtkQp09TcGV_-q53pi2uivp4EwcaM/edit?tab=t.0).

  This PR refactors only the static TUF repository form we already have. The single-entry-per-service shape is valid per the TrustedRoot spec and leaves a clean migration path to list fields later (singular →
  plural naming, conversion webhook). If someone wants to take on the auto-management work, contributions are very welcome — but I'd prefer to keep the static form for this version and evolve incrementally.

osmman: I can accept this as an intermediate solution, but we need to properly fix it before releasing the v1 API.

The alpha version has `spec.keys` which is the original source of data. All long-running deployments have generated data based on it. It allowed configuring whatever number of services controlled by a list of default values, so it has to be a list even now. In old key rotation scenarios `spec.keys` was used to reference old shard public keys. The later addition of `CtlogService`, `FulcioService`, `RekorService`, `TsaService` as parallel extended data butchered the whole function of the keys list and enforced it to be only a static list, which was not the original design.

I don't expect a 1:1 conversion for this. For proper migration we would need to read the exact data stored in the TUF repository and extract them to spec. Any changes to spec on an already initialized repo have zero impact.

For now you can skip `ValidFrom`, `PublicKey`, `OIDCIssuers` and other overrides and set only `ServiceReference`. These can be added later as optional fields and will pass the codify check for v1 API.

internal/utils/service_ref_resolver.go
```diff
@@ -14,31 +15,58 @@ import (
...
+func ResolveExternalServiceUrl(ctx context.Context, cl client.Client, serviceRef apis.ServiceReferencer, instanceNamespace string, instance apis.AddressableConditionAware) (string, error) {
+	ref := serviceRef.GetServiceRef()
+	if ref.URL != "" {
+		return ref.URL, nil
+	}
+
+	if err := serviceRefOrAutoload(ctx, cl, ref, instanceNamespace, instance); err != nil {
+		return "", err
+	}
+	if !meta.IsStatusConditionTrue(instance.GetConditions(), "Ready") {
```
osmman: Use already existing Condition constant for Ready

internal/controller/rekor/actions/server/deployment.go
```diff
@@ -109,10 +109,11 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1.Rekor) *acti
 
 func (i deployAction) ensureServerDeployment(instance *rhtasv1.Rekor, sa string, labels map[string]string, internalTrillianUrl string) func(*v2.Deployment) error {
 	return func(dp *v2.Deployment) error {
-		m := portRe.FindStringSubmatchIndex(internalTrillianUrl)
-		if m == nil {
+		matches := portRe.FindAllStringSubmatchIndex(internalTrillianUrl, -1)
+		if len(matches) == 0 {
 			return fmt.Errorf("error parsing Trillian URL %q: no port found", internalTrillianUrl)
```
osmman: can you wrap error like https://github.com/securesign/secure-sign-operator/pull/2126/changes#diff-4f50435c3cb2844552499b43c19d70127630138973b60f46fb0625940450dcf3R127 

it allow to test error via `errors.Is()` and it will be consistent with other errors.

internal/controller/tuf/tuf_controller_test.go
```diff
@@ -194,10 +193,10 @@ var _ = Describe("TUF controller", func() {
 					}, Equal(state.Pending.String())),
 					WithTransform(func(condition *metav1.Condition) string {
 						return condition.Message
-					}, ContainSubstring("no items found")),
+					}, ContainSubstring("failed to resolve service url")),
```
osmman, line 196: You use ErrorResolveServiceUrl

internal/utils/service_ref_resolver_test.go
```diff
@@ -24,104 +24,170 @@ func testScheme() *runtime.Scheme {
...
-	}, "ns", &rhtasv1.Trillian{})
+	instance := &rhtasv1.Trillian{}
+	err := serviceRefOrAutoload(t.Context(), cl, rhtasv1.ServiceReference{
+		Ref: &rhtasv1.ServiceReferenceRef{Name: "my-trillian"},
+	}, "instance-ns", instance)
 
-	g.Expect(err).ToNot(HaveOccurred())
-	g.Expect(u).To(Equal("dns:///trillian-logserver.ns.svc:8091"))
+	g.Expect(err).To(HaveOccurred())
+	g.Expect(err).To(MatchError(ContainSubstring("failed to get service")))
```
osmman: You can use existing ErrServiceNotReady

osmman (approved):

PR #2128, opened 2026-07-23
feat: add PKCS#11/HSM support for Fulcio and CTLog signing

sampras343:
## Summary

Add PKCS#11/HSM support to the operator for **both Fulcio** (CA certificate signing) **and CTLog** (Signed Tree Head signing), enabling enterprises to use Hardware Security Modules for all signing keys via a vendor-agnostic plugin model. Integrates into the `spec.signer` hierarchy.

### Fulcio PKCS#11 (`spec.fulcio.signer`)
- `type: pkcs11` on `FulcioSigner` with CEL mutual-exclusion rules
- `FulcioPKCS11Config`: `configRef` (crypto11.conf) + `keyConfig` (HSM key ID/label)
- `signer.auth` (`Auth` struct) for server container env vars — replaces custom `CredentialsRef`/`ServerEnv`
- `signer.certificateChain.certificateChainRef` for pre-provisioned root CA from key ceremony
- Top-level `initContainers`/`volumes`/`volumeMounts` on `FulcioSpec` (pod-wide, not nested under pkcs11)

### CTLog PKCS#11 (`spec.ctlog.signer`)
- `type: pkcs11` on `CTlogSigner` with CEL mutual-exclusion rules
- `CTlogPKCS11Config`: `pinSecretRef`, `publicKeyRef`, `tokenLabel`, `modulePath`, `persistence`
- `signer.auth` for server container env vars (uniform with Fulcio)
- Config-driven dispatch: generates `keyspb.PKCS11Config` protobuf instead of `keyspb.PEMKeyFile`
- `Status.PublicKeyRef` populated from `EnsurePKCS11Config` for trust material resolution
- Top-level `initContainers`/`volumes`/`volumeMounts` on `CTlogSpec`

### Design decisions (per review feedback)
- **Reuse `Auth` struct** instead of custom `CredentialsRef`/`ServerEnv` — matches TSA's pattern
- **Move `initContainers`/`volumes`/`volumeMounts` to top-level spec** — pod-wide resources, not backend-specific
- **Drop `PKCS11Status` structs and drift detection** — `ObservedGeneration` suffices
- **Rename `configRef`/`modulePath`** — drop redundant `PKCS11` prefix (already scoped under `pkcs11:`)
- **Drop `initContainers` CEL rule** — allow CSI drivers/webhook injectors to provision libraries without declaring init containers
- **Stop auto-injecting `HSM_PIN`** — users set it explicitly in `initContainers[].env` if their vendor image needs it
- **Auth on `Signer`** — uniform placement for both Fulcio and CTLog
- **Call `Signer.SetDefaults()` from `SecureSign.SetDefaults()`** — `signer.type: file` is explicit on the parent CR, CEL rules don't need `!has(self.type)` guards

### Shared
- `PKCS11InitContainerSpec` — curated `corev1.Container` subset shared by both components
- v1alpha1 conversion: PKCS#11 fields preserved via `MarshalData` annotations (v1 API only)
- Vendor-agnostic sample CR with `<placeholder>` values

## E2E Validation (OCP 4.22)

| Test | Result |
|------|--------|
| Fulcio PKCS#11 with SoftHSM (P-256) | cosign sign + verify PASS |
| CTLog PKCS#11 with SoftHSM (P-256) | cosign sign + verify PASS |
| Dual PKCS#11 (Fulcio + CTLog) | cosign sign + verify PASS |
| File mode regression (no PKCS#11 fields) | cosign sign + verify PASS |
| `signer.type: file` defaulted on SecureSign CR | Verified |
| PKCS#11-built ct_server in file mode | Zero impact, identical behavior |

## Dependency

**SECURESIGN-5021:** RHTAS ct_server image must be built with `CGO_ENABLED=1 -tags=pkcs11,no_openssl`. The `go-toolset` base image already includes `gcc` and `glibc-devel` — no `dnf install` needed (hermetic build compatible).

## Test plan

- [ ] `make build` and `make test` pass
- [ ] Deploy file-mode CR — `signer.type: file` defaulted, no PKCS#11 conditions/args/init containers
- [ ] Deploy dual PKCS#11 CR — full stack reaches Ready
- [ ] `cosign sign + verify` passes in both file and PKCS#11 modes
- [ ] v1alpha1 round-trip: PKCS#11 fields preserved through conversion
- [ ] Mode switch: change `signer.type` from `pkcs11` to `file` — PKCS#11 artifacts cleaned up

Jira: SECURESIGN-5014


internal/controller/ctlog/utils/ctlog_config.go
```diff
@@ -175,6 +175,67 @@ func CreateCtlogConfig(trillianUrl string, treeID int64, rootCerts []RootCertifi
...
+	rootPems := make([]string, 0, len(rootCerts))
+	for i := range rootCerts {
+		rootPems = append(rootPems, fmt.Sprintf("%sfulcio-%d", rootsPemFileDir, i))
+	}
+
+	logConfig := configpb.LogConfig{
+		LogId:          treeID,
+		Prefix:         "trusted-artifact-signer",
+		RootsPemFile:   rootPems,
+		PrivateKey:     mustMarshalAny(pkcs11Key),
```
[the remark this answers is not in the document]
sampras343: Fixed in 8bb1ae8a. `CreateCtlogPKCS11Config` now takes a `logPrefix` parameter (matching file-mode `CreateCtlogConfig`), and `buildPKCS11Config` passes `instance.Spec.Prefix`. Good catch — this would have silently broken any deployment with a non-default prefix.

internal/controller/ctlog/actions/deployment.go
```diff
@@ -195,6 +220,189 @@ func (i deployAction) ensureDeployment(instance *rhtasv1.CTlog, sa string, label
...
+	}
+
+	// Operator-managed hsm-lib-export container: copies the PKCS#11 .so
+	// from the vendor image to the shared lib volume.
+	if pkcs11ModulePath != "" && len(specs) > 0 {
+		desiredNames[HSMLibExportContainerName] = struct{}{}
+		libExport := kubernetes.FindInitContainerByNameOrCreate(podSpec, HSMLibExportContainerName)
+		libExport.Image = specs[0].Image
+		libExport.Command = []string{"cp", pkcs11ModulePath, fmt.Sprintf("%s/", HSMLibMountPath)}
+		libExport.VolumeMounts = []core.VolumeMount{
```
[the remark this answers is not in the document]
sampras343: Fixed in 8bb1ae8a. Added a CEL rule `size(self.initContainers) > 0` on `CTlogPKCS11Config` and changed the field from optional to required. This rejects the invalid CR at admission time with a clear message: "at least one initContainer is required to provision the PKCS#11 library". Without the init container, `hsm-lib-export` is never created and the `--pkcs11_module_path` flag would point to an empty emptyDir volume.

osmman (changes requested):
**Request changes** — API isn't released yet, let's fix the shape now.

1. **Reuse `Auth` instead of `CredentialsRef`/`ServerEnv`.** We already have this pattern (Rekor/TSA/Trillian) — add `Auth` on `FulcioCert`, drop `CredentialsRef`/`ServerEnv`. Exception: CTLog's `PinSecretRef` stays, since the operator reads that value itself to build ct_server's config. Also: don't auto-inject the PIN into every init container — most don't need it, let users add it explicitly where they do.

2. **Move `InitContainers`/`Volumes`/`ServerVolumeMounts` out of `pkcs11` config to `FulcioSpec`/`CTlogSpec` top level.** These are pod-wide regardless of nesting, and not PKCS#11-specific. Rename `ServerVolumeMounts` → `VolumeMounts` while moving it (no longer needs the "Server" qualifier once it's not sitting next to per-init-container mounts).

3. **Drop the "at least one initContainer required for pkcs11" CEL rule.** It assumes one specific way of provisioning the HSM library — other valid approaches exist (CSI driver, mutating webhook) that wouldn't populate `initContainers` at all. Let pod startup failure be the real signal if provisioning didn't work.

4. **Drop `CTlogPKCS11Status`/`FulcioPKCS11Status` entirely.** Every field is a required spec value copied verbatim — the operator never generates or modifies any of them, so they don't belong in status. Their `driftDetected`/`hasPKCS11ConfigDrift` checks are redundant too (`ObservedGeneration` already covers it).

5. **(Discuss) Consider `spec.signer` shape, matching Rekor/TSA.** Fulcio uses `certificate`, CTLog uses flat fields — worth converging while unreleased. Bigger diff than 1-4, especially for Fulcio — fine as a fast-follow if too much for this PR.

Happy to go through any of these in more detail — the reasoning behind them.

sampras343 (commented):
All five points addressed in the latest push (`71a6670a`). Here is how each was resolved:

**1. Reuse `Auth` instead of `CredentialsRef`/`ServerEnv`**

Done. `CredentialsRef` and `ServerEnv` removed from both `FulcioPKCS11Config` and `CTlogPKCS11Config`. `Auth *Auth` added on `FulcioSigner` and `CTlogSigner` (uniform placement on both — uses the existing `Auth{Env, SecretMount}` type from `common.go`). CTLog `PinSecretRef` stays as requested — operator reads it server-side to embed in the protobuf config. HSM_PIN auto-injection into init containers removed (`HSMPinEnvVar` constant deleted) — users set it explicitly in `initContainers[].env` if their vendor image needs it. Auth env vars injected idempotently using `FindEnvByNameOrCreate` to prevent duplicate env vars on reconciliation.

**2. Move `InitContainers`/`Volumes`/`ServerVolumeMounts` to top-level**

Done. All three fields moved from `FulcioPKCS11Config`/`CTlogPKCS11Config` to `FulcioSpec`/`CTlogSpec`. `ServerVolumeMounts` renamed to `VolumeMounts`. Not placed on `PodRequirements` — only Fulcio/CTLog get these fields. Controller code updated to read from `instance.Spec.InitContainers`/`instance.Spec.Volumes`/`instance.Spec.VolumeMounts` instead of `pkcs11Config.*`.

**3. Drop the `initContainers` CEL rule**

Done. The `has(self.initContainers) && size(self.initContainers) > 0` rule removed from `CTlogPKCS11Config`. The `initContainers` field changed back from required to optional. Vault Agent Injector, CSI drivers, and cert-manager can now provision the PKCS#11 library without declaring init containers in the CR.

**4. Drop `CTlogPKCS11Status`/`FulcioPKCS11Status`**

Done. Both status structs deleted. `Status.PKCS11` field removed from `CTlogStatus` and `FulcioStatus`. `hasPKCS11ConfigDrift`/`driftDetected` functions deleted — `ObservedGeneration` check in `CanHandle` handles all spec changes. `Status.PublicKeyRef` (on CTlogStatus directly, not on the deleted PKCS11 sub-struct) retained for trust material resolution. All conversion code updated — `Status.PKCS11` restoration lines removed.

**5. `spec.signer` shape**

Done in this PR (not deferred). Both Fulcio and CTLog now use the `spec.signer` hierarchy introduced by the signer restructure PRs (#2155, #2156). PKCS#11 config is at `signer.pkcs11`, type selector is `signer.type: pkcs11`. Additionally, `SecureSign.SetDefaults()` now calls `Signer.SetDefaults()` for both Fulcio and CTLog, so `signer.type: file` is explicit on the parent CR — CEL rules do not need `!has(self.type)` guards.

**Also fixed from Qodo review:**
- Hardcoded CTLog prefix in `CreateCtlogPKCS11Config` → now accepts `logPrefix` parameter
- Field renames: `pkcs11ConfigRef` → `configRef`, `pkcs11ModulePath` → `modulePath` (already scoped under `pkcs11:`)

internal/controller/ctlog/actions/deployment.go
```diff
@@ -191,6 +216,178 @@ func (i deployAction) ensureDeployment(instance *rhtasv1.CTlog, sa string, label
...
+		// Clear previous VolumeSource before setting new one to prevent collision
+		tokensVol.VolumeSource = core.VolumeSource{}
+		if pkcs11Config.Persistence != nil && pkcs11Config.Persistence.Name != "" {
+			tokensVol.PersistentVolumeClaim = &core.PersistentVolumeClaimVolumeSource{
+				ClaimName: pkcs11Config.Persistence.Name,
+			}
+		} else {
+			tokensVol.EmptyDir = &core.EmptyDirVolumeSource{}
+		}
+	}
```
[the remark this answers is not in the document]
sampras343: Fixed in 54c3f351. The `hsm-tokens` volume source is now always reconciled — not just when the volume is absent. When the volume already exists from a previous reconcile, the operator clears the `VolumeSource` and re-applies the current `persistence` setting (PVC or EmptyDir), then calls `ensureVolumeDefaultMode`. This ensures persistence changes take effect on subsequent reconciles without requiring a stack delete/recreate.

internal/controller/ctlog/actions/deployment.go
```diff
@@ -191,6 +216,178 @@ func (i deployAction) ensureDeployment(instance *rhtasv1.CTlog, sa string, label
...
+	// (TerminationMessagePath, TerminationMessagePolicy, ImagePullPolicy).
+	reconcileInitContainers(&template.Spec, instance.Spec.InitContainers, pkcs11Config.ModulePath) //nolint:actionlint // template.Spec is the pod template, not the CR spec
+
+	// Inject auth env vars (e.g. HSM_PIN) if configured
+	if instance.Spec.Signer.Auth != nil {
+		for _, env := range instance.Spec.Signer.Auth.Env {
+			e := kubernetes.FindEnvByNameOrCreate(container, env.Name)
+			e.Value = env.Value
+			e.ValueFrom = env.ValueFrom
+		}
```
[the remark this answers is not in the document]
sampras343: Fixed in 54c3f351. Replaced the manual `Auth.Env` loop with the shared `ensure.ContainerAuth(container, auth)` function, which handles both `Env` (environment variables) and `SecretMount` (projected secret volumes mounted at `/var/run/secrets/tas/auth`). This is the same function used by Rekor, TSA, and Trillian for auth injection. The nil guard (`if auth != nil`) ensures file-mode deployments are unaffected.

sampras343:
**Qodo findings #5 and #6 addressed in 3dba406c:**

**#5. CTLog auth misplaced in sample CR** — Fixed. The sample CR had `auth:` at `spec.ctlog.auth` (top-level) instead of `spec.ctlog.signer.auth` (inside CTlogSigner). Moved to correct location matching the API definition. The API server would have rejected the previous sample.

**#6. Config key not validated** — Fixed. Replaced `ExistsSecret` (checks Secret name only) with `GetSecretData` (reads the actual key from the Secret). If the Secret exists but the specified key is missing, the operator now reports a clear error: `configRef Secret "name" key "key" not accessible` instead of silently marking the config as resolved while the volume mount would be empty.

internal/controller/ctlog/actions/ensure_pkcs11_config.go
```diff
@@ -0,0 +1,180 @@
...
+	if state.FromInstance(instance, constants.ReadyCondition) < state.Creating {
+		return false
+	}
+	// Fire if PKCS11 condition is not yet set
+	cond := meta.FindStatusCondition(instance.Status.Conditions, PKCS11Condition)
+	if cond == nil {
+		return true
+	}
+	// Fire if CR generation changed (e.g., tokenLabel updated)
+	if cond.ObservedGeneration != instance.GetGeneration() {
```
osmman, line 44: **PKCS#11 "rotation" fires on any spec change, not just PKCS#11 field changes.** `CanHandle()` returns true whenever `ObservedGeneration != Generation` — true for *any* spec edit (resources, TLS, monitoring, etc.), not just a PKCS#11 field change. Combined with `Handle()` unconditionally calling `handleRotation()` once `PKCS11Condition` is `True`, any unrelated spec bump on a PKCS#11-mode CTlog regenerates the server config Secret (new `GenerateName`) and restarts the pod. Consider comparing the actual PKCS#11 fields (tokenLabel, modulePath, configRef, etc.) against a stored/observed value before triggering rotation. Same gap in `fulcio/actions/ensure_pkcs11_config.go`.

internal/controller/fulcio/actions/ensure_pkcs11_config.go
```diff
@@ -0,0 +1,109 @@
...
+		return false
+	}
+	if state.FromInstance(instance, constants.ReadyCondition) < state.Creating {
+		return false
+	}
+	cond := meta.FindStatusCondition(instance.Status.Conditions, PKCS11Condition)
+	if cond == nil {
+		return true
+	}
+	if cond.ObservedGeneration != instance.GetGeneration() {
```
osmman, line 41: Same gap as CTLog's `ensure_pkcs11_config.go`: this only checks `ObservedGeneration != Generation`, which fires on *any* spec change, not specifically a PKCS#11 field change. `Handle()` then calls `handleRotation()` whenever `PKCS11Condition` is already `True` — so any unrelated Fulcio spec edit regenerates the server config Secret and restarts the pod.

internal/controller/fulcio/actions/deployment.go
```diff
@@ -120,187 +120,497 @@ func (i deployAction) ensureDeployment(instance *rhtasv1.Fulcio, sa string, labe
...
+		// preventing the transition to "Ready".
+		ensureVolumeDefaultMode(v)
+	}
+
+	// Operator-managed volumes: hsm-tokens and hsm-lib default to EmptyDir but
+	// can be overridden by user-defined volumes (e.g., PVC for token persistence,
+	// or omitted entirely if the custom Fulcio image bundles the vendor SDK).
+	if !hasVolume(&template.Spec, HSMTokensVolumeName) {
+		hsmTokensVol := kubernetes.FindVolumeByNameOrCreate(&template.Spec, HSMTokensVolumeName)
+		hsmTokensVol.EmptyDir = &core.EmptyDirVolumeSource{}
```
osmman, line 415: **Drift vs. CTLog's equivalent path.** Unlike `ctlog/actions/deployment.go` (which resets `tokensVol.VolumeSource = core.VolumeSource{}` before assigning EmptyDir/PVC), this sets `EmptyDir`/`PersistentVolumeClaim` directly without clearing the existing `VolumeSource` first — if a volume switches source type across reconciles, stale fields could linger. Also: `hasVolume`/`hasMountPath`/`ensureVolumeDefaultMode` (lines 538-568 below) are copy-pasted verbatim from the CTLog package; consider moving them into `internal/utils/kubernetes` next to `FindVolumeByNameOrCreate` so both stay in sync.

internal/controller/ctlog/actions/ensure_pkcs11_config.go
```diff
@@ -0,0 +1,180 @@
...
+		return e.handleRotation(ctx, instance)
+	}
+
+	p := instance.Spec.Signer.PKCS11
+	if p == nil {
+		return e.Error(ctx, fmt.Errorf("pkcs11 config is nil"), instance)
+	}
+
+	// Validate required refs
+	if p.PinSecretRef == nil {
```
osmman, line 61: The `PinSecretRef` nil-check+error (61-77) and `PublicKeyRef` nil-check+error (79-95) blocks are near-identical, as are the two `GetSecretData` validation blocks below (98-134). Worth collapsing into a small `validateSecretRef(...)`-style helper to cut the duplication.

internal/controller/fulcio/actions/deployment.go
```diff
@@ -120,187 +120,497 @@ func (i deployAction) ensureDeployment(instance *rhtasv1.Fulcio, sa string, labe
...
-			"--grpc-port=5554",
-			fmt.Sprintf("--log_type=%s", utils.GetOrDefault(instance.GetAnnotations(), annotations.LogType, string(constants.Prod))),
-			"--ca=fileca",
-			"--fileca-key",
-			"/var/run/fulcio-secrets/key.pem",
-			"--fileca-cert",
-			"/var/run/fulcio-secrets/cert.pem",
-			fmt.Sprintf("--ct-log-url=%s", ctlogUrl),
-		}
+func (i deployAction) ensureFileCADeployment(instance *rhtasv1.Fulcio, sa string, labels map[string]string, dp *v1.Deployment) error {
```
osmman, line 134: `ensureFileCADeployment` and `ensurePKCS11Deployment` (line 291) both re-derive the same `Replicas`/`Selector`/template labels/`ServiceAccountName`/`AutomountServiceAccountToken`, http/grpc/monitoring port setup, and the `fulcio-config`/`oidc-info` volume+mount wiring. Consider extracting the shared pod scaffolding into a helper (similar to the existing `setProbes`) so the two paths can't silently diverge.

internal/utils/kubernetes/config_map.go
```diff
@@ -45,6 +46,29 @@ func ListConfigMaps(ctx context.Context, c client.Client, namespace string, labe
 	return list, nil
 }
 
+func FindConfigMap(ctx context.Context, c client.Client, namespace string, label string) (*metav1.PartialObjectMetadata, error) {
```
osmman, line 49: `FindConfigMap` has no callers anywhere in the tree. Remove it if it's leftover from an earlier iteration, or note the intended future use if it's meant to stay.

api/v1/fulcio_types.go
```diff
@@ -7,11 +7,73 @@
...
 
-const FulcioSignerTypeFile = "file"
+const (
+	FulcioSignerTypeFile   = "file"
+	FulcioSignerTypePKCS11 = "pkcs11"
+)
+
+// PKCS11InitContainerSpec defines a curated subset of corev1.Container for PKCS#11 init containers.
+// These containers run before the main Fulcio server to perform vendor-specific HSM initialization.
+type PKCS11InitContainerSpec struct {
```
osmman, line 22: `PKCS11InitContainerSpec` is used identically by both `FulcioSpec.InitContainers` and `CTlogSpec.InitContainers` (and both `reconcileInitContainers` functions) — it's a generic curated init-container shape, not actually PKCS#11-specific. Suggest renaming to `InitContainerSpec` and moving it to `common.go` alongside the other shared types (`PodRequirements`, `ServiceAccountConfig`, etc.), since defining a shared type inside `fulcio_types.go` is misleading for CTLog callers.

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -243,11 +248,75 @@ func TestFulcioConversion(t *testing.T) {
...
+			// because deeply nested corev1 types have nil-vs-empty issues.
+			c.FillNoCustom(&s.ConfigRef)
+			c.FillNoCustom(&s.KeyConfig)
+		},
+	}
+}
+
+// v1OnlyFieldsFuzzerFuncs nils out v1-only fields that have no v1alpha1 equivalent
+// and therefore cannot survive the hub→spoke→hub roundtrip.
+func v1OnlyFieldsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
```
osmman, line 295: **This fuzzer exclusion looks like it's hiding real conversion-logic coverage, not just working around a test artifact.** Project convention is that every new v1 *spec* field is expected to survive the hub↔spoke roundtrip — the accepted exceptions are mainly status fields that were intentionally removed in the v1 API redesign, which don't apply here. The comment on this function says `InitContainers`/`Volumes`/`VolumeMounts`/`Auth` "have no v1alpha1 equivalent and therefore cannot survive the hub→spoke→hub roundtrip" — but that's not accurate: `fulcio_conversion.go` (lines 153-156) and `ctlog_conversion.go` (lines 105-108) explicitly restore all four of these fields from the `MarshalData`/`UnmarshalData` annotation stash, the exact same mechanism used for `Signer.PKCS11`. By nil-ing them out here instead of giving them the same "constrain to round-trip-safe fields" treatment as `fulcioPKCS11FuzzerFuncs`/`ctlogPKCS11FuzzerFuncs`, this test provides zero fuzz coverage of that restore code path — a future refactor could break or drop one of those four assignments and this suite would still pass trivially (nil in, nil out). Please fix the fuzzer to actually exercise this restore logic rather than bypassing it, since these are new spec fields this PR added, not a removed-status-field exception.

config/manifests/bases/rhtas-operator.clusterserviceversion.yaml
```diff
@@ -61,6 +61,41 @@ spec:
       kind: Tuf
       name: tufs.rhtas.redhat.com
       version: v1
+    - description: CTlog is the Schema for the ctlogs API
```
osmman, line 64: Why does this PR add a second, duplicate `owned` CRD entry per kind at `version: v1alpha1` (CTlog, Fulcio, Rekor, Securesign, TimestampAuthority, Trillian, Tuf — lines 64-98)? On `main` each kind is only listed once, at `version: v1`. This looks unrelated to PKCS#11/HSM support — possibly generated-file drift from running `make manifests`/`make bundle` with a different local toolchain version than CI uses. Was this intentional, or should it be dropped?

internal/controller/fulcio/actions/deployment.go
```diff
@@ -120,187 +120,497 @@ func (i deployAction) ensureDeployment(instance *rhtasv1.Fulcio, sa string, labe
...
+		MatchLabels: labels,
+	}
+
+	template := &spec.Template
+	template.Labels = labels
+	template.Spec.ServiceAccountName = sa
+	template.Spec.AutomountServiceAccountToken = &[]bool{true}[0]
+
+	// Clean up PKCS#11 resources that may remain from a previous pkcs11→file mode switch
+	template.Spec.InitContainers = nil
```
osmman, line 169: **InitContainers/Volumes/VolumeMounts/Auth are silently ignored outside PKCS#11 mode.** These were moved to `FulcioSpec`/`Signer` top level specifically because they're "pod-wide regardless of nesting, and not PKCS#11-specific" (per the original API-redesign feedback). But `reconcileInitContainers`, the `Spec.Volumes`/`Spec.VolumeMounts` loops, and the `ensure.ContainerAuth` call all live only inside `ensurePKCS11Deployment` (lines 337-398 below). This function (`ensureFileCADeployment`) never applies any of them — it only nils `InitContainers` here, unconditionally. A user setting `spec.fulcio.initContainers`/`volumes`/`volumeMounts` or `spec.fulcio.signer.auth` on a `type: file` CR gets silently ignored. These should be applied in `ensureDeployment` before branching on signer type, not only from the PKCS#11 path — the whole point of moving them out of the PKCS#11 config was that they're independent of signer type.

internal/controller/ctlog/actions/deployment.go
```diff
@@ -149,6 +163,17 @@ func (i deployAction) ensureDeployment(instance *rhtasv1.CTlog, sa string, label
 		volumeMount.MountPath = "/ctfe-keys"
 		volumeMount.ReadOnly = true
 
+		if isPKCS11 {
+			i.ensurePKCS11Deployment(instance, template, container)
+		} else {
+			// Clean up PKCS#11 resources that may remain from a previous pkcs11→file mode switch
+			template.Spec.InitContainers = nil
```
osmman, line 170: Same issue as Fulcio: `instance.Spec.InitContainers`/`Volumes`/`VolumeMounts`/`Signer.Auth` are only applied inside `ensurePKCS11Deployment` (lines 224-251). This `else` branch (file mode) just nils `InitContainers` and strips the HSM-named volumes/mounts — it never applies any user-defined `Volumes`/`VolumeMounts`/`Auth`. Since these fields are meant to be pod-wide, not PKCS#11-specific (per the original API-redesign feedback), they should be applied regardless of `signer.type`, not gated behind the `isPKCS11` branch.

internal/controller/ctlog/actions/deployment.go
```diff
@@ -191,6 +216,190 @@ func (i deployAction) ensureDeployment(instance *rhtasv1.CTlog, sa string, label
 	}
 }
 
+func (i deployAction) ensurePKCS11Deployment(instance *rhtasv1.CTlog, template *core.PodTemplateSpec, container *core.Container) {
```
osmman, line 219: I'd prefer removing the `//nolint:actionlint` suppressions below rather than keeping them. `InitContainers`/`Volumes`/`VolumeMounts`/`Auth` are cross-cutting pod concerns, not CTLog-specific — `internal/utils/kubernetes/ensure` already exists for exactly this (see `ensure.ContainerAuth`, used two lines below). Adding `ensure.InitContainers(...)`/`ensure.Volumes(...)`/`ensure.VolumeMounts(...)` there, alongside `ContainerAuth`, and calling them from both Fulcio and CTLog would remove the duplication flagged above *and* the nolint noise as a side effect (free functions in `ensure` aren't methods on an action-type receiver, so the analyzer never looks at them).

internal/controller/ctlog/actions/generate_signer.go
```diff
@@ -30,6 +30,10 @@ func NewGenerateSignerAction() action.Action[*rhtasv1.CTlog] {
 			ResolveRef:   resolveRef,
 			GenerateData: generateData,
 			AlignStatus:  alignStatus,
+			IsEnabled: func(i *rhtasv1.CTlog) bool {
+				// PKCS#11 mode manages keys on the HSM — no file-based signer secret needed.
+				return i.Spec.Signer.Type != rhtasv1.CTlogSignerTypePKCS11
```
osmman, line 35: **Negative check on signer type is fragile for future signer types.** `IsEnabled` returns `true` whenever `Type != CTlogSignerTypePKCS11` — today that's equivalent to `Type == CTlogSignerTypeFile` since those are the only two types, but it silently breaks the moment a third type (e.g. `kms`) is added: this would then generate/manage a file-based private key secret for a KMS-backed signer too, which is wrong. Please check `Type == rhtasv1.CTlogSignerTypeFile` instead, so unknown/future types fail closed (disabled) rather than fail open (enabled). Fulcio's `generate_signer.go` avoids this by not needing an `IsEnabled` gate at all and checking `== FulcioSignerTypePKCS11` positively wherever it branches — worth mirroring that style here.

internal/controller/ctlog/actions/ensure_pkcs11_config.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/securesign/operator/internal/state"
+	"github.com/securesign/operator/internal/utils/kubernetes"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch
+
+func NewEnsurePKCS11ConfigAction() action.Action[*rhtasv1.CTlog] {
```
osmman, line 19: **Missing test coverage: no unit tests for CTLog's PKCS#11 support, and no integration/e2e coverage for the feature at all.** This PR adds unit tests for Fulcio's PKCS#11 pieces (`fulcio/actions/ensure_pkcs11_config_test.go`, `fulcio/actions/fulcio_deployment_test.go`) but this file and `ctlog/actions/deployment.go` have none. More importantly, every test added here (including the Fulcio ones) exercises a single action in isolation — nothing runs the full reconcile flow through the existing envtest infra (`ctlog_controller_test.go`/`fulcio_controller_test.go`, `testonly.ControllerSuite`), and there's no `test/e2e` coverage either. This is a multi-action state machine (config validation → server config generation → deployment, with rotation/drift handling) — at least one controller-level test that creates a `signer.type: pkcs11` CR and asserts it reaches `Ready` with the expected init containers/volumes/mounts would exercise the interactions between actions that unit tests can't, and would have caught the rotation false-trigger and the ignored-outside-PKCS11-mode bugs flagged elsewhere in this review.

api/v1/ctlog_types.go
```diff
@@ -108,6 +131,39 @@ type CTlogFile struct {
...
+	// --pkcs11_module_path to ct_server.
+	//+required
+	//+kubebuilder:validation:MinLength=1
+	//+kubebuilder:validation:Pattern=`^/[a-zA-Z0-9._/-]+\.so[.0-9]*$`
+	ModulePath string `json:"modulePath"`
+
+	// Persistent storage for HSM tokens (key survives pod restarts).
+	// When nil, an emptyDir is used (key is regenerated on pod restart).
+	//+optional
+	Persistence *Pvc `json:"persistence,omitempty"`
```
osmman, line 164: **`Persistence *Pvc` doesn't do what the shared `Pvc` type implies, and duplicates the generic `Volumes`/`VolumeMounts` mechanism you already added.** Unlike Rekor/TUF/Trillian's own `pvc.go` actions (which actually provision a `PersistentVolumeClaim` using `Size`/`StorageClass`/`Retain`), there's no equivalent action for CTLog — `internal/controller/ctlog/actions/deployment.go` only ever reads `Persistence.Name` and plugs it into `PersistentVolumeClaimVolumeSource{ClaimName: ...}` (lines 263-265, 274-276). `Size`, `StorageClass`, `Retain` (and their CEL immutability rules on `Pvc`) are dead for this field — a user setting them would reasonably expect the operator to provision storage, and it won't. Since this is just "reference an existing PVC by name," it seems fully replaceable by having the user define their own `hsm-tokens` volume via the generic `spec.ctlog.volumes`/`volumeMounts` you already introduced, with no dedicated `Persistence` field needed at all.

That also surfaces a real bug in the current precedence: the comment at `deployment.go:248-249` says user-defined volumes are processed first "so operator-managed volumes take precedence for reserved names," but both the `if !hasVolume(...)` and `else` branches (256-280) unconditionally do `tokensVol.VolumeSource = core.VolumeSource{}` and rebuild it from `Persistence`/`EmptyDir` — so a user-defined `hsm-tokens` volume (e.g. a different PVC, or even a Secret) gets silently discarded and replaced every reconcile, regardless of what they configured. The two branches are functionally identical, so the `hasVolume` check isn't actually gating anything today.

config/samples/rhtas_v1_securesign_pkcs11.yaml
```diff
@@ -0,0 +1,172 @@
+# SecureSign CR with dual PKCS#11 HSM-backed signing (Fulcio + CTLog)
+#
+# Replace all <placeholder> values with your environment-specific configuration.
+# This sample is vendor-agnostic. For SoftHSM dev/test or production HSM examples,
+# see docs/pkcs11-hsm-support.md.
```
osmman, line 5: This sample is 172 lines — by far the largest in `config/samples/` (every other one is 9-73 lines). Files here are meant to feed the CSV's `alm-examples` annotation (see `config/samples/kustomization.yaml`: "Append samples you want in your CSV to this file as resources"), which gets stored as CR data in etcd — we intentionally keep these small. This one isn't wired into `kustomization.yaml` yet, but it shouldn't live here at all: please move it under `docs/` and link it from documentation instead, matching the pattern of `fulcio-key-rotation.md`/`ctlog-key-rotation.md`/etc. Also, this comment already points at `docs/pkcs11-hsm-support.md`, but that file doesn't exist anywhere in the repo — please add it (or fix the reference) as part of the move.

osmman (changes requested):
## Summary

API-redesign feedback is fully addressed. Remaining issues inline — headlines:

**Blocking:**
- PKCS#11 "rotation" fires on *any* spec change, not just PKCS#11 fields — spurious config regen + pod restarts (Fulcio + CTLog).
- `InitContainers`/`Volumes`/`VolumeMounts`/`Auth` are silent no-ops outside `type: pkcs11`, despite being moved to spec top level to be signer-agnostic.
- `Persistence` on `CTlogPKCS11Config` doesn't provision a PVC and silently overwrites a user-defined `hsm-tokens` volume every reconcile — conflicts with the new `Volumes` mechanism this PR introduced.
- No CTLog unit tests, no controller/e2e coverage — would have caught the bugs above.

**Worth fixing:** fragile `!= pkcs11` check in `generate_signer.go`; duplicated/drifting helpers between Fulcio and CTLog; roundtrip fuzzer skips testing the restore path it already has.

## Recommendation: split this PR

Too large (~15k lines) to review the PKCS#11 logic as deeply as it needs. Please split into:
1. Generic `InitContainers`/`Volumes`/`VolumeMounts`/`Auth` extension, signer-agnostic.
2. PKCS#11 signer support on top of (1).

Would also make proper test coverage realistic for the PKCS#11 piece.

sampras343:
Comments addressed in https://github.com/securesign/secure-sign-operator/pull/2182 with updated schema and integration with v1 APIs

PR #2131, opened 2026-07-24
fix: handle KMS and Tink signers in TSA generate_signer action

JasonPowr:
## Summary
- Fix nil pointer panic when deploying TSA with KMS or Tink signer
- The `isEnabled` callback was returning false for KMS/Tink, skipping
  status population that the deployment action depends on
- Add `resolveRef` handling for KMS/Tink that validates
  CertificateChainRef and Tink KeysetRef secrets before marking the
  signer as resolved
- Add `TerminalError` guard for Tink configured without KeysetRef
- Regression from generic signer refactor (6764d473)

### DIFF
--- a/internal/controller/tsa/actions/resolve_kms_tink_signer.go
+++ b/internal/controller/tsa/actions/resolve_kms_tink_signer.go
@@ -1,0 +1,151 @@
+package actions
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	rhtasv1 "github.com/securesign/operator/api/v1"
+	"github.com/securesign/operator/internal/action"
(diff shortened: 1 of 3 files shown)


osmman (changes requested):
The generateSigner framework was designed to resolve or generate key material for file-based signers. For KMS and Tink the operator doesn't generate or manage any keys, so making generateSigner handle them stretches it beyond its responsibility. GenerateData and MutateSecret become dead code paths, and resolveRef grows to branch on three different signer types.

I'd prefer a separate lightweight action for KMS/Tink that validates user-provided secrets and populates Status.Signer. Cleaner separation of concerns and no dead code in generateSigner.

osmman (approved):

PR #2146, opened 2026-07-27
fix: fix camelCase typo

bouskaJ:


JasonPowr (approved):

kdacosta0:
@bouskaJ Should this json tag also be updated in [v1alpha](https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/rekor_types.go#L212) to match the v1 fix? Or is there a reason to keep them different?

bouskaJ:
> @bouskaJ Should this json tag also be updated in [v1alpha](https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/rekor_types.go#L212) to match the v1 fix? Or is there a reason to keep them different?

@kdacosta0 Just to answer your question - we can't touch v1alpha API (backward compatibility). I am fixing that on the new one only

PR #2149, opened 2026-07-27
feat(console): use securesign Rekor URL instead of the public Rekor instance

fghanmi:


internal/controller/console/actions/ui/deployment.go
```diff
@@ -117,6 +118,13 @@ func ensureUIDeployment(instance *rhtasv1.Console, labels map[string]string) fun
 		apiURL := kubernetes.FindEnvByNameOrCreate(container, "CONSOLE_API_URL")
 		apiURL.Value = fmt.Sprintf("%s://%s:%d", scheme, apiHost, actions.ApiPort)
 
+		rekorURL := kubernetes.FindEnvByNameOrCreate(container, "NEXT_PUBLIC_REKOR_DEFAULT_DOMAIN")
+		url, err := apis.ServiceAsUrl(&instance.Spec.UI.Rekor)
```
kdacosta0: What happens when `spec.ui.rekor` is not set by user? `ServiceAsUrl` on a zero value `RekorService{}` returns `""` with no error, which I assume means we'd set `NEXT_PUBLIC_REKOR_DEFAULT_DOMAIN=""` on the container

kdacosta0 (changes requested):
Could we add at least a unit test for this function covering the case where `spec.ui.rekor` is set and when ommited?

[ensureUIDeployment](https://github.com/securesign/secure-sign-operator/pull/2149/changes#diff-a782c43c943a6837de3e9047aef061e853d18a897c1b9ff9df1149ede96efab8R94)

internal/controller/console/actions/ui/deployment.go
fghanmi: if spec.ui.rekor is not set (ie: NEXT_PUBLIC_REKOR_DEFAULT_DOMAIN="" ), TAS UI will fallback to rekor public instance https://rekor.sigstore.dev (tested)

kdacosta0: Thanks for clarification @fghanmi 

Should we document this fallback behaviour somewhere? Field description in the go struct (for like `oc explain`). Or maybe RHTAS doc?

fghanmi: default value is set on the UI. So IMHO probably better to add the explanation in the docs.

kdacosta0: Created jira for tracking https://redhat.atlassian.net/browse/SECURESIGN-5065

osmman:
@fghanmi can you try to rebase pr agains main. It could resolve problems in failing tests

api/v1/console_types.go
```diff
@@ -38,6 +38,9 @@ type ConsoleUI struct {
 	PodRequirements `json:",inline"`
 	// Define whether you want to export service or not
 	Ingress Ingress `json:"ingress,omitempty"`
+	// Rekor service configuration
+	//+optional
+	Rekor RekorService `json:"rekor,omitempty"`
```
osmman: There is a new `v1.ServiceReference` object to handle dependencies between services. Can you please use it. @bouskaJ can provide more info/examples.

fghanmi: done, thanks!

internal/controller/console/actions/ui/deployment.go
```diff
@@ -117,6 +117,11 @@ func ensureUIDeployment(instance *rhtasv1.Console, labels map[string]string) fun
 		apiURL := kubernetes.FindEnvByNameOrCreate(container, "CONSOLE_API_URL")
 		apiURL.Value = fmt.Sprintf("%s://%s:%d", scheme, apiHost, actions.ApiPort)
 
+		if instance.Spec.UI.Rekor.URL != "" {
```
osmman: Both `spec.ui.rekor` and `spec.api.tuf` use `ServiceReference` which accepts `.ref` or `.url` (mutually exclusive). Only `.url` is read by the controllers — `.ref` is silently ignored:

- `internal/controller/console/actions/ui/deployment.go:120` — reads `.URL` only
- `internal/controller/console/actions/api/deployment.go:50` — reads `.URL` only

A user setting `.ref` gets no error and no effect. The controller must resolve `.ref` by looking up the referenced CR's service URL.

osmman: @bouskaJ do you have some existing example/helper function to resolve url from `ServiceReference.ref`? I expect that there will have to be some logic to wait untill these services has know their ingress url.

internal/controller/console/actions/ui/deployment.go
```diff
@@ -117,6 +117,11 @@ func ensureUIDeployment(instance *rhtasv1.Console, labels map[string]string) fun
...
+			rekorURL := kubernetes.FindEnvByNameOrCreate(container, "NEXT_PUBLIC_REKOR_DEFAULT_DOMAIN")
```
osmman: When `spec.ui.rekor.url` is set, `NEXT_PUBLIC_REKOR_DEFAULT_DOMAIN` is added to the container. When the URL is later cleared or removed, the `if` block is skipped and no cleanup runs — the stale env var persists on the existing Deployment.

**Example:** User sets `spec.ui.rekor.url: https://rekor.internal.corp` → env var is created. Later they remove the field to switch back to the default public Rekor. The operator skips the `if` block but the old env var stays, so the UI keeps pointing at `rekor.internal.corp` — which may no longer exist.

**Fix:** Add an `else` branch with `kubernetes.RemoveEnvVarByName(container, "NEXT_PUBLIC_REKOR_DEFAULT_DOMAIN")` — same pattern used for `SSL_CERT_DIR` and `SIGNER_PASSWORD` elsewhere in the codebase.

osmman (changes requested):

internal/controller/console/actions/ui/deployment.go
bouskaJ: yes, Does the Console supposed to consume public or internal URL? @fghanmi

fghanmi: Ideally, it should consumes public URL, but in "can" also use internal URL

fghanmi: which means, could we use similar resolver on both TUF and rekor, just wondering
```
func init() {
	serviceresolver.Register(
		func(obj *rhtasv1.Tuf) (string, error) {
			return obj.GetServiceURL(), nil
		})
}
```

bouskaJ: Infra for public services is not yet merged (pending for @osmman re-review on https://github.com/securesign/secure-sign-operator/pull/2126)

fghanmi: I added this commit https://github.com/securesign/secure-sign-operator/pull/2149/changes/11d5d6a173a7690e5f982217bcc431c356cac4b6

Just an idea. It was tested fine.

bouskaJ: We may use public address - the resolver for public address will be present once #2126 is merged.

bouskaJ: @fghanmi you may rebase and use it like this https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/tuf/utils/service_urls.go#L59

osmman (approved):

PR #2154, opened 2026-07-28
feat: Add HAProxy rate-limiting support for Ingress resources (SECURESIGN-3115)

jourdee-lab:
## Summary

OpenShift Routes auto-generated from Ingress resources had no rate-limiting, leaving Fulcio, Rekor, TSA, and Console endpoints exposed to traffic spikes. This PR makes the operator apply HAProxy throttling annotations on OpenShift by default, exposes ingress customisation through the existing generic "ingress.annotations field", and fixes a bug where annotations removed from the CR were never removed from the live Ingress/Route.

## Initial design
- Added a dedicated Throttling CRD field (IngressThrottling struct) to the Fulcio, Rekor, TSA, and Console specs, with its own enabled/concurrentTCP/rateHTTP/rateTCP sub-fields
- Enabled by default with fixed per-component values; users opted out via throttling.enabled: false or overrode individual values on the struct
- Annotations were derived straight from the current CR spec on each reconcile

## What changed (feedback received)
- Vendor lock-in: removed the HAProxy-specific Throttling CRD struct; reused the existing generic ingress.annotations field instead. Operator layers HAProxy defaults onto it internally, OpenShift-only.
- Disable/override mechanism: changed from throttling.enabled: false to setting haproxy.router.openshift.io/rate-limit-connections: "false" (or any HAProxy key) directly in ingress.annotations.
- Defaults too restrictive: raised Fulcio/TSA to rateHTTP=100/rateTCP=100; Rekor to 200/200/concurrentTCP=200; Console kept at 50/100/100.
- Stale annotations bug: added EnsureIngressAnnotations, which tracks previously-applied keys in a new rhtas.redhat.com/managed-ingress-annotation-keys annotation so removed CR annotations are actually deleted from the Ingress/Route. Wired into Fulcio, Rekor, TSA, Console, TUF.
- Duplicated test constants (qodo bot): tests now use the shared internal/annotations constants; added parameterized test harnesses (internal/testing/action/haproxy_throttling.go, ingress_annotations.go) covering all five controllers.


osmman (changes requested):
## Design concern: HAProxy-specific API surface in the CRD

The `IngressThrottling` type is a 1:1 mapping of `haproxy.router.openshift.io/rate-limit-connections.*` annotations, which ties the CRD permanently to one specific ingress controller implementation. On OpenShift clusters running a different ingress controller (nginx, traefik, etc.) these fields are silently ignored — no error, no warning, no condition — and on non-OpenShift deployments (EKS, GKE, vanilla k8s) the entire struct is completely inert. Since CRD fields in v1 are permanent, we'd be locked into this HAProxy-specific surface even if we later need to support other rate-limiting backends.

### Proposed alternative

Add an `Annotations` field to the existing `Ingress` type — the same pattern already used for `Labels`:

```go
type Ingress struct {
    Enabled     *bool             `json:"enabled,omitempty"`
    Host        string            `json:"host,omitempty"`
    Labels      map[string]string `json:"labels,omitempty"`
    Annotations map[string]string `json:"annotations,omitempty"`  // new
}
```

This gives users full control over any ingress/route annotations their environment needs — HAProxy throttling, nginx rate-limiting, custom headers — without baking vendor-specific semantics into the API. For our managed default deployments, the operator would internally apply sensible HAProxy throttling annotations when running on OpenShift (just like we already set TLS annotations), without requiring any CRD configuration. Users who want to override or disable defaults can do so via the `annotations` map.

This keeps the CRD vendor-neutral and portable, still gives us throttling-by-default on OCP, lets users on any platform configure their own ingress annotations, and doesn't pollute the API surface with implementation-specific fields.

### Rate-limit values concern

Where do the proposed default values come from? They look too restrictive for API endpoints. HAProxy's `rate-http` uses a hardcoded **10-second sliding window**, so `rateHTTP=50` means only **5 req/sec per source IP** — automated signing clients, CI pipelines, or frontends making parallel requests will hit this very quickly. Similarly `rateTCP=100` uses a **3-second window** (~33 new connections/sec), which is tight. Only `concurrentTCP=100` looks reasonable for most workloads.

These are per-source-IP limits enforced by HAProxy's stick-table, not global. There's also a [known issue](https://github.com/openshift/origin/issues/18395) where TCP-based limits may not be reliably enforced for parallel connections, so these shouldn't be treated as a hard security boundary. I'd suggest bumping `rateHTTP` to at least 100-200 for API endpoints (Fulcio, Rekor, TSA) and being more conservative only for the Console UI.

internal/annotations/annotations.go
```diff
@@ -125,6 +125,18 @@
 //	  name: example
 //	  annotations:
 //	    rhtas.redhat.com/log-type: "dev"
+//
+// # Annotation: rhtas.redhat.com/managed-ingress-annotation-keys
```
bouskaJ: The `ManagedIngressAnnotationKeys` create a split-view. Status field is designed to store the actual state.

bouskaJ (changes requested):
Nice work on the HAProxy throttling feature! A few architectural suggestions on the annotation management approach:

  1. Labels have the same removal problem — Ingress.Labels from the CR spec has the same dynamic-set issue. If you're solving annotation tracking here, please address labels too.
  2. ManagedIngressAnnotationKeys creates a split-view — Storing operator state as a meta-annotation on the child object mixes concerns. The CR's .status is designed to hold the operator's observed/applied
  state. If tracking is needed, that's where it belongs.
  3. Consider SSA instead of hand-rolled tracking — Server-Side Apply is designed exactly for this problem. The API server tracks field ownership per-key via managedFields, so removed keys are cleaned up
  automatically. Our CreateOrUpdate doesn't use SSA, but you can apply user annotations as a separate SSA Patch after it — as long as the ensure chain doesn't touch user annotations, Update only claims ownership
  of changed fields, so SSA ownership is preserved.
 Something like:
```
  // After CreateOrUpdate (handles spec, TLS, HAProxy):
  patch := &networkingv1.Ingress{
      TypeMeta:   metav1.TypeMeta{APIVersion: "networking.k8s.io/v1", Kind: "Ingress"},
      ObjectMeta: metav1.ObjectMeta{
          Name: name, Namespace: ns,
          Annotations: userAnnotations,
      },
  }
  client.Patch(ctx, patch, client.Apply,
      client.FieldOwner("securesign-user-annotations"),
      client.ForceOwnership)
  ```
No tracking annotation, no exclude lists, no ordering dependencies between ensure functions.

File: api/v1alpha1/conversion_roundtrip_test.go [hunk 2]
```diff
@@ -50,6 +50,10 @@ func enabledFieldsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
 			if s.Enabled == nil {
 				s.Enabled = ptr.To(c.Bool())
 			}
+			// empty map doesn't survive JSON omitempty roundtrip
+			if len(s.Annotations) == 0 {
```
osmman, line 54: These sanitisation of outputs should be handled `HubAfterMutation` or `SpokeAfterMutation` rather than  `FuzzerFunc`. 

Currently we have example in `TestRekorConversion` where we stripped migration annotation.

osmman:
@jourdee-lab CTlog has been extended to allow Ingress configuration can you please extend your PR to handle it.

https://github.com/securesign/secure-sign-operator/pull/2230

internal/utils/kubernetes/ingress.go
```diff
@@ -67,3 +70,112 @@ func EnsureIngressTLS() func(ingress *networkingv1.Ingress) error {
...
+//
+// This alone is not enough to get user labels onto an OpenShift Route: the
+// route-controller-manager snapshots Ingress labels onto the auto-created
+// Route only once, at Route creation time, and does not re-sync them on later
+// Ingress updates. Callers must also merge user labels into the Ingress
+// object as part of the same CreateOrUpdate call that first gives it a spec
+// (see the ensure.Labels call in each ingress action) so they are already
+// present when the Route gets created; this SSA call remains responsible for
+// cleaning up labels/annotations the user later removes from the CR.
+func ApplyIngressUserMetadata(ctx context.Context, cli client.Client, name, namespace string, userAnnotations, userLabels map[string]string) error {
```
osmman, line 160: This mutates the Ingress outside the `CreateOrUpdate` loop, so it can't respect the pause guard.

Fix: don't call this separately. Move it into the `CreateOrUpdate` mutate chain so it goes through the same pause check as everything else.

internal/utils/kubernetes/ingress.go
```diff
@@ -67,3 +70,112 @@ func EnsureIngressTLS() func(ingress *networkingv1.Ingress) error {
...
+const haproxyRateLimitConnectionsDisabledValue = "false"
+
+// EnsureIngressHAProxyThrottling sets OCP-native HAProxy rate-limiting and
+// connection-limit annotations on the Ingress, propagated to the auto-created Route.
+// Component defaults are always applied. If the user sets
+// "haproxy.router.openshift.io/rate-limit-connections" to "false" in their
+// Ingress.Annotations, all HAProxy rate-limiting annotations are removed instead.
+// User-provided annotation overrides are applied separately via SSA
+// (ApplyIngressUserMetadata) after CreateOrUpdate.
+func EnsureIngressHAProxyThrottling(userAnnotations map[string]string, defaults IngressThrottlingDefaults) func(ingress *networkingv1.Ingress) error {
```
osmman, line 102: I unable to find usage of it outside of tests

internal/testing/action/haproxy_throttling.go
```diff
@@ -0,0 +1,191 @@
...
+	v1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	haproxyRateLimitConnections   = "haproxy.router.openshift.io/rate-limit-connections"
+	haproxyRateLimitConcurrentTCP = "haproxy.router.openshift.io/rate-limit-connections.concurrent-tcp"
+	haproxyRateLimitRateHTTP      = "haproxy.router.openshift.io/rate-limit-connections.rate-http"
+	haproxyRateLimitRateTCP       = "haproxy.router.openshift.io/rate-limit-connections.rate-tcp"
```
osmman, line 20: These 4 keys duplicate the constants in `internal/annotations`. If one ever changes there, this test won't catch it.

Fix: import and use the `internal/annotations` constants instead of redeclaring them.

osmman (changes requested):

osmman:
@jourdee-lab Support for Ingress has been added to CTlog in `main` branch.

PR #2161, opened 2026-07-29
docs: add v1 Securesign sample demonstrating imagePullSecrets

anithapriyanatarajan:
This pull request adds a sample `Securesign` Custom Resource (CR) for the `v1` API and updates the vanilla-Kubernetes install docs to use it, demonstrating the per-component `imagePullSecrets` field that is only available on `v1`.

## Changes

- **New v1 sample CR:** added `config/samples/rhtas_v1_securesign.yaml`, a `rhtas.redhat.com/v1` sample mirroring `rhtas_v1alpha1_securesign.yaml`, with `imagePullSecrets` configured on every component.
- **docs/kubernetes.md:** the EKS install walkthrough now applies `rhtas_v1_securesign.yaml` instead of the `v1alpha1` sample, with a note explaining the `v1` sample demonstrates `imagePullSecrets` and that the `v1alpha1` sample remains available for backward compatibility (it cannot express `imagePullSecrets`).

## Validation

The initial version of the sample had two fields that don't exist on the `v1` API and would fail CRD validation:
- `fulcio.certificate` → corrected to `fulcio.signer.certificateChain` (`FulcioSpec.Signer` is required in `v1`).
- `ctlog.signer` was missing (`CTlogSpec.Signer` is required in `v1` but doesn't exist in `v1alpha1`, so it's easy to miss when mirroring the old sample).

Both were caught and fixed, then verified with a real `kubectl apply --server-side --dry-run=server` against the `v1` CRD, which now succeeds with no validation errors.

Assisted-by: Co-pilot(Claude Sonnet 5)

### DIFF
--- a/config/samples/rhtas_v1_securesign.yaml
+++ b/config/samples/rhtas_v1_securesign.yaml
@@ -1,0 +1,108 @@
+# Securesign sample on the v1 API (the CRD storage version).
+#
+# This mirrors rhtas_v1alpha1_securesign.yaml but uses apiVersion rhtas.redhat.com/v1
+# and additionally demonstrates per-component `imagePullSecrets`, which is only
+# available on the v1 API. Follow docs/kubernetes.md ("Image Registry") to create
+# the referenced `redhat-registry` pull secret before applying this CR, e.g.:
+#
+#   kubectl create secret docker-registry redhat-registry \
+#     --docker-server=registry.redhat.io \
+#     --docker-username=<your-username-or-sa-name> \
+#     --docker-password=<your-password-or-token> \
+#     -n <tas-namespace>
+#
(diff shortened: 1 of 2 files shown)


osmman:
@bouskaJ what is our plan with samples and updating docs for v1 api? Will we do it in one batch when most of v1 changes are done or we should them updated as changes are made?

bouskaJ:
> @bouskaJ what is our plan with samples and updating docs for v1 api? Will we do it in one batch when most of v1 changes are done or we should them updated as changes are made?

yes, I am still waiting with doc/samples update for (at least mostly) finished the v1 API. It does not make sense to do it right now and change with every new change IMHO.

anithapriyanatarajan:
Would close this PR and re-open later once V1 changes are all complete. Thank you

PR #2170, opened 2026-07-29
feat: remove RekorSearchUI and migrate to Console CR

osmman:
## Why

The `RekorSearchUI` component was embedded directly in the Rekor CR spec (v1alpha1 only). With the introduction of the standalone `Console` CR, SearchUI functionality is superseded. Existing v1alpha1 users need a migration path that preserves their SearchUI configuration as a Console CR, and orphaned SearchUI resources must be cleaned up.

## What

**Migration path** — a shared `internal/migration` package provides annotation-based helpers (`Set`, `Read`, `Pop`, `Propagate`, `Remove`) for preserving v1alpha1-only fields through the conversion webhook. The v1alpha1→v1 converter (`ConvertTo`) serializes `RekorSearchUI` into a migration annotation; the Rekor controller's new `migrationAction` reads it, creates a `Console` CR when `Enabled=true`, and removes the annotation.

**Cleanup** — a new `cleanupAction` deletes orphaned SearchUI resources (Deployment, Service, Ingress, ServiceAccount, RoleBinding) owned by the Rekor CR and removes the stale `UiAvailable` status condition. Transient API errors cause a requeue rather than best-effort continuation.

**v1 API removal** — `RekorSearchUI` field, image constant (`RekorSearchUi`), kustomize replacement block, and all SearchUI controller actions are removed from the v1 API surface.

**Key design decisions:**
- Converter remains unconditional (`migration.Set` always fires) — it faithfully preserves v1alpha1 data without content-aware logic; the migration action decides behavior based on `Enabled`
- `ensure_rekor` uses JSON Patch to surgically remove the migration annotation from Securesign, avoiding full-object update conflicts
- Migration action uses `kubernetes.CreateOrUpdate` for Console CR creation, which handles retry on Conflict/AlreadyExists
- Console CR is created as a top-level resource (no owner reference) since it has its own independent controller

**Test coverage:**
- Unit tests for migration action (5 cases) and cleanup action (5 cases)
- Integration tests via envtest for Console CR creation and orphaned resource cleanup
- Conversion roundtrip fuzz tests updated with `HubAfterMutation`
- e2e upgrade test verifies Console CR migration and old resource removal

## Breaking changes

The `spec.rekorSearchUI` field is removed from the v1 `Rekor` CRD. Users on v1alpha1 with SearchUI enabled will have a `Console` CR automatically created on upgrade. Users on v1 are unaffected — SearchUI was never part of the v1 API.

Refs: SECURESIGN-5066


kdacosta0:
Will need to do org wide cleanup of Rekor Search UI for >= 1.5.0 (e.g. sigstore-e2e, pipelines, quickstarts?)
Probably should add Jira tasks so we track it

osmman:
> Will need to do org wide cleanup of Rekor Search UI for >= 1.5.0 (e.g. sigstore-e2e, pipelines, quickstarts?) Probably should add Jira tasks so we track it

Yes I will create a jira to request change in sigstore-e2e for Console. I prepared https://github.com/securesign/pipelines/pull/569 to exclude these tests for deployments without Search UI.

osmman:
Depends on https://github.com/securesign/secure-sign-operator/pull/2149 to correctly configure Rekor ServiceReference during migration

config/default/images.env
fghanmi, line 46: not related to this PR changes, but could you please update console (outdated) images ? 
```
RELATED_IMAGE_CONSOLE_UI=registry.redhat.io/rhtas/rhtas-console-ui-rhel9@sha256:a31b0cb34323e6d12a23b01e6bdbda52f434682d81fd3fead4a94ecae87eaea2

RELATED_IMAGE_CONSOLE_API=registry.redhat.io/rhtas/rhtas-console-rhel9@sha256:c0716b6dfc61064e60c6292bee9d34d5c7eeeb2253b6f2957d2d443b37c33a20
```

internal/controller/rekor/actions/ui/migration.go
```diff
@@ -0,0 +1,102 @@
...
+					Host:    searchUI.Host,
+					Labels:  searchUI.RouteSelectorLabels,
+				},
+				Rekor: rhtasv1.ServiceReference{
+					Ref: &rhtasv1.ServiceReferenceRef{
+						Name:      instance.Name,
+						Namespace: instance.Namespace,
+					},
+				},
+			}
```
fghanmi, line 82: could you please add TUF ref here as well ? 
```
object.Spec.Api.Tuf = rhtasv1.ServiceReference{
				Ref: tufRef,
}
```
It's always better to have the TUF public public URL (if possible) instead of the internal one.

fghanmi (changes requested):

config/default/images.env
osmman: Cure, currently non existing Console UI image is cause of failure. I am currently working on it to fix it via [Konflux-nudge](https://github.com/securesign/pipelines/pull/575) to have correct long live image.

internal/controller/console/actions/api/deployment.go
```diff
@@ -48,14 +48,9 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1.Console) *ac
...
 
-	var tufURL string
-	if instance.Spec.Api.Tuf.URL != "" || instance.Spec.Api.Tuf.Ref != nil {
-		tufURL, err = utils.ResolveExternalServiceUrl(ctx, i.Client, instance.Spec.Api.Tuf, instance.Namespace, &rhtasv1.Tuf{})
-		if err != nil {
-			return i.Error(ctx, fmt.Errorf("error resolving TUF URL: %w", err), instance)
-		}
-	} else {
-		tufURL = fmt.Sprintf("http://tuf.%s.svc", instance.Namespace)
+	tufURL, err := utils.ResolveInternalServiceUrl(ctx, i.Client, instance.Spec.Api.Tuf, instance.Namespace, &rhtasv1.Tuf{})
```
fghanmi: Is there any reason for this change ? 
Same as rekor url [here](https://github.com/securesign/secure-sign-operator/blob/69a62c707ff2d6e2c9f13bb18126beb3c45ffdee/internal/controller/console/actions/ui/deployment.go#L52), we prefer that we get TUF public url (if not, fallback to local url)

config/default/images.env
osmman: updated images manually

internal/controller/console/actions/api/deployment.go
osmman: Console API is an in-cluster service, so using the internal service URL is simpler than routing through ingress. I switched it to use `ResolveInternalServiceUrl` with a new `serviceresolver` registration for TUF (same pattern as Trillian), which also handles user-configured URL/Ref overrides.

Are there plans for the Console dashboard to show the TUF URL to users? If so, I'll change it back to use the ingress endpoint.

fghanmi: > Are there plans for the Console dashboard to show the TUF URL to users?

Yes, this is what we do currently and this is why we prefer the route URL. 
Example:
image

osmman: Thank you, I will use External url by default.

internal/controller/rekor/actions/ui/migration.go
osmman: I let to resolved it by autodiscovery (ResolveExternalServiceUrl) in Console controller. Rekor have not got any usefull data from which I can extract value.

internal/controller/console/actions/api/deployment.go
```diff
@@ -48,14 +48,9 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1.Console) *ac
...
 	)
 
-	var tufURL string
-	if instance.Spec.Api.Tuf.URL != "" || instance.Spec.Api.Tuf.Ref != nil {
-		tufURL, err = utils.ResolveExternalServiceUrl(ctx, i.Client, instance.Spec.Api.Tuf, instance.Namespace, &rhtasv1.Tuf{})
-		if err != nil {
-			return i.Error(ctx, fmt.Errorf("error resolving TUF URL: %w", err), instance)
-		}
-	} else {
-		tufURL = fmt.Sprintf("http://tuf.%s.svc", instance.Namespace)
```
fghanmi, line 58: without this, when instance.Spec.Api.Tuf.URL and instance.Spec.Api.Tuf.Ref are nil, tufURL would be empty string, right ?

fghanmi: please ignore.

internal/controller/rekor/actions/ui/migration.go
osmman: @fghanmi is TUF url mandatory for API server or optional. Because by enabling autodiscovery and removing your old fallback I make it mandatory.

fghanmi (approved):

PR #2173, opened 2026-07-30
refactor: improve fuzzers

bouskaJ:


api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -73,194 +101,375 @@ func enabledFieldsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
...
+			c.FillNoCustom(s)
+			s.Address = randHttpUrl(c, false)
+			// port is always set (defaulter)
+			s.Port = ptr.To(int32(c.Intn(65534) + 1))
+		},
+	}
+}
+
+// ctlogSignerFuzzerFuncs constrains the CTlog v1 signer to only fill fields that
+// survive roundtrip — Type has no v1alpha1 equivalent and is always "file" today,
```
osmman: s.Type should not have custom fuzzer that should be lossless conversion bahave same like Fulcio

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -73,194 +101,375 @@ func enabledFieldsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
...
+				s.File = &rhtasv1.CTlogFile{}
 			case 1:
-				s.Trillian = rhtasv1.ServiceReference{Ref: &rhtasv1.ServiceReferenceRef{}}
-				c.FillNoCustom(s.Trillian.Ref)
+				// Populated File
+				s.File = &rhtasv1.CTlogFile{}
+				c.FillNoCustom(&s.File.PrivateKeyRef)
+				c.FillNoCustom(&s.File.PrivateKeyPasswordRef) //nolint:staticcheck
+				c.FillNoCustom(&s.File.PublicKeyRef)
+				if s.File.PrivateKeyRef == nil && s.File.PrivateKeyPasswordRef == nil && s.File.PublicKeyRef == nil { //nolint:staticcheck
```
osmman: these fuzzer tests tests that any data are not lost during round trip we do not care about validity of data that is out of scope these tests

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -73,194 +101,375 @@ func enabledFieldsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
...
-				s.Trillian = rhtasv1.ServiceReference{}
+				// nil File
 			}
 		},
-		func(s *rhtasv1.RekorSpec, c randfill.Continue) {
-			c.FillNoCustom(s)
+	}
+}
+
+// tsaSignerFuzzerFuncs ensures only one signer (File/Kms/Tink) is set at a time.
```
osmman: these round trip tests should not try to simulate validation rules, the data are not verified

api/v1alpha1/conversion_roundtrip_test.go
osmman: this whole function can be deleted

api/v1alpha1/conversion_roundtrip_test.go
bouskaJ: this is not a validation rule - updated comment
```
// tsaSignerFuzzerFuncs constrains to one signer at a time — not a validation rule,
// but a conversion limitation: if multiple signers (Kms/Tink) are configured, their
// per-signer Auth is merged into a single v1 Auth field and we can't split it back.
```

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -73,194 +101,347 @@ func enabledFieldsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
...
+}
+
+func ctlogServiceFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
+	return []interface{}{
+		func(s *rhtasv1.ServiceReference, c randfill.Continue) {
+			*s = randServiceReference(c, randHttpUrl)
+		},
+		func(s *CtlogService, c randfill.Continue) {
+			c.FillNoCustom(s)
+			s.Address = randHttpUrl(c, false)
```
osmman: It looks like all *Service are not generating empty cases for Address, Port, ...

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -38,6 +38,34 @@ func rhtasScheme() *runtime.Scheme {
 	return s
 }
 
+// randUrlPath generates a URL-safe path
+func randUrlPath(c randfill.Continue) string {
+	if c.Intn(3) == 0 {
```
osmman: It is missing scenarios or it should be `bool`. Expect that there should be scenario for empty tring

osmman: expect some similiar problem with randHttpUrl and randGrpcUrl

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -73,194 +101,347 @@ func enabledFieldsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
...
+			// port is defined as parameter - do not inject
+			s.Address = randGrpcUrl(c, false)
+
+			// port is always set (defaulter)
+			s.Port = ptr.To(int32(c.Intn(65534) + 1))
+		},
+	}
+}
+
+func ctlogServiceFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
```
osmman, line 134: This function (and its siblings `ctlogServiceFuzzerFuncs`/`rekorServiceFuzzerFuncs`/`tsaServiceFuzzerFuncs`/`fulcioServiceFuzzerFuncs`) always forces `Address` and `Port` to be set together, non-empty. That's mimicking "valid-looking" data rather than testing what's actually reachable — and it's hiding a real bug.

**Bug:** `Convert_v1alpha1_*Service_To_v1_ServiceReference` silently drops `Port` (and, for Ctlog, `Prefix`) whenever `Address` is empty. `buildURL()` in `conversion_overrides_utils.go` (and the equivalent inline logic in `Convert_v1alpha1_TrillianService_To_v1_ServiceReference`) returns `""` immediately when `address == ""`, discarding port/path — even though neither needs to be tied to address.

**Example:** `TrillianService.Port` has a real CRD default of `8091`, but `Address` has no default. A freshly created `CTlog` CR with no explicit Trillian address config legitimately looks like:

```go
Trillian: { Address: "", Port: &8091 }
```

Converting to v1 and back to v1alpha1 currently produces:

```go
Trillian: { Address: "", Port: nil }
```

Port `8091` is lost. Same pattern affects `CtlogService`, `FulcioService`, `RekorService`, `TsaService` via `buildURL`.

Once these fuzzers stop forcing Address+Port together, `TestCTlogConversion`/`TestRekorConversion`/`TestTufConversion`/`TestSecuresignConversion` reproduce this deterministically.

**Fix direction:** encode port (and prefix) independently of whether address is set, instead of bailing out on `address == ""`.

bouskaJ: It is not a bug - it is intention. Empty `Address` means use autoconfiguration. We do not rely on the default value of `port` (that is fragile) but we use registered internal service so port is irrelevant and intentionally dropped. The same apply for prefix on Ctlog

api/v1alpha1/conversion_roundtrip_test.go
bouskaJ: I have added the empty address case with reasoning comment why port and prefix are irrelevant in those cases (empty address means autoconfigure from ref or autodiscover -> port + prefix are resolved from component)

api/v1alpha1/conversion_roundtrip_test.go
osmman: This is still a bug, not intentional design. The reasoning doesn't hold up.

**Prefix:** `CtlogService{ Address: "", Prefix: "custom-value" }` (a valid, currently-working release-1.4 config, e.g. Ctlog sharding where one instance serves multiple prefixes).

After round-tripping through v1 and back, conversion drops it to `CtlogService{ Address: "", Prefix: "" }`. Because `Prefix` is `json:"prefix,omitempty"`, that empty string is indistinguishable from "never set" in the marshaled output, so the apiserver's structural defaulting silently backfills it with `trusted-artifact-signer` instead — not an error, not `"custom-value"`. The migrated Fulcio CR ends up pointed at the wrong Ctlog shard with no indication anything changed.

**Port:** `TrillianService{ Address: "", Port: 9999 }` (a valid, currently-working release-1.4 config — release-1.4 read `Port` directly and required it, `server_config.go` errors if `Port == nil`).

After round-tripping through v1 and back, conversion drops it to `TrillianService{ Address: "", Port: nil }`. Because `Port` is `json:"port,omitempty"`, that's indistinguishable from "never set" in the marshaled output, so the apiserver's structural defaulting silently backfills it with `8091` instead — not `9999`. Whether the reconciler even reads that value is a separate question (it currently doesn't, since #2114 hardcodes `actions.ServerPort`), but the conversion itself still silently discards the user's configured port either way.

Both cases are why the fuzzer shouldn't force "valid-looking" Address+Port combinations: `Address: "", Port/Prefix: <custom>` is legal and previously worked. Had the fuzzer generated it, this test would have caught the #2114 regression before it shipped, instead of after.

Please fix the conversion (encode port/prefix independently of address) rather than justify dropping them.

PR #2175, opened 2026-07-31
feat: add InitContainers, Volumes, VolumeMounts for Fulcio and CTLog

sampras343:
## Summary

Add a reusable `PodExtensions` API enabling users to inject init containers, volumes, and volume mounts into component deployments. This is the foundation for PKCS#11/HSM signer support, where users need to mount HSM token stores, copy vendor PKCS#11 libraries via init containers, and attach SoftHSM configuration.

## Why

The operator currently controls the full pod spec — users cannot add custom volumes, init containers, or volume mounts to Fulcio, CTLog, or any other component. This blocks any integration that requires vendor-specific pod customization (HSM devices, sidecar proxies, custom CA bundles mounted from PVCs, CSI-backed secrets).

Rather than adding PKCS#11-specific fields, this PR introduces a generic `PodExtensions` struct that can serve any pod customization use case. It is embedded on all component specs (Fulcio, CTLog, Rekor, TSA, TUF, Trillian) so the API is consistent from the start.

## Design decisions

- **`AdditionalVolume` instead of raw `core.Volume`** — Restricts volume sources to Secret, ConfigMap, EmptyDir, PVC, CSI, and Projected. This reduces CRD OpenAPI schema size and prevents users from specifying unsupported cloud-provider volume types. Exactly-one-source is enforced via `MinProperties=1` / `MaxProperties=1` (CEL was not viable — nested volume source schemas exceed the cost budget by 2.7x).

- **`PodExtensions` as inline-embedded struct** — Follows the `PodRequirements` pattern. JSON paths remain flat (`spec.initContainers`, not `spec.podExtensions.initContainers`). One type to maintain as fields are added to more components.

- **User-first, operator-after ordering** — In the `CreateOrUpdate` chain, `PodExtensions` is applied before operator ensure functions. This means operator-managed volumes always win on name collisions, protecting the deployment from user misconfiguration.

- **DefaultMode at webhook level** — User-defined volumes get `DefaultMode=0644` set by the webhook defaulter (`AdditionalVolume.SetDefaults`), not by the reconciler. This prevents infinite reconciliation loops where nil vs *420 causes a diff every cycle.

- **`ensure.Optional` for signer dispatch** — `ensureCommonDeployment` is extracted as a standalone ensure function. Signer-specific logic (`ensureFileCADeployment`) is wrapped with `ensure.Optional(type == file)`, preparing for the PKCS#11 path in PR #2182.

- **Controller wiring only for Fulcio and CTLog** — PodExtensions is embedded on all 6 components for API consistency, but only Fulcio and CTLog controllers currently wire it into their deployments. Other components will wire it when they adopt the feature.


osmman (dismissed):
The user pod-resource logic (`ReconcileUserPodResources`) is nested inside `ensureFileCADeployment`/`ensureCommonDeployment`. It should be its own entry in the `CreateOrUpdate(...)` list, split from the base deployment and from the signer-specific part:

```go
kubernetes.CreateOrUpdate(ctx, i.Client, dp,
    i.ensureDeployment(instance, RBACName, labels),   // generic: replicas, selector, labels, SA, ports, probes
    i.ensureFileSigner(instance),                     // file-CA specific: args, cert/config/oidc volumes+mounts
    deployment.PodResources(instance.Spec.InitContainers, instance.Spec.Volumes,
        instance.Spec.VolumeMounts, containerName),
    ensure.ControllerReference[*v1.Deployment](instance, i.Client),
    ensure.Labels[*v1.Deployment](...),
    deployment.PodRequirements(instance.Spec.PodRequirements, containerName),
    deployment.PodSecurityContext(),
)
```

Reason: whatever the user sets in `Spec.InitContainers`/`Volumes`/`VolumeMounts` should just get applied to the deployment — the ensure function has no business knowing which signer is active. And since it's generic, the same fields and the same `PodResources` trait belong on **Rekor**, **TSA**, and **TUF** too, not something reimplemented per component. Nesting it in Fulcio's file-CA function blocks all of that.

Same applies to CTLog's `ensureDeployment`.

sampras343:
@osmman Addressed in 0af195e2.

`ReconcileUserPodResources` is now extracted from `ensureCommonDeployment`/`ensureDeployment` into a standalone `deployment.PodResources(...)` ensure function, placed in the `CreateOrUpdate` chain alongside `PodRequirements`, `Proxy`, etc.:

```go
kubernetes.CreateOrUpdate(ctx, i.Client, dp,
    deployment.PodResources(instance.Spec.InitContainers,
        instance.Spec.Volumes, instance.Spec.VolumeMounts, containerName),
    i.ensureFileCADeployment(instance, RBACName, labels),
    ensure.ControllerReference[*v1.Deployment](instance, i.Client),
    ensure.Labels[*v1.Deployment](...),
    deployment.PodRequirements(...),
    deployment.PodSecurityContext(),
    ...
)
```

The inner `ensureCommonDeployment` (Fulcio) and `ensureDeployment` (CTLog) no longer touch user-defined init containers, volumes, or volume mounts — they only handle component-specific scaffolding (replicas, selector, ports, probes, signer-specific args/volumes).

`PodResources` is placed before the signer-specific function so operator-managed volumes always win on name collision. The function lives in `internal/utils/kubernetes/ensure/deployment/` — ready for Rekor, TSA, and TUF to adopt (if needed).

internal/utils/kubernetes/ensure/pod_spec.go
```diff
@@ -67,3 +68,94 @@ func ensureContainerSecurityContext(container *core.Container) {
...
+
+	for _, vol := range volumes {
+		v := kubernetes.FindVolumeByNameOrCreate(podSpec, vol.Name)
+		v.VolumeSource = vol.VolumeSource
+		EnsureVolumeDefaultMode(v)
+	}
+
+	for _, vm := range volumeMounts {
+		m := kubernetes.FindVolumeMountByNameOrCreate(container, vm.Name)
+		m.MountPath = vm.MountPath
```
bouskaJ: VolumeMount reconciliation silently drops fields — ReconcileUserPodResources only copies MountPath, SubPath, and ReadOnly, but the API accepts the full corev1.VolumeMount including SubPathExpr,
  MountPropagation, and RecursiveReadOnly. Users can set these in the CR and they'll be silently ignored. Fix: assign the whole struct

api/v1/common.go
```diff
@@ -266,6 +266,42 @@ type ServiceAccountConfig struct {
...
+	Image string `json:"image"`
+	// Entrypoint array. Not executed within a shell.
+	//+optional
+	Command []string `json:"command,omitempty"`
+	// Arguments to the entrypoint.
+	//+optional
+	Args []string `json:"args,omitempty"`
+	// List of environment variables to set in the container.
+	//+optional
+	Env []core.EnvVar `json:"env,omitempty"`
```
bouskaJ, line 212: Missing +listType=map / +listMapKey=name SSA markers. Not only `Env`. Review all new fields

api/v1/ctlog_types.go
```diff
@@ -71,6 +72,15 @@ type CTlogSpec struct {
 	// ConfigMap with additional bundle of trusted CA
 	// +optional
 	TrustedCA *LocalObjectReference `json:"trustedCA,omitempty"`
+	// InitContainers to run before the main server container.
+	//+optional
+	InitContainers []InitContainerSpec `json:"initContainers,omitempty"`
```
bouskaJ: Consider grouping into a reusable struct — The three fields are added as loose top-level fields. A PodExtensions struct (similar to existing PodRequirements) would give one type to maintain as these fields are added to Rekor/TSA/Trillian.

api/v1/fulcio_types.go
```diff
@@ -37,6 +38,15 @@ type FulcioSpec struct {
 	// ConfigMap with additional bundle of trusted CA
 	//+optional
 	TrustedCA *LocalObjectReference `json:"trustedCA,omitempty"`
+	// InitContainers to run before the main server container.
+	//+optional
+	InitContainers []InitContainerSpec `json:"initContainers,omitempty"`
+	// Additional volumes to attach to the deployment pods.
+	//+optional
+	Volumes []core.Volume `json:"volumes,omitempty"`
```
bouskaJ: InitContainerSpec carefully curates 10 fields from core.Container — good. But Volumes embeds raw `core.Volume`, which drags in 30+ volume source types (AWSElasticBlockStore, Flocker, Cinder, etc.) adding ~12,000 lines of OpenAPI schema across all CRDs. We may  solved this with AdditionalVolume restricting sources to Secret, ConfigMap, EmptyDir, PVC, CSI, and Projected (as Strimzi operator does).

api/v1/common.go
```diff
@@ -266,6 +266,42 @@ type ServiceAccountConfig struct {
 	ImagePullSecrets []core.LocalObjectReference `json:"imagePullSecrets,omitempty"`
 }
 
+// InitContainerSpec defines a curated subset of corev1.Container for custom init containers.
+// These containers run before the main server to perform vendor-specific initialization.
+type InitContainerSpec struct {
```
bouskaJ, line 193: Consider adding restartPolicy to InitContainerSpec to support native sidecar containers. The current InitContainerSpec covers the "copy library and exit" PKCS11 pattern, but some HSM integration approaches require a long-running sidecar container alongside the main process (e.g., a PKCS11 proxy
  daemon or a vendor-specific hardware access server).

 Since Kubernetes 1.29, native sidecar containers are expressed as init containers with `restartPolicy: Always` — they start before the main container and keep running for the pod's lifetime.

Adding a single field would unlock this:
```
  // +optional
  // +kubebuilder:validation:Enum=Always
  RestartPolicy *core.ContainerRestartPolicy `json:"restartPolicy,omitempty"`
```
 Without this option, users whose HSM setup requires a running daemon cannot express it through the API.
Worth considering now while the API is still taking shape.

bouskaJ (changes requested):

sampras343:
**Addressing all review comments from @bouskaJ in commit 888a83d7:**

1. **VolumeMount drops fields** (`pod_spec.go:119`) — Fixed. `ReconcileUserPodResources` now assigns the whole `VolumeMount` struct instead of field-by-field copy, preserving `SubPathExpr`, `MountPropagation`, and `RecursiveReadOnly`.

2. **Missing SSA markers** (`common.go:289`) — Fixed. Added `+listType=map` / `+listMapKey=name` on all list fields: `InitContainers`, `Volumes`, `VolumeMounts` (on `PodExtensions`) and `Env`, `VolumeMounts` (on `InitContainerSpec`).

3. **Group into reusable struct** (`ctlog_types.go:77`) — Done. Introduced `PodExtensions` struct in `common.go` with `json:",inline"` embedding — same pattern as `PodRequirements`. JSON paths remain unchanged (`spec.initContainers`, not `spec.podExtensions.initContainers`). Ready to embed in Rekor/TSA/TUF when those components adopt it.

4. **Raw core.Volume schema bloat** (`fulcio_types.go:46`) — Done. Introduced `AdditionalVolume` / `AdditionalVolumeSource` restricting volume sources to Secret, ConfigMap, EmptyDir, PVC, CSI, and Projected. Includes CEL validation ensuring at least one source is set. CRD schema reduced by ~7400 lines. `ToVolume()` converter handles the ensure layer translation.

5. **Add restartPolicy for native sidecars** (`common.go:271`) — Done. Added `RestartPolicy *core.ContainerRestartPolicy` with `+kubebuilder:validation:Enum=Always` to `InitContainerSpec`. Wired into `ReconcileInitContainers`. Setting `restartPolicy: Always` on an init container creates a native sidecar (Kubernetes 1.29+).

All changes tested on OCP 4.22 — stack deployed in file mode, cosign sign + verify passed.

sampras343:
**@osmman** — Your review comment about extracting `ReconcileUserPodResources` into a standalone `deployment.PodResources(...)` ensure function was addressed in commit 5411254d.

With 888a83d7, the signature is further simplified — `PodResources` now takes a single `PodExtensions` struct (matching the `PodRequirements` pattern):

```go
deployment.PodResources(instance.Spec.PodExtensions, containerName),
```

Both Fulcio and CTLog use identical call sites in their `CreateOrUpdate` chains, fully decoupled from signer-specific logic.

api/v1/common.go
sampras343: Done. Added RestartPolicy *core.ContainerRestartPolicy with +kubebuilder:validation:Enum=Always to InitContainerSpec. Wired into ReconcileInitContainers. Setting restartPolicy: Always on an init container creates a native sidecar (Kubernetes 1.29+).

api/v1/fulcio_types.go
sampras343: Done. Introduced AdditionalVolume / AdditionalVolumeSource restricting volume sources to Secret, ConfigMap, EmptyDir, PVC, CSI, and Projected. Includes CEL validation ensuring at least one source is set. CRD schema reduced by ~7400 lines. ToVolume() converter handles the ensure layer translation.

I will also document this in the ADR

api/v1/ctlog_types.go
sampras343: Done. Introduced PodExtensions struct in common.go with json:",inline" embedding — same pattern as PodRequirements.

api/v1/common.go
sampras343: Fixed. Added +listType=map / +listMapKey=name on all list fields: InitContainers, Volumes, VolumeMounts (on PodExtensions) and Env, VolumeMounts (on InitContainerSpec).

internal/utils/kubernetes/ensure/pod_spec.go
sampras343: Fixed. ReconcileUserPodResources now assigns the whole VolumeMount struct instead of field-by-field copy, preserving SubPathExpr, MountPropagation, and RecursiveReadOnly.

api/v1/common.go
```diff
@@ -266,6 +266,124 @@ type ServiceAccountConfig struct {
...
+	// Additional volume mounts for the main server container.
+	//+optional
+	// +listType=map
+	// +listMapKey=name
+	VolumeMounts []core.VolumeMount `json:"volumeMounts,omitempty"`
+}
+
+// AdditionalVolume defines a named volume with a restricted set of source types.
+// This avoids exposing the full corev1.VolumeSource (30+ types) in the CRD schema.
+// +kubebuilder:validation:XValidation:rule="has(self.secret) || has(self.configMap) || has(self.emptyDir) || has(self.persistentVolumeClaim) || has(self.csi) || has(self.projected)",message="at least one volume source must be specified"
```
bouskaJ: Multiple volume sources allowed per AdditionalVolume
  The CEL validation only checks that at least one source is present, but doesn't enforce mutual exclusivity. A user can set both secret and configMap on the same volume — the CRD accepts it, CreateOrUpdate
  succeeds, but the pod fails to schedule with no useful operator error. Fix with a CEL rule: [has(self.secret), ...].filter(b, b).size() == 1.

internal/utils/kubernetes/ensure/pod_spec.go
```diff
@@ -67,3 +68,95 @@ func ensureContainerSecurityContext(container *core.Container) {
 		container.SecurityContext.RunAsUser = nil
 	}
 }
+
+// HasVolume returns true if the pod spec already contains a volume with the
+// given name.
+func HasVolume(spec *core.PodSpec, name string) bool {
```
bouskaJ: no usage

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -398,6 +399,88 @@ func securesignStatusFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{}
...
+					{Name: volName, MountPath: "/mnt/" + c.String(5)},
+				}
+			} else {
+				s.Volumes = nil
+				s.VolumeMounts = nil
+			}
+		},
+	}
+}
+
```
bouskaJ: I think we can have common fuzzer for the `PodExntensions` type

internal/utils/kubernetes/ensure/deployment/deployment.go
```diff
@@ -55,6 +55,18 @@ func PodRequirements(requirements rhtasv1.PodRequirements, containerName string)
 	}
 }
 
+// PodResources applies user-defined init containers, volumes, and volume mounts
+// to a Deployment. It is a standalone ensure function so callers can compose it
+// independently from signer-specific logic.
+func PodResources(ext rhtasv1.PodExtensions, containerName string) func(*v1.Deployment) error {
```
bouskaJ: While we work with the `PodExtensions` struct, I'd prefer to keep the `PodExtensions` name. Under the `PodResources` naming, I imagine things like memory, disk space, etc.

internal/controller/ctlog/actions/deployment.go
```diff
@@ -145,6 +140,18 @@ func (i deployAction) ensureDeployment(instance *rhtasv1.CTlog, sa string, label
...
+		// so operator always wins if a user volume has the same name.
+		// Set the full VolumeSource to clear any conflicting source a user
+		// volume may have introduced (e.g. EmptyDir).
+		volume := kubernetes.FindVolumeByNameOrCreate(&template.Spec, volumeName)
+		volume.VolumeSource = core.VolumeSource{
+			Secret: &core.SecretVolumeSource{
+				SecretName: instance.Status.ServerConfigRef.Name,
+			},
+		}
+		ensure.EnsureVolumeDefaultMode(volume)
```
bouskaJ: you can keep the block where it was and add the comment on `CreateOrUpdate` where the actual ordering matters.

internal/utils/kubernetes/ensure/pod_spec.go
```diff
@@ -67,3 +68,95 @@ func ensureContainerSecurityContext(container *core.Container) {
...
+	}
+	return false
+}
+
+// EnsureVolumeDefaultMode applies the same DefaultMode that the Kubernetes API
+// server would apply (0644 / octal 0644 = 420 decimal) to volume sources that
+// support it. Without this, a user-specified volume that omits DefaultMode
+// would differ from the API server's response on every reconcile (nil vs *420),
+// causing an infinite update loop.
+func EnsureVolumeDefaultMode(v *core.Volume) {
```
bouskaJ: We can default `AdditionalVolumeSource` on the API level

internal/controller/fulcio/actions/deployment.go
```diff
@@ -68,7 +68,8 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1.Fulcio) *act
 				Namespace: instance.Namespace,
 			},
 		},
-		i.ensureDeployment(instance, RBACName, labels, ctlogUrl),
+		deployment.PodResources(instance.Spec.PodExtensions, containerName),
+		i.ensureFileCADeployment(instance, RBACName, labels, ctlogUrl),
```
bouskaJ:
```suggestion
        i.ensureCommonDeployment(instance, RBACName, labels, ctlogUrl)
		ensure.Optional(instance.Spec.Signer.Type == "file", i.ensureFileCADeployment(instance, RBACName, labels, ctlogUrl)),
```

api/v1/fulcio_types.go
```diff
@@ -36,7 +36,8 @@ type FulcioSpec struct {
 	Monitoring MonitoringConfig `json:"monitoring,omitempty"`
 	// ConfigMap with additional bundle of trusted CA
 	//+optional
-	TrustedCA *LocalObjectReference `json:"trustedCA,omitempty"`
+	TrustedCA     *LocalObjectReference `json:"trustedCA,omitempty"`
+	PodExtensions `json:",inline"`
```
bouskaJ, line 40: While we are working on the new API it would be wise to consider adding this to all components and avoid future API changes (API additions are allowed so it is not mandatory but it is always good to be one step before).

internal/utils/kubernetes/ensure/pod_spec.go
sampras343: removed

api/v1/common.go
sampras343: Fixed with MinProperties=2 / MaxProperties=2 (name + one source) on AdditionalVolumeSource. CEL was not viable — nested volume source types cause any CEL rule to exceed the cost budget by 2.7x regardless of complexity. MaxProperties=1 achieves identical exactly-one-source enforcement at the schema level with zero cost.

Is this acceptable @bouskaJ ?

api/v1alpha1/conversion_roundtrip_test.go
sampras343: Done. Extracted `fuzzPodExtensions` helper, single `podExtensionsFuzzerFuncs` replaces the two duplicates.

internal/utils/kubernetes/ensure/deployment/deployment.go
sampras343: Done. Function and all call sites renamed.

internal/controller/ctlog/actions/deployment.go
sampras343: Done. User PodExtensions applied first, operator ensures run after so operator wins on name collisions. Comment added to both Fulcio and CTLog CreateOrUpdate chains.

internal/controller/fulcio/actions/deployment.go
sampras343: Done. `ensureCommonDeployment` extracted as standalone ensure function (sets shared scaffolding + common args). `ensureFileCADeployment` wrapped with ensure.Optional(type == file || type == ""), ready for PKCS11 path

internal/utils/kubernetes/ensure/pod_spec.go
sampras343: Done. `AdditionalVolume.SetDefaults()` and `PodExtensions.SetDefaults()` set `DefaultMode` at webhook level. Wired into all 6 component defaulters. Removed EnsureVolumeDefaultMode from the `ReconcileUserPodResources` path. The function itself is retained because operator-managed volumes (fulcio-config, fulcio-cert, oidc-info, ctlog keys) are `core.Volume` created inline in the controller and don't go through the webhook defaulter.

api/v1/fulcio_types.go
sampras343: Done. Embedded on all 6 component specs (Fulcio, CTLog, Rekor, TSA, TUF, Trillian). SetDefaults wired. Conversion restore lines added in  both standalone component ConvertTo functions and securesign_conversion.go. Controller wiring for Rekor, TSA, TUF, and Trillian deferred to when those components adopt the feature, the API fields are reserved for future use.

internal/utils/kubernetes/ensure/pod_spec.go
bouskaJ: Operator volumes are under our control. I would rather remove this function and set the mode inline on the volume source ref (transparency, readability)

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -412,6 +413,56 @@ func securesignStatusFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{}
...
+		ext.Volumes = nil
+		ext.VolumeMounts = nil
+	}
+}
+
+// podExtensionsFuzzerFuncs generates roundtrip-safe PodExtensions values
+// for both FulcioSpec and CTlogSpec.
+func podExtensionsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
+	return []interface{}{
+		func(s *rhtasv1.FulcioSpec, c randfill.Continue) {
```
bouskaJ: Do not fuzz FulcioSpec/CtlogSpec - just register PodExtensions fuzzer similarly as we do on https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/conversion_roundtrip_test.go#L531

bouskaJ (changes requested):

api/v1alpha1/conversion_roundtrip_test.go
sampras343: Fuzzer now registers on `*rhtasv1.PodExtensions` directly, the randfill framework applies it automatically to any struct embedding PodExtensions.

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -412,6 +413,50 @@ func securesignStatusFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{}
...
+		ext.VolumeMounts = nil
+	}
+}
+
+// podExtensionsFuzzerFuncs registers a type-level fuzzer for PodExtensions.
+// The randfill framework applies it to any struct embedding PodExtensions.
+func podExtensionsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
+	return []interface{}{
+		func(s *rhtasv1.PodExtensions, c randfill.Continue) {
+			fuzzPodExtensions(s, c)
```
bouskaJ: could you inline the func?

config/manager/kustomization.yaml
```diff
@@ -4,6 +4,6 @@ resources:
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
-- newTag: 1.5.0
-  name: controller
-  newName: registry.redhat.io/rhtas/rhtas-rhel9-operator
+- name: controller
+  newName: quay.io/rh-ee-sacm/secure-sign-operator
```
bouskaJ: do not overwrite image

bouskaJ (changes requested):

config/manager/kustomization.yaml
sampras343: Ah apologies. Was testing the feature on a local cluster. Pushed these by mistake. Reverted.

api/v1alpha1/conversion_roundtrip_test.go
sampras343: Done

internal/utils/kubernetes/ensure/pod_spec.go
sampras343: Removed. and added inline wherever required

bouskaJ (approved):

PR #2179, opened 2026-08-04
fix: trillian logsigner actions use wrong condition type

sampras343:
## Summary

Two pre-existing bugs in the Trillian logsigner controller actions that prevent the SecureSign stack from reliably reaching Ready state:

- **Wrong condition type:** `logsigner/service.go` and `logsigner/monitoring.go` use `actions.ServerCondition` ("LogServerAvailable") instead of `actions.SignerCondition` ("LogSignerAvailable"). When the logsigner service is created, it overwrites the logserver's condition to "Creating", causing `LogServerAvailable` to oscillate between Creating and Ready.

- **Missing ObservedGeneration:** All four deployment/service actions in logserver and logsigner set conditions without `ObservedGeneration`, inconsistent with the rollout check actions.

The combined effect blocks the dependency chain: Trillian never reaches Ready → CTLog can't complete → TUF can't resolve keys → SecureSign stays stuck.

### Files changed
| File | Change |
|------|--------|
| `logsigner/service.go` | `ServerCondition` → `SignerCondition` + add `ObservedGeneration` |
| `logsigner/monitoring.go` | `ServerCondition` → `SignerCondition` |
| `logsigner/deployment.go` | Add `ObservedGeneration` |
| `logserver/deployment.go` | Add `ObservedGeneration` |
| `logserver/service.go` | Add `ObservedGeneration` |

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/controller/trillian/...` passes
- [ ] Deploy SecureSign CR — all conditions reach Ready without oscillation
- [ ] Verify Trillian LogServerAvailable and LogSignerAvailable both stabilize at Ready

### DIFF
--- a/internal/controller/trillian/actions/logserver/deployment.go
+++ b/internal/controller/trillian/actions/logserver/deployment.go
@@ -77,10 +77,11 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1.Trillian) *a
 
 	if result != controllerutil.OperationResultNone {
 		meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
-			Type:    actions.ServerCondition,
-			Status:  metav1.ConditionFalse,
-			Reason:  state.Creating.String(),
-			Message: "Deployment created",
+			Type:               actions.ServerCondition,
+			Status:             metav1.ConditionFalse,
+			Reason:             state.Creating.String(),
+			Message:            "Deployment created",
+			ObservedGeneration: instance.Generation,
 		})
 		return i.ReturnOnChange(i.PersistStatus)(ctx, instance)
 	} else {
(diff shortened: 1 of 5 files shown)


osmman (approved):

osmman:
This PR should not be part of stack it is independent fix.

sampras343:
> This PR should not be part of stack it is independent fix.

Add this as a part of the stack because without this fix, my in-cluster testing breaks and overall stack does not come up.
I believe once this is merged it will go out of stack and the dependent PR will point to main

osmman:
> > This PR should not be part of stack it is independent fix.
> 
> Add this as a part of the stack because without this fix, my in-cluster testing breaks and overall stack does not come up. I believe once this is merged it will go out of stack and the dependent PR will point to main

This shouldn't be part of the stack, bundling it means it only merges after the whole stack does. Sorry you're hitting this in your deployment, but I checked our CI (Kind and OCP deployments) and see no failures related to it across ~2k runs (30 day window), so I'd rather this land independently and quickly.

PR #2181, opened 2026-08-04
feat: add signer Auth support for Fulcio and CTLog

sampras343:
## Summary

Add an Auth field on FulcioSigner and CTlogSigner, enabling users to inject authentication credentials into the signer deployment. This is needed for PKCS#11/HSM integration where the signer process requires environment variables (e.g. SOFTHSM2_CONF) and secret mounts (e.g. vendor credentials) to authenticate with the HSM backend.

## Why

The operator's signer configuration currently has no way to pass authentication context to the running container. PKCS#11 modules, cloud KMS providers, and vendor HSM drivers all need environment variables or mounted secrets to authenticate. Without this, users cannot configure HSM-backed signing — the container starts but cannot access the HSM.

The Auth type already exists in the codebase (used by Trillian and Rekor for database authentication). This PR reuses it on FulcioSigner and CTlogSigner, keeping the API consistent.

## Design decisions

- **Reuses existing Auth type** — `Auth{Env []EnvVar, SecretMount []SecretKeySelector}` from `common.go`. No new types needed.

- **Auth on the signer, not the spec** — Authentication is signer-specific (different backends need different credentials). Placing it on `FulcioSigner`/`CTlogSigner` rather than `FulcioSpec`/`CTlogSpec` makes the scope clear.

- **Shared ensure/auth.go untouched** — The existing `ContainerAuth` implementation is intentionally not modified. It's shared by Trillian (which calls it twice per reconcile for user auth + DB auth). Changing its behavior here would break that pattern.

- **Signer defaults removed from Securesign** — Sub-resource webhooks handle their own defaults. The Securesign umbrella CR only wires service bindings.


internal/utils/kubernetes/ensure/auth.go
```diff
@@ -21,26 +21,32 @@ func Auth(containerName string, auth *rhtasv1.Auth) func(spec *core.PodSpec) err
...
-				}
-			}
-			authProjected := kubernetes.FindVolumeByNameOrCreate(templateSpec, authVolumeName)
-			if authProjected.Projected == nil {
-				authProjected.Projected = &core.ProjectedVolumeSource{}
-			}
+		if auth == nil || (len(auth.Env) == 0 && len(auth.SecretMount) == 0) {
+			kubernetes.RemoveVolumeByName(templateSpec, authVolumeName)
+			kubernetes.RemoveVolumeMountByName(container, authVolumeName)
+			return nil
```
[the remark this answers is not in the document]
sampras343: **Acknowledged — pre-existing limitation, not a regression from this PR.**

Main's `ContainerAuth` also never removes env vars — it only upserts via `FindEnvByNameOrCreate`. This is the same pattern used by all existing callers (Rekor, Trillian, TSA).

The suggested marker-based approach (a `TAS_AUTH_MANAGED_ENVS` env var tracking managed names) adds non-trivial complexity:
- Auth env var names are user-specified and can overlap with names set by other ensure steps (proxy, GODEBUG, FIPS vars)
- Safe pruning requires distinguishing "auth-managed" vs "operator-managed" vs "Kubernetes-defaulted" env vars across all components
- The marker itself becomes a new contract that all callers must respect

This deserves a separate design discussion and PR since it affects the shared `ensure.Auth` function used by Rekor, Trillian, and TSA — not just the Fulcio/CTLog callers added here.

For now, the behavior matches main: removing `auth.Env` entries from the CR does not auto-remove them from the deployment. Users can work around this by setting the env var to an empty value rather than removing it.

internal/utils/kubernetes/ensure/auth.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	authVolumeName = "auth"
+	authVolumeName = "signer-auth"
 	AuthMountPath  = constants.SecretMountPath + "/auth"
 )
```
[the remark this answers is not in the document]
sampras343: **Fixed in 61cffc7f.**

Added `legacyAuthVolumeName = "auth"` constant and unconditional cleanup at the top of every `ContainerAuth` call:

```go
kubernetes.RemoveVolumeByName(templateSpec, legacyAuthVolumeName)
kubernetes.RemoveVolumeMountByName(container, legacyAuthVolumeName)
```

This runs before any other logic (both the nil cleanup path and the non-nil upsert path), so on upgrade from main (where the volume was named `"auth"`) to this PR (where it's `"signer-auth"`), the old volume and mount are always removed — regardless of whether auth is currently configured or not.

No conditional shape-checking needed since `RemoveVolumeByName`/`RemoveVolumeMountByName` are no-ops if the volume doesn't exist, and the legacy `"auth"` name was exclusively operator-managed (not user-facing).

osmman (dismissed):
**1. Don't touch the shared `ensure/auth.go` implementation. It broke Trillian and should stay out of this PR.**

`ContainerAuth` now resets the volume's `VolumeSource` on every call instead of appending. Trillian (`trillian/utils/db_support.go:31,39,107,113`) calls `ContainerAuth` twice per reconcile on the same pod spec (user auth, then DB-secret auth) to compose both onto one volume. The second call now wipes the `SecretMount` entries added by the first. Any Trillian CR with `spec.auth.secretMount` set and an operator-provisioned DB will silently lose that mount. This function is shared by Rekor, Trillian, and TSA, not just Fulcio/CTLog, and this PR only tests the two new callers.

The stale-resource cleanup you're trying to add here is a real, pre-existing gap, but it isn't specific to Fulcio/CTLog and the fix you picked breaks another consumer. This needs a generic convention across all `ensure` helpers, designed and tested on its own. Please file it as a separate **Jira bug** report and revert the behavior change in `auth.go` for this PR. The existing cleanup limitation was not introduced by you, so it shouldn't block this feature.

**2. Volume rename `auth` to `signer-auth` is unnecessary migration that doesn't solve the collision.**

Renaming forces a rolling restart on every existing Rekor/Trillian/TSA/Fulcio/CTLog deployment, but doesn't prevent collisions. A user can name a volume `signer-auth` just as easily as `auth`. Revert the rename.

A user reusing the `auth` volume name is a valid way to override or replace our generated auth mount. It gives them a workaround when they need one. Just document the reserved name so they know the effect.

sampras343:
Addressed in b9e979f7.

**1. auth.go reverted** — `ContainerAuth` is now identical to main. The VolumeSource reset and nil/empty cleanup path are removed. Trillian's double-call pattern (user auth + DB auth composing onto the same projected volume) is preserved. Filed the stale-resource cleanup as a separate concern — it needs a generic convention across all `ensure` helpers, designed and tested against all consumers.

**2. Volume rename reverted** — back to `"auth"`. No unnecessary rolling restart on existing deployments. Will document the reserved name.

**3. gofmt fixed** — extra blank line removed from `ctlog/actions/deployment.go`.

`auth_test.go` removed since it tested the reverted behavior. The auth *feature* (field on signer, ContainerAuth call in both deployments, conversion roundtrip, deployment tests) is unchanged and E2E verified on OCP 4.22.

api/v1/securesign_defaults.go
```diff
@@ -1,7 +1,8 @@
 package v1
 
 func (s *Securesign) SetDefaults() {
-	// keep securesign minimal - defaulted on sub-resource level
+	s.Spec.Fulcio.Signer.SetDefaults()
```
bouskaJ: The Securesign resource is meant to be an umbrella — keep defaults on it minimal. The current defaults are service-binding only (wiring sub-resources together via ServiceReferenceRef). I don't think we need to default signer types on Securesign — they are correctly defaulted on the sub-resources.

bouskaJ (changes requested):

api/v1/securesign_defaults.go
sampras343: Agreed and removed.

api/v1/securesign_defaults.go
```diff
@@ -1,7 +1,7 @@
 package v1
 
 func (s *Securesign) SetDefaults() {
-	// keep securesign minimal - defaulted on sub-resource level
+	// keep securesign minimal - signer defaults are handled by sub-resource webhooks
```
bouskaJ:
```suggestion
	// keep securesign minimal - component defaults are handled by sub-resource webhooks
```
this is not just about signer defautls - it apply to all

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -481,6 +481,58 @@ func ctlogVolumesFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
...
+				c.FillNoCustom(&s.File.PublicKeyRef)
+				if s.File.PrivateKeyRef == nil && s.File.PrivateKeyPasswordRef == nil && s.File.PublicKeyRef == nil { //nolint:staticcheck
+					s.File.PrivateKeyRef = &rhtasv1.SecretKeySelector{}
+					c.FillNoCustom(s.File.PrivateKeyRef)
+				}
+			}
+		},
+	}
+}
+
```
bouskaJ: why do you create custom fuzzer? We already have resources with Auth and they work without any customization.

api/v1/securesign_defaults.go
sampras343: Comment updated to "component defaults" instead of "signer defaults"

api/v1alpha1/conversion_roundtrip_test.go
sampras343: Custom auth fuzzers removed — default FillNoCustom handles Auth round-trip correctly

internal/controller/fulcio/actions/deployment.go
```diff
@@ -204,6 +204,11 @@ func (i deployAction) ensureFileCADeployment(instance *rhtasv1.Fulcio) func(depl
 		container := kubernetes.FindContainerByNameOrCreate(&dp.Spec.Template.Spec, containerName)
 		template := &dp.Spec.Template
 
+		// Apply or clean up auth env vars and secret mounts
+		if err := ensure.ContainerAuth(container, instance.Spec.Auth)(&template.Spec); err != nil {
```
bouskaJ: I would extract this into CreateUpdate function - it may be part of the regular deployment. Do not restrict that to the FileCaDeployment

sampras343: I will wait for the https://github.com/securesign/secure-sign-operator/pull/2203 deployment Auth extraction in the your PR to be merged and rebase on top with these changes.

bouskaJ (approved):

PR #2182, opened 2026-08-04
feat: add PKCS#11/HSM signer support for Fulcio and CTLog

sampras343:
## Summary

Add PKCS#11/HSM-backed signer support for Fulcio and CTLog, enabling hardware security module integration for certificate authority signing and certificate transparency log signing.

## Why

Organizations with compliance requirements (FIPS 140-2, PCI-DSS, eIDAS) need signing keys to reside in hardware security modules rather than Kubernetes Secrets. Without HSM support, RHTAS cannot meet these requirements — the CA private key and CTLog signing key are stored as plain Secrets, which is insufficient for regulated environments.

## What

- **Unified PKCS#11 API**: A shared `PKCS11Config` core struct (modulePath, tokenLabel, slotNumber, pinSecretRef, keyID, keyLabel) reusable across components. Fulcio extends it with `configRef` (crypto11 JSON escape hatch for vendor-specific HSMs). CTLog extends it with `publicKeyRef` (trillian needs PEM public key for key matching).

- **Fulcio PKCS#11 signer**: Validates PKCS#11 secrets, mounts crypto11 config and certificate chain, passes `--ca=pkcs11ca` and `--hsm-caroot-id` args. Two modes — structured fields (operator generates crypto11 JSON) or configRef (user provides vendor-specific JSON).

- **CTLog PKCS#11 signer**: Generates trillian `keyspb.PKCS11Config` protobuf from structured fields, passes `--pkcs11_module_path` arg. Content-hash drift detection for config rotation.

- **Shared HSM helpers**: `EnsureHSMResources` / `CleanupHSMResources` for hsm-tokens and hsm-lib volume management, with user PVC preservation.

- **CEL validation**: Required fields enforced per component at admission — CTLog requires modulePath+tokenLabel+pinSecretRef+publicKeyRef; Fulcio requires configRef OR modulePath+tokenLabel+pinSecretRef, plus keyID.

- **E2E test suite**: `//go:build pkcs11` tagged tests with SoftHSM key ceremonies, `WithPKCS11Signer` fixture, Makefile target, and CI job.


internal/controller/fulcio/actions/ensure_pkcs11_config.go
```diff
@@ -0,0 +1,123 @@
...
+// and compared on subsequent reconciles to detect configuration changes without
+// reacting to unrelated generation bumps (e.g., monitoring toggle).
+func computePKCS11Hash(instance *rhtasv1.Fulcio) string {
+	h := sha256.New()
+	if instance.Spec.Signer.PKCS11 != nil {
+		cfg := instance.Spec.Signer.PKCS11
+		if cfg.ConfigRef != nil {
+			h.Write([]byte(cfg.ConfigRef.Name))
+			h.Write([]byte(cfg.ConfigRef.Key))
+		}
```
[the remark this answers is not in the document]
sampras343: **Fixed in b492fdaf.** `computePKCS11Hash` now includes `certificateChainRef.Name` and `certificateChainRef.Key` in the hash, with field-name prefixes as separators to prevent collisions. Also fixed `fmt.Fprintf` errcheck violations by using `h.Write` instead.

internal/controller/fulcio/actions/deployment.go
```diff
@@ -330,3 +348,134 @@ func (i deployAction) ensureFileCADeployment(instance *rhtasv1.Fulcio, sa string
...
+					{Key: instance.Status.Certificate.CARef.Key, Path: instance.Status.Certificate.CARef.Key},
+				},
+			},
+		}
+		ensure.EnsureVolumeDefaultMode(pkcs11CertVol)
+
+		hsmTokensVol := kubernetes.FindVolumeByNameOrCreate(&template.Spec, HSMTokensVolumeName)
+		if !ensure.HasVolume(&template.Spec, HSMTokensVolumeName) || (hsmTokensVol.EmptyDir == nil && hsmTokensVol.PersistentVolumeClaim == nil) {
+			hsmTokensVol.VolumeSource = core.VolumeSource{EmptyDir: &core.EmptyDirVolumeSource{}}
+		}
```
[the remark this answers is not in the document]
sampras343: **Fixed in b492fdaf.** Now checks `instance.Spec.Volumes` for a user-defined `hsm-tokens` volume before defaulting to EmptyDir — matching the CTLog pattern exactly.

internal/controller/ctlog/actions/ensure_pkcs11_config.go
```diff
@@ -0,0 +1,179 @@
...
+		return a.Error(ctx,
+			reconcile.TerminalError(fmt.Errorf("spec.signer.pkcs11 is nil but signer type is pkcs11")),
+			instance,
+		)
+	}
+
+	// Validate PinSecretRef
+	if _, err := a.validateSecretRef(ctx, instance, p.PinSecretRef, "spec.signer.pkcs11.pinSecretRef"); err != nil {
+		return a.ReturnOnChange(a.PersistStatus)(ctx, instance)
+	}
```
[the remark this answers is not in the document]
sampras343: **Fixed in b492fdaf.** Two changes:

1. `validateSecretRef` failure now returns `RequeueAfter(5s)` instead of `ReturnOnChange`, preventing later actions from proceeding with invalid config.

2. `buildPKCS11Config` in `server_config.go` now explicitly rejects empty PIN and empty public key data with clear error messages.

Also fixed lint issues: extracted `ReasonResolved` constant (goconst), removed unused `[]byte` return from `validateSecretRef` (unparam).

sampras343:
### Re: PIN duplicated into config Secret (Qodo finding #4)

**Acknowledged — this is an upstream architectural requirement, not something we can change.**

The `ct_server` binary reads the HSM PIN from its protobuf config file via `keyspb.PKCS11Config.Pin`. This is how the [trillian PKCS#11 signer](https://github.com/google/trillian/blob/main/crypto/keyspb/keyspb.pb.go) is designed — the PIN must be in the config for `ct_server` to authenticate with the HSM at startup.

The generated config Secret is:
- Owned by the CTLog CR (controller reference) — deleted when the CR is deleted
- Access-controlled via Kubernetes RBAC (same as any other Secret)
- No more exposed than the original `pinSecretRef` Secret (both are in the same namespace)

This is consistent with how other operators handle HSM PINs in protobuf configs (e.g., [sigstore/scaffolding](https://github.com/sigstore/scaffolding/blob/main/cmd/ctlog/createctconfig/main.go)).

Mitigating the blast radius further would require upstream changes to `ct_server` to support reading the PIN from an environment variable or mounted file instead of the protobuf config — that's out of scope for this PR.

internal/controller/fulcio/actions/ensure_pkcs11_config.go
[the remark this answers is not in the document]
sampras343: Already addressed — `computePKCS11Hash` includes `certificateChainRef.Name` and `certificateChainRef.Key` with labeled field separators (lines 155-158). CA chain reference changes trigger re-validation.

internal/controller/fulcio/actions/deployment.go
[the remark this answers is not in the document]
sampras343: Already addressed — HSM volume setup is now in `pkcs11helpers.EnsureHSMResources` which checks `instance.Spec.Volumes` for user-defined `hsm-tokens` before defaulting to EmptyDir. User-provided PVC/CSI/hostPath sources are preserved.

internal/controller/ctlog/actions/ensure_pkcs11_config.go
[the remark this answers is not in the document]
sampras343: Already addressed — `validateSecretRef` failure now returns `RequeueAfter(5s)` (lines 75-79, 83-87) instead of `ReturnOnChange`, preventing later actions from proceeding. `buildPKCS11Config` validates non-empty PIN and public key (server_config.go lines 308-313).

api/v1/ctlog_types.go
```diff
@@ -84,16 +84,52 @@ type CTlogSpec struct {
...
+	//+required
+	//+kubebuilder:validation:MinLength=1
+	//+kubebuilder:validation:Pattern=`^/`
+	ModulePath string `json:"modulePath"`
+}
 
 // CTlogSigner defines the desired state of the CTlog Signer
+// +kubebuilder:validation:XValidation:rule="self.type != 'pkcs11' || has(self.pkcs11)",message="pkcs11 configuration is required when type is pkcs11"
+// +kubebuilder:validation:XValidation:rule="self.type != 'pkcs11' || !has(self.file)",message="file configuration must not be set when type is pkcs11"
+// +kubebuilder:validation:XValidation:rule="self.type != 'file' || !has(self.pkcs11)",message="pkcs11 configuration must not be set when type is file"
```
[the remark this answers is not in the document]
sampras343: By design — PKCS#11 is opt-in via `type: pkcs11`. When `type` is empty/omitted, the webhook defaults it to `file` and the `pkcs11` block is unused (no-op). This is intentional: it allows users to remove `type: pkcs11` without also having to delete the `pkcs11` config block, making mode switching easier. A CEL rule requiring `type: pkcs11` when `pkcs11` is present would force users to clean up both fields atomically, which is worse UX.

internal/controller/ctlog/actions/deployment.go
```diff
@@ -202,6 +225,24 @@ func (i deployAction) ensureDeployment(instance *rhtasv1.CTlog, sa string, label
...
+	podSpec *core.PodSpec,
+	container *core.Container,
+) error {
+	pkcs11helpers.EnsureHSMResources(podSpec, container, instance.Spec.Volumes)
+	return nil
+}
+
+// cleanupPKCS11Resources removes operator-managed PKCS#11 volume mounts and volumes
+// when the signer type is not pkcs11 (e.g. switching back to file mode).
+func (i deployAction) cleanupPKCS11Resources(podSpec *core.PodSpec, container *core.Container) {
```
bouskaJ: use [OptionalToggle](https://github.com/securesign/secure-sign-operator/blob/main/internal/utils/kubernetes/ensure/toggleable.go#L22) with automatic cleanup

api/v1/ctlog_types.go
```diff
@@ -78,16 +78,37 @@ type CTlogSpec struct {
 }
 
 const CTlogSignerTypeFile = "file"
+const CTlogSignerTypePKCS11 = "pkcs11"
```
bouskaJ: Make `SignerType` constants reusable across all components

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -476,6 +476,107 @@ func podExtensionsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
...
+					Env: []core.EnvVar{
+						{Name: "TEST_VAR", Value: c.String(10)},
+					},
+					SecretMount: []rhtasv1.SecretKeySelector{
+						{LocalObjectReference: rhtasv1.LocalObjectReference{Name: "secret-" + c.String(5)}, Key: "key"},
+					},
+				}
+			} else {
+				s.Auth = nil
+			}
```
bouskaJ: AFAIK this is done by automatic fuzzer function - no need to set manually unless you need a specific format

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -476,6 +476,107 @@ func podExtensionsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
...
+				// PKCS#11 mode
+				s.Type = rhtasv1.FulcioSignerTypePKCS11
+				s.PKCS11 = &rhtasv1.FulcioPKCS11Config{
+					PKCS11Config: rhtasv1.PKCS11Config{
+						KeyID:    ptr.To(int32(c.Intn(100))),
+						KeyLabel: "key-" + c.String(5),
+					},
+					ConfigRef: &rhtasv1.SecretKeySelector{
+						LocalObjectReference: rhtasv1.LocalObjectReference{Name: "pkcs11-config-" + c.String(5)},
+						Key:                  "config.json",
```
bouskaJ: you use `c.FillNoCustom(s.PKCS)` instead of hardcoding values

bouskaJ (commented):
I’ve started the review, but I suggest to put this PR on hold until the foundation for the user-spec deletion is in place to avoid a massive refactor. I’ll try to help Jourdan or submit my own proposal today.

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -476,6 +476,107 @@ func podExtensionsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
 	}
 }
 
+// signerPKCS11FuzzerFuncs generates roundtrip-safe PKCS#11 config on FulcioSigner
+// and Auth on FulcioSpec. When PKCS#11 mode is chosen, File must be nil and vice-versa.
+func signerPKCS11FuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
```
osmman: The custom fuzzers `signerPKCS11FuzzerFuncs` and `ctlogPKCS11FuzzerFuncs` are not needed. The reason is that conversion does not contain any specific format for parsing and any input data even not valid will be converted. Roundtrip tests do not need valid data or configurations.

internal/controller/fulcio/actions/ensure_pkcs11_config.go
```diff
@@ -0,0 +1,163 @@
...
+			reconcile.TerminalError(fmt.Errorf("PKCS#11 configuration is required when signer type is pkcs11")),
+			instance,
+		)
+	}
+
+	pkcs11Cfg := instance.Spec.Signer.PKCS11
+
+	if pkcs11Cfg.ConfigRef == nil {
+		return a.Error(ctx,
+			reconcile.TerminalError(fmt.Errorf("spec.signer.pkcs11.configRef is required")),
```
osmman, line 64: Bug: this hard-fails with a TerminalError whenever configRef is nil, but the CRD XValidation rule (api/v1/fulcio_types.go) accepts the alternative structured form (modulePath+tokenLabel+pinSecretRef, no configRef). That mode is schema-valid but never implemented here - a CR using it passes admission and then permanently fails reconcile with a terminal error. Either implement the structured-config path or tighten the CRD validation to require configRef.

internal/controller/fulcio/actions/deployment.go
```diff
@@ -205,6 +209,13 @@ func (i deployAction) ensureFileCADeployment(instance *rhtasv1.Fulcio) func(depl
 		container := kubernetes.FindContainerByNameOrCreate(&dp.Spec.Template.Spec, containerName)
 		template := &dp.Spec.Template
 
+		// Clean up PKCS#11-specific volumes when switching back to file mode
+		kubernetes.RemoveVolumeByName(&template.Spec, PKCS11ConfigVolumeName)
+		kubernetes.RemoveVolumeByName(&template.Spec, PKCS11CertVolumeName)
+		kubernetes.RemoveVolumeMountByName(container, PKCS11ConfigVolumeName)
+		kubernetes.RemoveVolumeMountByName(container, PKCS11CertVolumeName)
+		pkcs11helpers.CleanupHSMResources(&template.Spec, container)
```
osmman, line 216: ensureFileCADeployment (switch back to file mode) cleans up PKCS#11 volumes/mounts but never clears the PKCS11Condition (FulcioPKCS11ConfigAvailable) status condition. CTLog's equivalent path removes it explicitly. Here it's left stale forever after switching signer type back to file.

internal/controller/fulcio/actions/deployment.go
```diff
@@ -310,3 +321,109 @@ func (i deployAction) ensureFileCADeployment(instance *rhtasv1.Fulcio) func(depl
...
+		// Clean up file-mode volumes that should not be present in PKCS#11 mode
+		kubernetes.RemoveVolumeByName(&template.Spec, "fulcio-cert")
+		kubernetes.RemoveVolumeMountByName(container, "fulcio-cert")
+
+		pkcs11Cfg := instance.Spec.Signer.PKCS11
+		configRef := pkcs11Cfg.ConfigRef
+
+		container.Args = append(container.Args,
+			"--ca=pkcs11ca",
+			fmt.Sprintf("--pkcs11-config-path=%s/%s", PKCS11ConfigMountPath, configRef.Key),
```
osmman, line 352: Potential nil-pointer panic: configRef is dereferenced here (and at L382/L385) assuming EnsurePKCS11ConfigAction already validated it's non-nil. If action ordering ever changes (or this ensure func is reached before that validation runs), this panics the reconciler instead of failing gracefully.

internal/controller/ctlog/actions/server_config.go
```diff
@@ -377,5 +428,22 @@ func (i serverConfig) configMatchingAnnotations(ctx context.Context, instance *r
 		annotations[labels.LabelNamespace+"/logPrefix"] = instance.Spec.Prefix
 	}
 
+	if instance.Spec.Signer.Type == rhtasv1.CTlogSignerTypePKCS11 && instance.Spec.Signer.PKCS11 != nil {
+		p := instance.Spec.Signer.PKCS11
+		pinData, pinErr := kubernetes.GetSecretData(ctx, i.Client, instance.Namespace, p.PinSecretRef)
```
osmman: This re-fetches the PIN and public-key secrets a second time purely to compute the comparison hash, right after buildPKCS11Config (L314/L322) already fetched the same secrets moments earlier in the same Handle() call. Doubles the API server load on every reconcile of a PKCS#11 CTLog instance, indefinitely.

internal/controller/ctlog/actions/server_config.go
```diff
@@ -377,5 +428,22 @@ func (i serverConfig) configMatchingAnnotations(ctx context.Context, instance *r
...
+		pubKeyData, pubErr := kubernetes.GetSecretData(ctx, i.Client, instance.Namespace, p.PublicKeyRef)
+		if pinErr == nil && pubErr == nil {
```
osmman: The pkcs11ContentHash annotation is only set if pinErr == nil && pubErr == nil - on a transient secret-read failure it's silently omitted rather than surfaced as an error. Since equality.Semantic.DeepDerivative treats a missing expected key as 'don't care', a real content change occurring during that transient failure window can go undetected, silently deferring a needed secret regeneration.

internal/controller/fulcio/actions/ensure_pkcs11_config.go
```diff
@@ -0,0 +1,163 @@
...
+	})
+
+	return a.ReturnOnChange(a.PersistStatus)(ctx, instance)
+}
+
+// computePKCS11Hash computes a deterministic SHA-256 hash of the PKCS#11 configuration
+// fields used for drift detection. This hash is stored in the PKCS11Condition message
+// and compared on subsequent reconciles to detect configuration changes without
+// reacting to unrelated generation bumps (e.g., monitoring toggle).
+func computePKCS11Hash(instance *rhtasv1.Fulcio) string {
```
osmman, line 144: computePKCS11Hash here and pkcs11SpecHash in internal/controller/ctlog/actions/ensure_pkcs11_config.go:165 are two independently hand-written SHA-256 drift-detection functions covering overlapping fields. Worth extracting one shared helper in internal/controller/common/pkcs11 so the hashing scheme can't silently diverge between Fulcio and CTLog.

api/v1/common.go
```diff
@@ -319,6 +319,36 @@ func (v *AdditionalVolume) ToVolume() core.Volume {
...
+	//+kubebuilder:validation:Pattern=`^/.+\..+$`
+	ModulePath string `json:"modulePath,omitempty"`
+	// Token label identifying the HSM slot.
+	//+optional
+	//+kubebuilder:validation:MinLength=1
+	TokenLabel string `json:"tokenLabel,omitempty"`
+	// Numeric slot ID (alternative to tokenLabel).
+	//+optional
+	//+kubebuilder:validation:Minimum=0
+	SlotNumber *int32 `json:"slotNumber,omitempty"`
```
osmman: SlotNumber is defined on the shared PKCS11Config type but is never read anywhere in the controller/config-generation code (grepped repo-wide, only this declaration and generated deepcopy reference it). Either wire it into the PKCS#11 config generation or drop it from the API until it's used.

internal/controller/fulcio/actions/deployment.go
```diff
@@ -90,6 +93,7 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1.Fulcio) *act
 		deployment.TrustedCA(instance.GetTrustedCA(), containerName),
 		deployment.PodRequirements(instance.Spec.PodRequirements, containerName),
 		deployment.PodSecurityContext(),
+		deployment.Auth(containerName, instance.Spec.Auth),
```
osmman: deployment.Auth(containerName, instance.Spec.Auth) is called twice in this same chain (also L88) - leftover duplicate from this PR's edit. Same issue in internal/controller/ctlog/actions/deployment.go L76/L82.

test/e2e/pkcs11/pkcs11_install_test.go
```diff
@@ -0,0 +1,165 @@
+//go:build pkcs11
```
osmman: Is there some test infrastructure prerequisite to have separate `pkcs11` tag and job?

osmman (changes requested):

tommyd450:
https://github.com/securesign/secure-sign-operator/pull/2194 is blocked by this pr. I can try and make changes but it might be smarter for me to wait for the changes here to be solidified.

api/v1/ctlog_types.go
sampras343: Done. All code now uses `SignerTypeFile` / `SignerTypePKCS11` directly from `api/v1/common.go`.

But I also notice the SignerType `kms` under `rekor_types.go` maybe we should streamline it in a different PR

internal/controller/ctlog/actions/deployment.go
sampras343: Done

api/v1alpha1/conversion_roundtrip_test.go
sampras343: done

api/v1alpha1/conversion_roundtrip_test.go
sampras343: Fixed. `podExtensionsFuzzerFuncs` now only constrains `AdditionalVolume` (needed for the MaxProperties=2 schema constraint). InitContainers, VolumeMounts, and all other fields use the automatic fuzzer.

api/v1alpha1/conversion_roundtrip_test.go
sampras343: Removed.

internal/controller/fulcio/actions/ensure_pkcs11_config.go
sampras343: Tightened the CRD. I feel for the initial release making configRef required and ensuring that user provides a valid one would be sufficient.
Auto creating this crypto11.conf based on other inputs seems like an overkill. It's a good to have. So can be extended as future feature.

Removed the Mode A alternative (modulePath+tokenLabel+pinSecretRef).
The controller's TerminalError on nil configRef is now consistent with the CRD validation and it can't be reached in normal operation.

internal/controller/fulcio/actions/deployment.go
sampras343: Fixed - added `meta.RemoveStatusCondition(&instance.Status.Conditions, PKCS11Condition)` in `ensureFileCADeployment` alongside the existing volume cleanup matching CTLog's pattern.

This is a temporary fix. These will be migrated to one of the potential clean up features worked on by @bouskaJ and @jourdee-lab .

This PR would be on hold till those features are merged and adapted here.

internal/controller/fulcio/actions/deployment.go
sampras343: Added nil guard for configRef before dereferencing. Returns error instead of panicking if action order changes. 
The CRD now also requires configRef, so this path can't be reached in normal operation. Still adopted defense in depth.

internal/controller/ctlog/actions/server_config.go
sampras343: Fixed
Replaced the content-based hash (which re-fetched PIN and public-key secrets) with pkcs11SpecHash, which hashes spec-level fields only (refnames, tokenLabel, modulePath) without any API calls. This matches the file-mode pattern where privateKeyRef annotation tracks the ref name/key, not secret content. Secret content drift is handled separately by ensurePKCS11Config's condition hash.

internal/controller/ctlog/actions/server_config.go
sampras343: The annotation now uses pkcs11SpecHash which reads spec-level fields only, never makes API calls, and never fails. The silent-omission bug is eliminated.

internal/controller/fulcio/actions/ensure_pkcs11_config.go
sampras343: Fixed.
Extracted pkcs11helpers.HashCoreConfig(h, cfg) in `internal/controller/common/pkcs11`.
Both fulcio and ctlog now call the shared heklper for core hashing `PKCS11Config` fields and also append their component specific fields.

api/v1/common.go
sampras343: SlotNumber is intentionally part of the shared PKCS11Config for forward compatibility. It's not wired into the controller today because CTLog's `keyspb.PKCS11Config` proto doesn't have a slot number field, and Fulcio currently only supports Mode B (`configRef`) where the user provides the full crypto11 JSON.

Where is `SoltNumber` used? Usually an alternative for tokenLabel in HSM devices like Thales Luna.

In future - It will be consumed when Fulcio Mode A (structured config generation) is implemented where crypto11 supports slot-based selection for HSMs like Thales Luna where partition slot IDs are preferred over token labels.

Do you want me to drop this for the time being and re introduce later? @osmman

internal/controller/fulcio/actions/deployment.go
sampras343: Removed duplicate call in both files

test/e2e/pkcs11/pkcs11_install_test.go
sampras343: The pkcs11 build tag separates these tests because they require SoftHSM infrastructure (PVCs for token persistence, key ceremony Jobs, derived Secrets) that must be set up before the test runs. The standard E2E environment doesn't have these prerequisites. In CI, the PKCS#11 tests are expected to run as a separate parallel pipeline task (run-tas-pkcs11-e2e) that first deploys the prerequisites, then runs the tests.

The tests are being performed in this PR https://github.com/securesign/secure-sign-operator/pull/2227

api/v1/common.go
osmman: yes please

test/e2e/pkcs11/pkcs11_install_test.go
osmman: Looked into this, I don't think a separate tag/pipeline is needed. All SoftHSM prerequisites (ConfigMap, Secret, PVCs, ceremony Jobs) are created in the test's own namespace via the standard `steps.CreateNamespace`, same as every other e2e test. Nothing cluster-scoped, and the PVCs are plain RWO so kind's default provisioner handles them. That's different from `ha` (needs `nfs-csi: true`) or `custom_install` (different install method), which do change cluster-wide behavior.

We already have precedent for this pattern without a dedicated tag: `byodb_test.go` stands up an external Postgres and still runs under `integration`. Suggest folding this into the standard e2e run rather than a new tag/job.

api/v1/common.go
sampras343: Done

test/e2e/pkcs11/pkcs11_install_test.go
sampras343: done

.github/workflows/main.yml
```diff
@@ -252,11 +252,19 @@ jobs:
...
+
       - name: Load Docker image into Kind cluster
         run: |
           podman save ${{ env.IMG }} -o operator-oci.tar
           kind load image-archive operator-oci.tar
 
+      - name: Load SoftHSM init image into Kind cluster
+        run: |
+          podman save quay.io/securesign/softhsm-init:1.0-test -o softhsm-init-oci.tar
+          kind load image-archive softhsm-init-oci.tar
```
osmman: I think that pulling and pushing `quay.io/securesign/softhsm-init:1.0-test` image is unnecessary. Kubernetes cluster should pull it directly from quay.io registry without problem.

.tekton/rhtas-operator-bundle-pull-request.yaml
```diff
@@ -46,7 +46,7 @@ spec:
   - name: manager-registry-url
     value: registry.redhat.io/rhtas/rhtas-rhel9-operator
   - name: fips-check
-    value: "true"
+    value: "false"
```
osmman: Why?

sampras343: reverted. Was testing some e2e tests. Overlooked this while reverting

.github/workflows/main.yml
sampras343: Done

api/v1/fulcio_types.go
```diff
@@ -43,11 +41,25 @@ type FulcioSpec struct {
 	Auth *Auth `json:"auth,omitempty"`
 }
 
+// FulcioPKCS11Config holds the Fulcio PKCS#11/HSM signer configuration.
+// +kubebuilder:validation:XValidation:rule="has(self.configRef)",message="configRef is required for Fulcio PKCS#11 signer"
+// +kubebuilder:validation:XValidation:rule="has(self.keyID)",message="keyID is required for Fulcio PKCS#11 signer"
+type FulcioPKCS11Config struct {
+	PKCS11Config `json:",inline"`
```
bouskaJ: PKCS11Config embeds  ModulePath, TokenLabel, PinSecretRef, KeyLabel,... but Fulcio's PKCS#11 mode only uses ConfigRef + KeyID. Are we going to use all of them in future? If not, we should remove them from the shared struct

internal/controller/ctlog/actions/server_config.go
```diff
@@ -40,6 +40,7 @@ var serverConfigAnnotations = []string{
 	labels.LabelNamespace + "/rootCertificatesHash",
 	labels.LabelNamespace + "/privateKeyRef",
 	labels.LabelNamespace + "/logPrefix",
+	labels.LabelNamespace + "/pkcs11SpecHash",
```
bouskaJ, line 43: inconsistent - you register `pkcs11SpecHash` but write `labels.LabelNamespace+"/pkcs11ContentHash"`

api/v1/common.go
```diff
@@ -319,6 +324,32 @@ func (v *AdditionalVolume) ToVolume() core.Volume {
...
+	//+optional
+	//+kubebuilder:validation:MinLength=1
+	TokenLabel string `json:"tokenLabel,omitempty"`
+	// Reference to a Secret key containing the HSM user PIN.
+	//+optional
+	PinSecretRef *SecretKeySelector `json:"pinSecretRef,omitempty"`
+	// PKCS#11 CKA_ID of the signing key.
+	//+optional
+	//+kubebuilder:validation:Minimum=0
+	KeyID *int32 `json:"keyID,omitempty"`
```
bouskaJ: AI review:  PKCS11Config.KeyID should be *int64, not *int32 — PKCS#11 CKA_ID has no inherent size limit. int32 risks silent truncation. This is an API field that can't be changed without a version bump once shipped.

internal/controller/ctlog/actions/ensure_pkcs11_config.go
```diff
@@ -0,0 +1,176 @@
...
+
+	c := meta.FindStatusCondition(instance.Status.Conditions, PKCS11Condition)
+	if c == nil || c.Status != metav1.ConditionTrue {
+		return true
+	}
+
+	// Drift detection: compare current spec hash against stored hash.
+	currentHash := pkcs11SpecHash(instance.Spec.Signer.PKCS11)
+	return !strings.Contains(c.Message, fmt.Sprintf("[hash=%s]", currentHash))
```
bouskaJ: Hash-in-condition-message is fragile — Storing config hash in the condition Message field and parsing it back with strings.Contains is brittle. Use an annotation

api/v1/fulcio_types.go
sampras343: CTLog uses all of them. `modulePath`, `tokenLabel`, and `pinSecretRef` are required fields for CTLog's PKCS#11 config. They feed directly into the `keyspb.PKCS11Config` protobuf that CT-Go consumes.

sampras343: For Fulcio, we initially thought to go about in two Modes.
Mode A:
- User provides all these information inline
- User does not provide a `configRef`
- Operator creates the necessary configs based on the inline information

Mode B:
- User provides only `configRef` with all the details
- Use user provided information directly.

For this Release only Mode B is implemented and all the responsibility rely on user to provide the right config.
You are right in noticing that these fields are dormant in Fulcio right now.

internal/controller/ctlog/actions/server_config.go
sampras343: Fixed. Consistent `pkcs11SpecHash` used throughout.

api/v1/common.go
sampras343: Fixed

internal/controller/ctlog/actions/ensure_pkcs11_config.go
sampras343: Fixed to use annotations

api/v1/fulcio_types.go
bouskaJ: As we are going to release this feature as a TP we have two options:
- we can document that those fields does not work (on Fulcio)
- do not add them to the shared struct and add them later once implemented. Adding new fields is an allowed operation on the API files

bouskaJ:
once you address https://github.com/securesign/secure-sign-operator/pull/2182#discussion_r3851067806 LGTM

api/v1/fulcio_types.go
sampras343: @bouskaJ `PKCS11Config.KeyID` is used in place of `--hsm-caroot-id` for Fulcio Deployment.

The keyID is defined in the shared spec.

sampras343: To keep it consistent with CTLog we can either include the `PKCS11Config` here and make use of keyID field.

Or remove the `PKCS11Config` and use only keyID field as a part of `FulcioPKCS11Config` 
What do you suggest?

bouskaJ: we can keep shared `PKCS11Config` and start with one entry `keyID` for now. Once more fields become shared, we can shuffle them.

sampras343: All the other fields from the shared `PKCS11Config` are used by `CTLog`

So the fields of  `PKCS11Config`:
|Field |Fulcio |CTLog|
|---|---|---|
|ModulePath|No|Yes|
|PinSecretRef|No|Yes|
|TokenLabel|No|Yes|
|KeyID|Yes|No|
|KeyLabel|Optional|Optional|

`KeyLabel` is an optional alternative for `KeyID`
The above fields in the shared struct are applicable for either of them depending on how the user wants to configure

sampras343: Decision:
CTlogPKCS11Config — owns ModulePath, TokenLabel, PinSecretRef, PublicKeyRef
FulcioPKCS11Config — owns ConfigRef, KeyID

Remove PKCS11Config completely and reintroduce later when needed

sampras343: Addressed

sampras343:
> once you address [#2182 (comment)](https://github.com/securesign/secure-sign-operator/pull/2182#discussion_r3851067806) LGTM

Addressed

osmman (approved):

PR #2184, opened 2026-08-05
feat: add KMS signer support for Fulcio (SECURESIGN-4462)

kdacosta0:
## What changed

Adds KMS signer support for Fulcio — operators can now configure Fulcio to use a remote KMS key (AWS, GCP, Azure, HashiCorp Vault) instead of a local file-based ECDSA key.

### API
- `FulcioSigner.Kms` field (`type: kms` + `kms.keyResource` URI)
- 5 XValidation rules enforcing mutual exclusion between file and kms (with `has(self.type)` guards for omitempty safety)
- `certificateChainRef` required for KMS (Fulcio server hard-requires `--kms-cert-chain-path`)

### Controller
- `resolve_kms_signer` action — resolves cert chain secret, sets CertCondition
- `generate_signer` skipped for KMS (`IsEnabled` guard: `type == file || type == ""`)
- Deployment: `--ca=kmsca` + `--kms-resource` + `--kms-cert-chain-path` args, cert volume with CA only (no private key)
- Shared volume/mount setup extracted into `ensureSignerVolumes` helper (eliminates duplication between file and KMS deployment functions)
- FIPS validation skips private key check for file signer only (positive type check)
- Stale `PASSWORD` env var cleaned up on File→KMS switch

### Conversion (v1 ↔ v1alpha1)
- Mutual exclusion guard: KMS branch nils File, file branch nils Kms
- Auth restored unconditionally (orthogonal to signer type)
- Fuzzer constrained to valid states (file, kms, or empty/default type)

## Why

SECURESIGN-4462 — customers need remote key management for Fulcio CA signing keys.

## Tested
- Unit tests: conversion roundtrip fuzz (incl. `Type=""`), deployment args, File→KMS PASSWORD cleanup, resolve_kms_signer, generate_signer guard, XValidation CEL
- Cluster: 12 manual tests (XValidation rejection, KMS reconcile, conversion roundtrip read+write, type switching, file signer regression)
- AWS KMS: cosign sign with real RSA_2048 key (eu-north-1) on OpenShift — tlog entry created with index: 0

## Stack

Depends on: #2181 (Auth support)
Parallel with: #2182 (PKCS#11) — no code dependency, either can merge first.


kdacosta0:
Waiting for #2181 to merge

api/v1/fulcio_types.go
```diff
@@ -12,6 +12,7 @@ import (
 )
 
 const FulcioSignerTypeFile = "file"
+const FulcioSignerTypeKMS = "kms"
```
bouskaJ: We have similar types on all constants (supporting KMS) can we have share them?

internal/controller/fulcio/actions/constants.go
```diff
@@ -11,7 +11,8 @@ const (
 	ServiceMonitorName = "fulcio-metrics"
 	RBACName           = "fulcio"
 
-	CertCondition = "FulcioCertAvailable"
+	CertCondition  = "FulcioCertAvailable"
+	ReasonResolved = "Resolved"
```
bouskaJ: this reason is used in all components. If you want to make it constant you should use shared constant (and unify constant other reasons for unification)

internal/controller/fulcio/actions/deployment.go
```diff
@@ -80,6 +80,8 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1.Fulcio) *act
 		i.ensureCommonDeployment(instance, RBACName, labels, ctlogUrl),
 		ensure.Optional(instance.Spec.Signer.Type == rhtasv1.FulcioSignerTypeFile || instance.Spec.Signer.Type == "",
 			i.ensureFileCADeployment(instance)),
+		ensure.Optional(instance.Spec.Signer.Type == rhtasv1.FulcioSignerTypeKMS,
```
bouskaJ: Use OptionalToogable with cleanup https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/rekor/actions/server/deployment.go#L168

internal/controller/fulcio/actions/resolve_kms_signer.go
```diff
@@ -0,0 +1,109 @@
...
+		Status:             metav1.ConditionTrue,
+		Reason:             ReasonResolved,
+		Message:            "Using existing secret",
+		ObservedGeneration: instance.GetGeneration(),
+	})
+
+	return a.ReturnOnChange(a.PersistStatus)(ctx, instance)
+}
+
+func (a *resolveKMSSignerAction) ensureCertChainLabel(ctx context.Context, instance *rhtasv1.Fulcio, ref *rhtasv1.SecretKeySelector) error {
```
bouskaJ, line 83: - Label management (ensureCertChainLabel vs ensureLabelsOnSecret) — both list secrets by label, remove from non-targets, apply to target. The framework version is more robust.

internal/controller/fulcio/actions/deployment.go
```diff
@@ -186,6 +194,55 @@ func (i deployAction) ensureCommonDeployment(instance *rhtasv1.Fulcio, sa string
...
+						Path: "ca.crt",
+						Mode: ptr.To(int32(0666)),
+					},
+				},
+			},
+		},
+	}
+
+	if fips.Enabled() {
+		container.Args = append(container.Args, "--client-signing-algorithms", fips.ClientSigningAlgorithms)
```
bouskaJ: It is called `ensureSignerVolumes` but it configures FIPS alhorithms

internal/controller/fulcio/actions/deployment.go
```diff
@@ -285,28 +320,64 @@ func (i deployAction) ensureFileCADeployment(instance *rhtasv1.Fulcio) func(depl
...
+		template := &dp.Spec.Template
+
+		// Clean up stale file-mode env var from a previous reconcile (e.g. File→KMS switch)
+		container.Env = slices.DeleteFunc(container.Env, func(e core.EnvVar) bool {
+			return e.Name == "PASSWORD"
+		})
+
+		container.Args = append(container.Args,
+			"--ca=kmsca",
+			fmt.Sprintf("--kms-resource=%s", instance.Spec.Signer.Kms.KeyResource),
```
bouskaJ: Inconsistent CLI arg formatting — --kms-resource=<val> uses = syntax while --kms-cert-chain-path <val> uses space-separated.

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -475,6 +475,34 @@ func podExtensionsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
 	}
 }
 
+// fulcioSignerFuzzerFuncs constrains FulcioSigner to valid discriminated union states:
+// file (with optional File struct), KMS (with Kms), or empty/default type.
+func fulcioSignerFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
+	return []interface{}{
+		func(s *rhtasv1.FulcioSigner, c randfill.Continue) {
+			c.FillNoCustom(&s.CertificateChain)
```
bouskaJ: Fuzzer doesn't enforce XValidation invariants

  fulcioSignerFuzzerFuncs fills CertificateChain before choosing the signer type and never clears the opposing union arm, so it can generate states that violate the new XValidation rules:

  - Rule 1: type=file/"" + certificateChainRef requires file.privateKeyRef — but CertificateChain is pre-filled and cases 0/2 may skip File or get nil PrivateKeyRef
  - Rule 3: type=kms requires certificateChainRef — but FillNoCustom can leave the pointer nil
  - Rule 4: type!=kms requires kms absent — but cases 0/2 never clear s.Kms
  - Rule 5: type=kms requires file absent — but case 1 never clears s.File

  The fix is the same pattern rekorSignerFuzzerFuncs already uses — choose the type first, nil the opposing arm, fill fields after, and enforce cross-field dependencies

api/v1alpha1/fulcio_conversion.go
```diff
@@ -143,11 +143,17 @@ func (src *Fulcio) ConvertTo(dstRaw conversion.Hub) error {
...
-			dst.Spec.Signer.File = &rhtasv1.FulcioFile{}
+	if dst.Spec.Signer.Type == rhtasv1.FulcioSignerTypeKMS {
+		dst.Spec.Signer.File = nil
+		dst.Spec.Signer.Kms = restored.Spec.Signer.Kms
+	} else {
+		dst.Spec.Signer.Kms = nil
+		// If original v1 had File=&{} (empty struct), preserve it
+		if dst.Spec.Signer.File == nil && restored.Spec.Signer.File != nil {
+			emptyFile := &rhtasv1.FulcioFile{}
+			if equality.Semantic.DeepEqual(restored.Spec.Signer.File, emptyFile) {
```
bouskaJ: If auto-conversion couldn't produce a File (nil), just restore whatever the annotation had — no reason to guard on empty struct:

  } else {
      dst.Spec.Signer.Kms = nil
      if dst.Spec.Signer.File == nil {
          dst.Spec.Signer.File = restored.Spec.Signer.File
      }
  }

  The DeepEqual check against an empty struct is overly cautious and would silently drop a non-empty File from the annotation if one ever appeared.

bouskaJ (changes requested):

api/v1/fulcio_types.go
kdacosta0: @bouskaJ regarding signer types constants, I want to be sure I get the scope right before making changes.

Currently we have duplications here:
- Fulcio: File/KMS
- Rekor: Secret/Memory/KMS
- CTLog: File


Some string values overlap like `file` in Fulcio+CTLog, and `kms` in Fulcio+Rekor

I'd move all of them in (`file`, `kms`, `secret`, `memory`) in `common.go` and have all components share the constants. Is that what you had in mind?

Also, TSA uses  [structs](https://github.com/securesign/secure-sign-operator/blob/main/api/v1/timestampauthority_types.go#L55-L68) (`File *File`, `Kms *KMS`, `Tink *Tink`) instead of a Type string enum. Should we unify that in a followup PR?

bouskaJ (commented):
We already have a generic signer implementation by @osmman Can we expand it with KMS. If not, will your extension work correctly with it?

@osmman I guess you are more eligible for this PR reveiw than me.

api/v1alpha1/fulcio_conversion.go
```diff
@@ -143,11 +142,13 @@ func (src *Fulcio) ConvertTo(dstRaw conversion.Hub) error {
...
-			dst.Spec.Signer.File = &rhtasv1.FulcioFile{}
+	if dst.Spec.Signer.Type == rhtasv1.SignerTypeKMS {
+		dst.Spec.Signer.File = nil
+		dst.Spec.Signer.Kms = restored.Spec.Signer.Kms
+	} else {
+		dst.Spec.Signer.Kms = nil
+		if dst.Spec.Signer.File == nil {
+			dst.Spec.Signer.File = restored.Spec.Signer.File
 		}
 	}
```
osmman: This branches on `SignerTypeKMS` to decide whether to clear `File` and restore `Kms`, or clear `Kms` and conditionally restore `File`. That's a validation rule ("File and Kms are mutually exclusive"), not a data conversion - the converter shouldn't need to know it.

Concrete failure: object has `Type=kms`, `Kms` set. User edits it via v1alpha1 by setting `Certificate.PrivateKeyRef` (the only way v1alpha1 can express a signer). The primary conversion correctly derives `File` from that edit, but this block unconditionally restores `Type=kms` from the annotation, nils `File`, and restores the stale `Kms` - the user's edit is silently reverted with no error.

Suggested fix - restore each field independently, whenever the primary conversion left it nil, with no cross-field clearing:
```go
dst.Spec.Signer.Type = restored.Spec.Signer.Type
if dst.Spec.Signer.File == nil {
    dst.Spec.Signer.File = restored.Spec.Signer.File
}
if dst.Spec.Signer.Kms == nil {
    dst.Spec.Signer.Kms = restored.Spec.Signer.Kms
}
```
`Type` and `Kms` are pure hub-only fields (v1alpha1 can never populate them - the primary conversion always writes `Type=file` as a placeholder and never touches `Kms`), so restoring them unconditionally/when-nil is always safe and requires no knowledge of which combinations are valid. If XValidation ever allows `File` and `Kms` together, this already preserves both losslessly instead of needing another rewrite.

api/v1alpha1/securesign_conversion.go
```diff
@@ -52,11 +52,17 @@ func (src *Securesign) ConvertTo(dstRaw conversion.Hub) error {
...
+	} else {
+		dst.Spec.Fulcio.Signer.Kms = nil
+		// If original v1 had File=&{} (empty struct), preserve it
+		if dst.Spec.Fulcio.Signer.File == nil && restored.Spec.Fulcio.Signer.File != nil {
+			emptyFile := &rhtasv1.FulcioFile{}
+			if equality.Semantic.DeepEqual(restored.Spec.Fulcio.Signer.File, emptyFile) {
+				dst.Spec.Fulcio.Signer.File = &rhtasv1.FulcioFile{}
+			}
 		}
 	}
```
osmman: Same issue as `fulcio_conversion.go` for the embedded Fulcio signer - branches on `SignerTypeKMS` and clears one field based on the other's presence, so an explicit v1alpha1 edit that adds file-based signer data on a previously-KMS Fulcio silently reverts to the stale KMS config.

Same fix applies: restore `Type`, `File`, `Kms` independently (each only when the primary conversion left it nil), no cross-field clearing.

api/v1alpha1/conversion_roundtrip_test.go
```diff
@@ -475,6 +475,43 @@ func podExtensionsFuzzerFuncs(_ runtimeserializer.CodecFactory) []interface{} {
...
+						s.File.PrivateKeyRef = &rhtasv1.SecretKeySelector{}
+						c.FillNoCustom(s.File.PrivateKeyRef)
+					}
+				} else if c.Bool() {
+					s.File = &rhtasv1.FulcioFile{}
+					c.FillNoCustom(s.File)
+				}
+			}
+		},
+	}
```
osmman: This custom fuzzer's `Type`-first branching and the `CertificateChainRef`-conditional `File` synthesis exist only to avoid fuzzing `File`/`Kms` combinations that the current buggy `ConvertTo` mishandles - it's a symptom of the conversion bug flagged above, not an independent need.

Once `fulcio_conversion.go`/`securesign_conversion.go` restore `Type`, `File`, and `Kms` independently with no cross-field clearing, every combination (including both `File` and `Kms` set) round-trips losslessly through the annotation. At that point this function and its two registrations (lines 539 and 591) can be deleted - plain generic fill handles `FulcioSigner` correctly.

api/v1alpha1/fulcio_conversion.go
osmman, line 150: **Bug**: this branches on `SignerTypeKMS` to decide how to restore `File`/`Kms` - that is a validation rule (mutual exclusivity), and the converter should not know or enforce it.

**Concrete break**: object has `Type=kms`. A user edits it via v1alpha1 by setting `Certificate.PrivateKeyRef` - the only signer field v1alpha1 can express. This block ignores that edit: it restores the stale `Type=kms`/`Kms` from the annotation and nils out the `File` the edit just produced. The user's change is silently discarded, no error.

**Do this instead** - restore each field independently, only when the primary conversion left it nil, no branching on type:
```go
dst.Spec.Signer.Type = restored.Spec.Signer.Type
if dst.Spec.Signer.File == nil {
    dst.Spec.Signer.File = restored.Spec.Signer.File
}
if dst.Spec.Signer.Kms == nil {
    dst.Spec.Signer.Kms = restored.Spec.Signer.Kms
}
```
`Type` and `Kms` are hub-only fields v1alpha1 can never set, so restoring them unconditionally/when-nil is always safe - no need to know which `Type`/`File`/`Kms` combinations are valid.

api/v1alpha1/securesign_conversion.go
osmman: **Same bug as `fulcio_conversion.go`**, embedded Fulcio signer.

**Do this**: replace this whole block with the same independent per-field restore:
```go
dst.Spec.Fulcio.Signer.Type = restored.Spec.Fulcio.Signer.Type
if dst.Spec.Fulcio.Signer.File == nil {
    dst.Spec.Fulcio.Signer.File = restored.Spec.Fulcio.Signer.File
}
if dst.Spec.Fulcio.Signer.Kms == nil {
    dst.Spec.Fulcio.Signer.Kms = restored.Spec.Fulcio.Signer.Kms
}
```
Delete the `SignerTypeKMS` branch and the `emptyFile` special-case (lines 60-66) with it - `restored.Spec.Fulcio.Signer.File` already *is* that empty struct when that was the original value, so assigning it directly makes the special-case dead code.

api/v1alpha1/conversion_roundtrip_test.go
osmman: **Delete this function - it is compensating for the bug above, not testing a real constraint.**

The `Type`-first branching and the `CertificateChainRef`-conditional `File` synthesis here exist only so the fuzzer never generates a `File`/`Kms` combination that the current, broken `ConvertTo` mishandles.

Once `fulcio_conversion.go` and `securesign_conversion.go` restore `Type`/`File`/`Kms` independently (see the other two comments), every combination - including both `File` and `Kms` set - round-trips losslessly through the annotation. There is nothing left for a custom fuzzer to guard against.

**Do this**: delete `fulcioSignerFuzzerFuncs` and its two registrations in the `FuzzerFuncs` lists below. Plain generic fill already handles `FulcioSigner` correctly.

api/v1alpha1/securesign_conversion.go
kdacosta0: We should also fix this in [CTlog signer](https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/securesign_conversion.go#L77-L83)
@osmman

osmman: take that code snipped like example. The logic from CTlog has been introduced to preserve information about empty struct in v1 api.

api/v1alpha1/fulcio_conversion.go
kdacosta0: fixed

api/v1alpha1/securesign_conversion.go
kdacosta0: fixed

api/v1alpha1/conversion_roundtrip_test.go
kdacosta0: fixed

kdacosta0:
Rebased on `main` now that #2182 (PKCS#11/HSM signer support) is merged, and resolved the resulting conflicts across the Fulcio/CTLog signer types (CEL validation rules, deployment ensure logic, generated CRDs/deepcopy).

Also added cleanup: switching a Fulcio signer directly from `pkcs11` → `kms` now removes the leftover PKCS#11 volumes/mounts, HSM resources, and the stale `PKCS11Condition` status - **mirroring the cleanup that already existed for `pkcs11` → `file`.**

api/v1alpha1/ctlog_conversion.go
```diff
@@ -78,12 +77,8 @@ func (src *CTlog) ConvertTo(dstRaw conversion.Hub) error {
...
 	dst.Spec.TrustedCA = restored.Spec.TrustedCA
 	dst.Spec.Signer.Type = restored.Spec.Signer.Type
-	// If original v1 had File=&{} (empty struct), preserve it
-	if dst.Spec.Signer.File == nil && restored.Spec.Signer.File != nil {
-		emptyFile := &rhtasv1.CTlogFile{}
-		if equality.Semantic.DeepEqual(restored.Spec.Signer.File, emptyFile) {
-			dst.Spec.Signer.File = &rhtasv1.CTlogFile{}
-		}
+	if dst.Spec.Signer.File == nil {
+		dst.Spec.Signer.File = restored.Spec.Signer.File
```
osmman: Why are you editing Ctlog?

internal/controller/fulcio/actions/deployment.go
osmman: Duplicate `--client-signing-algorithms` in PKCS#11 + FIPS mode.

This PR moved the FIPS arg into `ensureCommonDeployment` (runs for all signer types) and removed it from `ensureFileCADeployment`, but the copy here in `ensurePKCS11Deployment` was left in. Since both run on the same container, a PKCS#11 deployment with FIPS enabled gets --client-signing-algorithms twice. Please delete these lines:

api/v1alpha1/ctlog_conversion.go
kdacosta0: Wanted to make Ctlog conversion consistent with Fulcio and Rekor. But its genuinly outside of Fulcio KMS scope, will revert it as it's untested by this PR and out of scope

internal/controller/fulcio/actions/deployment.go
kdacosta0: fixed

osmman (approved):

bouskaJ (approved):

PR #2186, opened 2026-08-06
feat: add openbao:// KMS URI support (SECURESIGN-5076)

kdacosta0:
## What changed

- Add `openbao://` as a valid KMS URI scheme for Rekor and TSA signer configuration
- Move shared `KMS` struct from `timestampauthority_types.go` to `common.go` — it is used by both TSA and Fulcio

## Why

SECURESIGN-5076 — OpenBao (open-source Vault fork) is already supported by `sigstore/sigstore` as an alternative scheme for the hashivault KMS provider. The operator's CEL validation was rejecting valid `openbao://` URIs.

`KMS` struct lived in TSA types despite being shared across components — moved to `common.go` alongside `Auth`, `SecretKeySelector`, and other shared types for discoverability.

## Note

PR #2184 (Fulcio KMS signer) uses the `KMS` struct — after this merges, #2184 needs a trivial rebase to pick up the new location in `common.go`. Same package, no import changes.


bouskaJ (approved):
@kdacosta0 is it tested?

api/v1/common.go
```diff
@@ -232,6 +232,15 @@ type Pvc struct {
 	AccessModes []PersistentVolumeAccessMode `json:"accessModes,omitempty"`
 }
 
+// KMS configures a remote key management service for signing operations.
+// Used by TSA and Fulcio signer backends.
+// XValidation is also present on RekorSigner.KMS (string field) in rekor_types.go.
```
bouskaJ: do not comment where it is used and where it is validated - the comment is propagated into API desc.

api/v1/rekor_types.go
```diff
@@ -109,7 +109,8 @@ type RekorSigner struct {
 	//   - azurekms://keyname
 	//   - gcpkms://keyname
 	//   - hashivault://keyname
-	// +kubebuilder:validation:XValidation:rule="self == 'secret' || self == 'memory' || self.matches('^awskms://.+$') || self.matches('^gcpkms://.+$') || self.matches('^azurekms://.+$') || self.matches('^hashivault://.+$')",message="KMS must be 'secret', 'memory', or a valid URI with a key path (e.g., awskms:///key-id)"
+	//   - openbao://keyname
+	// +kubebuilder:validation:XValidation:rule="self == 'secret' || self == 'memory' || self.matches('^awskms://.+$') || self.matches('^gcpkms://.+$') || self.matches('^azurekms://.+$') || self.matches('^hashivault://.+$') || self.matches('^openbao://.+$')",message="KMS must be 'secret', 'memory', or a valid URI with a key path (e.g., awskms:///key-id)"
 	KMS string `json:"kms,omitempty"`
```
bouskaJ: can we reuse the KMS struct instead of the plain string?

bouskaJ (changes requested):

api/v1/rekor_types.go
osmman: It will be good to migrate Rekor to use same structures like TSA, CTlog and Fulcio for signer config

kdacosta0:
> @kdacosta0 is it tested?

@bouskaJ Yes, tested on ocp - signing and verification was successful

api/v1/rekor_types.go
```diff
@@ -95,35 +95,41 @@ type RekorAttestations struct {
...
 	// Reference to the signer private key.
-	//
-	// Optional field. When KMS is set to "secret", this field can be left empty, in which case the operator will automatically generate a signer key.
+	// When type is "secret", this field can be left empty — the operator will automatically generate a signer key.
 	// +optional
 	KeyRef *SecretKeySelector `json:"keyRef,omitempty"`
+
+	// Authentication configuration for the signer backend.
+	//+optional
+	Auth *Auth `json:"auth,omitempty"`
```
bouskaJ: We already have the top-level Auth Do not duplicate it

api/v1/common.go
```diff
@@ -232,6 +232,14 @@ type Pvc struct {
 	AccessModes []PersistentVolumeAccessMode `json:"accessModes,omitempty"`
 }
 
+// KMS configures a remote key management service for signing operations.
+// +kubebuilder:validation:XValidation:rule="self.keyResource.matches('^(gcpkms|azurekms|hashivault|openbao|awskms)://.+$')",message="keyResource must be a valid KMS URI (gcpkms://, azurekms://, hashivault://, openbao://, or awskms://)"
```
bouskaJ, line 158: ISSUE 1 (Must Fix): v1alpha1 CEL rejects openbao:// at admission

  File: /home/jbouska/git/operator/api/v1alpha1/rekor_types.go, line 121

  The v1alpha1 CEL validation rule:
  self == '' || self == 'secret' || self == 'memory' || self.matches('^awskms://.+$') || self.matches('^gcpkms://.+$') || self.matches('^azurekms://.+$') || self.matches('^hashivault://.+$')

  This does NOT include openbao://. The conversion round-trip test (TestRekorAuthRestorationRoundtrip) tests the Go conversion code, but not the CRD admission.

  The failure scenario:
  1. User creates a v1 Rekor with type: kms, kms.keyResource: "openbao://key". Passes v1 CEL. Stored as v1 (storage version).
  2. Someone reads it via v1alpha1 API. Conversion webhook converts it. The GET response has kms: "openbao://key". No validation on read.
  3. User updates any field via v1alpha1 API (e.g., replicas). The v1alpha1 CRD validation runs on the full object. CEL rejects kms: "openbao://key". The update fails.

  This creates an object that is readable but un-updatable via v1alpha1. It is not just a theoretical concern -- any tooling (Argo CD, Flux, Helm, even kubectl apply) that operates on the v1alpha1 version will
  break.

api/v1/rekor_types.go
```diff
@@ -95,35 +95,41 @@ type RekorAttestations struct {
...
+	RekorSignerTypeSecret = "secret"
+	RekorSignerTypeMemory = "memory"
+	RekorSignerTypeKMS    = "kms"
+)
+
+// RekorSigner defines the signer configuration for the Rekor transparency log.
+// +kubebuilder:validation:XValidation:rule="!has(self.type) || self.type != 'kms' || has(self.kms)",message="kms is required when type is 'kms'"
+// +kubebuilder:validation:XValidation:rule="!has(self.type) || self.type != 'secret' || !has(self.kms)",message="kms should not be configured when type is 'secret'"
+// +kubebuilder:validation:XValidation:rule="!has(self.type) || self.type != 'memory' || !has(self.kms)",message="kms should not be configured when type is 'memory'"
 type RekorSigner struct {
```
bouskaJ, line 110: ISSUE 3 (Consider): Missing CEL rules for keyRef/passwordRef when type is not secret

  File: /home/jbouska/git/operator/api/v1/rekor_types.go, lines ~100-127

  The current CEL rules validate kms presence:
  type != 'kms'    || has(kms)     // kms required when type=kms
  type != 'secret' || !has(kms)    // kms forbidden when type=secret
  type != 'memory' || !has(kms)    // kms forbidden when type=memory

  But there is no validation preventing nonsensical combinations:
  - type: kms with keyRef set (keyRef is a Kubernetes Secret reference for the local signer, meaningless for KMS)
  - type: kms with passwordRef set (same reasoning)
  - type: memory with keyRef or passwordRef (memory signer is ephemeral)

  Since v1 is greenfield, this is the time to get the validation right. The old v1 API also didn't validate these, but the old API had an overloaded string field that made such rules harder to express. Now with
  the discriminated union, adding these rules is straightforward:

  !has(self.type) || self.type != 'kms'    || !has(self.keyRef)
  !has(self.type) || self.type != 'kms'    || !has(self.passwordRef)
  !has(self.type) || self.type != 'memory' || !has(self.keyRef)
  !has(self.type) || self.type != 'memory' || !has(self.passwordRef)

  This prevents users from setting fields the operator will ignore, reducing confusion.

bouskaJ (changes requested):

bouskaJ (approved):

PR #2188, opened 2026-08-06
fix: update TUF URL resolution logic and adjust tests for external ingress handling

jourdee-lab:
**Summary**

The FIPS e2e "Console is running" test was timing out after 180s. The console-api and console-ui pods were stuck in PodInitializing with their init containers never completed, so the Console CR never reached a Ready state.

**Root Cause**

The console-api's wait-for-tuf init container curls the TUF URL in an unbounded loop (no retry limit/exit). With commit [ff65bdef](https://github.com/securesign/secure-sign-operator/commit/ff65bdef3b1b18002f7af3fd44a485ec2d459142) accidentally dropped the fallback that used the in-cluster address so when no explicit TUF ref is set on the Console CR, making it always autodiscover TUF's external URL instead. Since the FIPS suite enables Ingress on all components, that external URL is an unreachable public route from inside the pod, so the init container hangs forever which cascades to console-ui and blocking Console from ever going Ready. The sibling Rekor code path ui/deployment.go still has this guard, which confirms the drop was accidental.

**Current Fix**

Restored the guard in internal/controller/console/actions/api/deployment.go: to use the in-cluster TUF URL unless the Console CR explicitly configures one, matching the existing [URL pattern](https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/tuf/actions/status-url.go#L48). Added a regression test in console_controller_test.go (fails on old code, passes with fix); full controller test suite passes.

**Additional Fix**

The verify-enterprise-contract check was failing "policy", a new Konflux policy now requires operator bundles to grant RBAC (create, delete, update/patch) on networking.k8s.io/networkpolicies for their operands. Currently the operator's ClusterRole only had this for ingresses. This fix also includes +kubebuilder:rbac marker for networkpolicies in internal/controller/types.go and regenerated config/rbac/role.yaml via make manifests.

### DIFF
--- a/internal/controller/console/actions/api/deployment.go
+++ b/internal/controller/console/actions/api/deployment.go
@@ -48,9 +48,14 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1.Console) *ac
 		result controllerutil.OperationResult
 	)
 
-	tufURL, err := utils.ResolveExternalServiceUrl(ctx, i.Client, instance.Spec.Api.Tuf, instance.Namespace, &rhtasv1.Tuf{})
-	if err != nil {
-		return i.Error(ctx, fmt.Errorf("error resolving TUF URL: %w", err), instance)
+	var tufURL string
+	if instance.Spec.Api.Tuf.URL != "" || instance.Spec.Api.Tuf.Ref != nil {
+		tufURL, err = utils.ResolveExternalServiceUrl(ctx, i.Client, instance.Spec.Api.Tuf, instance.Namespace, &rhtasv1.Tuf{})
+		if err != nil {
+			return i.Error(ctx, fmt.Errorf("error resolving TUF URL: %w", err), instance)
+		}
+	} else {
+		tufURL = fmt.Sprintf("http://tuf.%s.svc", instance.Namespace)
 	}
 
 	l := labels.For(actions.ApiComponentName, actions.ApiDeploymentName, instance.Name)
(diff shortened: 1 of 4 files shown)


internal/controller/console/actions/api/deployment.go
```diff
@@ -48,9 +48,14 @@ func (i deployAction) Handle(ctx context.Context, instance *rhtasv1.Console) *ac
...
-	if err != nil {
-		return i.Error(ctx, fmt.Errorf("error resolving TUF URL: %w", err), instance)
+	var tufURL string
+	if instance.Spec.Api.Tuf.URL != "" || instance.Spec.Api.Tuf.Ref != nil {
+		tufURL, err = utils.ResolveExternalServiceUrl(ctx, i.Client, instance.Spec.Api.Tuf, instance.Namespace, &rhtasv1.Tuf{})
+		if err != nil {
+			return i.Error(ctx, fmt.Errorf("error resolving TUF URL: %w", err), instance)
+		}
+	} else {
+		tufURL = fmt.Sprintf("http://tuf.%s.svc", instance.Namespace)
```
JasonPowr: Just a small nit, besides this looks good to me :) 

use `tufConstants.DeploymentName` instead of the hardcoded "tuf" string, consistent with https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/tuf/actions/status-url.go#L48

JasonPowr (approved):

PR #2215, opened 2026-08-13
feat: implement user specification tracking for pod resources and reconcile logic

jourdee-lab:
### Summary
Adds stale resource cleanup for user-specified volumes and volume mounts during reconciliation, this is PR is working in conjuction with #2229 

### Problem
The LastUserSpecApplied annotation added in this PR stores one just one flat JSON payload for PodExtensions tracking not parts like TLS cert volumes, TrustedCA volumes, env vars. This would use an incompatible flat schema under the same key which wouldn't work together, as whichever reconciles second on a given
object silently clobbers the other's tracked state, since both implementations self-heal on unmarshal failure by resetting to empty.

  ### Review feedback/ What Changed
- Add ensure.ReadNamespacedState / ensure.WriteNamespacedState: generic,
  exported helpers that read/write one concern's key, doing a proper
  read-merge-write so other concerns' data is never touched. Future
  tracking needs (e.g. Auth) can reuse these instead of re-solving the
  same problem.
- Scope UserSpecState/ReadLastApplied/WriteLastApplied to the
  "podExtensions" key via the new helpers.
- Remove UserSpecState.Annotations/.Labels and their Equal method, plus
  ManagedKeys/KeySet with dead scaffolding with no production caller.
- Update the LastUserSpecApplied doc comment to describe the namespaced
  schema.
- Add TestPodExtensions_PreservesOtherAnnotationConcerns, an end-to-end
  test proving a PodExtensions reconcile leaves another concern's
  annotation data intact.

### DIFF
--- a/internal/annotations/annotations.go
+++ b/internal/annotations/annotations.go
@@ -147,6 +147,22 @@ const (
 	// trust material change and accept the newly observed value.
 	RefreshTrustMaterial = "rhtas.redhat.com/refresh-trust-material"
 
+	// LastUserSpecApplied tracks the names of user-specified resources (volumes,
+	// volume mounts, annotations, labels) applied during the last reconcile.
+	// The value is a JSON-encoded object used to detect removals on the next reconcile.
+	//
+	// We use an annotation-based diff approach instead of Server-Side Apply (SSA)
+	// because SSA field ownership tracks at the field level, not the semantic
+	// "user-specified resource" level. When the operator and user both touch
+	// PodSpec.Volumes, SSA cannot distinguish operator-managed volumes from
+	// user-specified ones — conflicts arise on shared slice fields, and removing
+	// a single user volume requires the operator to re-apply the entire field,
+	// which risks clobbering operator-managed entries. This annotation sidesteps
+	// SSA ownership entirely: it records which named resources the user specified,
+	// so the next reconcile can compute the diff and remove stale items without
+	// affecting operator-managed resources in the same slice.
+	LastUserSpecApplied = "rhtas.redhat.com/last-user-spec-applied"
+
 	TLS = "service.beta.openshift.io/serving-cert-secret-name"
 )
 
--- a/internal/annotations/annotations.go
+++ b/internal/annotations/annotations.go
@@ -147,9 +147,17 @@ const (
 	// trust material change and accept the newly observed value.
 	RefreshTrustMaterial = "rhtas.redhat.com/refresh-trust-material"
 
-	// LastUserSpecApplied tracks the names of user-specified resources (volumes,
-	// volume mounts, annotations, labels) applied during the last reconcile.
-	// The value is a JSON-encoded object used to detect removals on the next reconcile.
+	// LastUserSpecApplied tracks, per concern, the names of user-specified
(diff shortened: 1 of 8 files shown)


internal/utils/kubernetes/ensure/user_spec_tracking.go
```diff
@@ -0,0 +1,95 @@
...
+	"slices"
+	"sort"
+
+	"github.com/securesign/operator/internal/annotations"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// UserSpecState tracks the names of user-specified resources that were applied
+// during the last reconcile. Stored as JSON in the LastUserSpecApplied annotation.
+type UserSpecState struct {
```
bouskaJ, line 20: Do not create function operates on fixed list. Let's try to reuse the Toggleable struct like this
``` 
// EnsureWithCleanup applies Ensure and removes stale items from Managed in one pass
  func EnsureWithCleanup[T any](target T, t Toggleable[T]) error {
      if err := t.Ensure(target); err != nil {
          return err
      }
      return removeManaged(target, t.Managed)
  }
```
  Then the call site would look like what you're describing:
```
  ensure.EnsureWithCleanup(podSpec, ensure.Toggleable[*core.PodSpec]{
      Ensure: applyCurrentExtensions,
      Managed: diffFromAnnotation(prev, current),
  })
```

internal/utils/kubernetes/ensure/deployment/deployment.go
```diff
@@ -56,13 +56,17 @@ func PodRequirements(requirements rhtasv1.PodRequirements, containerName string)
 }
 
 // PodExtensions applies user-defined init containers, volumes, and volume mounts
-// to a Deployment. It is a standalone ensure function so callers can compose it
-// independently from signer-specific logic.
+// to a Deployment. It reads the previous state from the tracking annotation,
+// removes stale resources, upserts desired ones, and writes the current state back.
 func PodExtensions(ext rhtasv1.PodExtensions, containerName string) func(*v1.Deployment) error {
```
bouskaJ, line 60: do not implement per-type function. We need something generic like https://github.com/securesign/secure-sign-operator/blob/main/internal/utils/kubernetes/ensure/toggleable.go. It is not reusable directly (Ensure + cleanup needs to be done in one call) so you would need to create a sibling function.

internal/annotations/annotations.go
```diff
@@ -147,6 +147,11 @@ const (
 	// trust material change and accept the newly observed value.
 	RefreshTrustMaterial = "rhtas.redhat.com/refresh-trust-material"
 
+	// LastUserSpecApplied tracks the names of user-specified resources (volumes,
```
bouskaJ: Add a better rationalize why we did not go with SSA

bouskaJ (changes requested):

sampras343 (commented):
- Simple and focused
- PodExtensions exclusively owns the annotation 
- Extension to other parts like TLS cert volumes, TrustedCA volumes, env vars needs additional code
- Name only tracking will not consider field-level mutations within a named resource

bouskaJ:
we will move to SSA in 1.6 release - closing

PR #2228, opened 2026-08-18
fix: swapping from quay to in memory test registry

tommyd450:


test/e2e/support/registry.go
```diff
@@ -0,0 +1,224 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+const (
+	registryImage = "docker.io/library/registry:2"
```
osmman: Have we got configured mirror for not pulling directly from docker.io? If not than our job will be unstable due rate limiting from docker registry.

osmman: We are using `mirror.gcr.io` in other automations

tommyd450: Nope, will look into it, was hoping to move this along on friday to remove secret config for running e2e tests on cluster.

osmman: you can cahnge it to `mirror.gcr.io/library/registry:2`

tommyd450: fixed.

PR #2229, opened 2026-08-18
feat: user-spec-cleanup

bouskaJ:


sampras343 (commented):
- Bugs are valid and needs to be fixed
- This solves the problem on a single object level
- Generic enough that no new code needs to be added for a new domain
  - Works for any client Object
- Full managed state (env var values, volume sources, mount paths) in the annotation, not just names.
- Arch advantage

bouskaJ:
there is a problem with merging resulting annotation on multiple calls per resource. Do you think we can compact all user-spec fields into one call? It would reduce the complexity of the code.

sampras343:
> there is a problem with merging resulting annotation on multiple calls per resource. Do you think we can compact all user-spec fields into one call? It would reduce the complexity of the code.

@bouskaJ If I'm understanding correctly, 

If Auth runs first and writes `{"env": ["AUTH_KEY"]}`, then PodExtensions runs and needs to merge `{"volumes": ["custom-vol"]}` into the annotation without taking into account of what Auth wrote, that merge is where all the complexity lives.

And your thought is that the operator already knows at the call site that both Auth and PodExtensions need cleanup on this Deployment. So, instead of letting each one independently read-modify-write the annotation, collect them upfront and do it once.

Therefore, no merge needed because you never have partial writes.

internal/utils/kubernetes/ensure/toggleable.go
```diff
@@ -29,6 +36,519 @@ func OptionalToggle[T any](condition bool, t Toggleable[T]) func(T) error {
...
+func UserSpecifiedToggle[T client.Object](t Toggleable[T]) func(T) error {
+	return func(live T) error {
+		stored := live.GetAnnotations()[annotations.LastUserSpecApplied]
+
+		// Extract lastApplied for only the fields this call manages
+		var lastAppliedTyped T
+		if stored != "" {
+			// Unmarshal full stored annotation
+			fullLastApplied := reflect.New(reflect.TypeOf(live).Elem()).Interface().(T)
+			if err := json.Unmarshal([]byte(stored), fullLastApplied); err != nil {
```
osmman, line 78: use json encoder from kubernetes utils it handle edge cases better for empty/nil values

bouskaJ:
> > there is a problem with merging resulting annotation on multiple calls per resource. Do you think we can compact all user-spec fields into one call? It would reduce the complexity of the code.
> 
> @bouskaJ If I'm understanding correctly,
> 
> If Auth runs first and writes `{"env": ["AUTH_KEY"]}`, then PodExtensions runs and needs to merge `{"volumes": ["custom-vol"]}` into the annotation without taking into account of what Auth wrote, that merge is where all the complexity lives.
> 
> And your thought is that the operator already knows at the call site that both Auth and PodExtensions need cleanup on this Deployment. So, instead of letting each one independently read-modify-write the annotation, collect them upfront and do it once.
> 
> Therefore, no merge needed because you never have partial writes.

yes, exactly. The use-case will be:
```
       managed := &v1.Deployment{}
   //      traverse over Auth and set managed Fields on managed
   //      traverse over InitContainers and set managed Fields on managed

	toggle := ensure.Toggleable[*v1.Deployment]{{
		Ensure: func(d *v1.Deployment) error {
			
			if err :=  ensure.Auth(containerName, auth)(&d.Spec.Template.Spec) {
                           return err
                        }

                       
			if err :=  ensure.InitContainers(...)(&d.Spec.Template.Spec) {
                           return err
                        }
		},
		Managed: managed
	}

	return ensure.UserSpecifiedToggle(toggle)
```

All user-spec fields processed in one `UserSpecifiedToggle` call

internal/utils/kubernetes/ensure/toggleable.go
osmman: k8s.io/apimachinery/pkg/util/json

bouskaJ:
Agreed to move to full SSA on 1.6 release - closing this one

PR #2230, opened 2026-08-18
feat: add ingress support to CTlog

osmman:
## Why

CTlog is the only signing-path component without optional Ingress support, unlike Fulcio, Rekor, TUF, and TSA. CTlog is also the first component whose backend pod can terminate real TLS itself, so exposing it externally needs OpenShift Route `reencrypt` termination instead of the `edge` termination the other components use — edge would decrypt to plaintext at the router and break an HTTPS-only backend.

## What

- Add an optional `Ingress` field to `CTlogSpec`, with defaults/deepcopy/conversion/CRDs regenerated.
- Add an ingress reconcile action for CTlog, registered in the reconciler and watched via `Owns(&networkingv1.Ingress{})`.
- Select Route `reencrypt` termination instead of `edge` when CTlog's own TLS is enabled.
- Refactor `EnsureIngressTLS` into a thin wrapper around a new exported `EnsureIngressTermination(termination string)`, reused for the edge/reencrypt selection without changing behavior for existing callers.
- Port the ingress-aware URL branch into CTlog's status-url action so `Status.Url` reflects the external ingress URL when ingress is enabled.
- Restore the `Ingress` field during `v1alpha1<->v1` round-trip conversion for both `CTlog` and `Securesign`.
- Add CTlog to the e2e `WithIngress()` helper.

### DIFF
--- a/internal/controller/ctlog/actions/monitor/statefulset.go
+++ b/internal/controller/ctlog/actions/monitor/statefulset.go
@@ -26,7 +26,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	rhtasv1 "github.com/securesign/operator/api/v1"
-	ctlogutils "github.com/securesign/operator/internal/controller/ctlog/utils"
 	tlsensure "github.com/securesign/operator/internal/utils/tls/ensure"
 )
 
@@ -68,7 +67,13 @@ func (i statefulSetAction) Handle(ctx context.Context, instance *rhtasv1.CTlog) 
 			ObservedGeneration: instance.Generation,
 		})
 	}
-	ctlogServerHost, err := serviceresolver.Resolve(instance)
+
+	// Deliberately not serviceresolver.Resolve/ResolveExternalServiceUrl like Rekor's monitor:
+	// unlike Rekor (which validates a signed checkpoint), this monitor compares the log URL
+	// byte-for-byte against the URL stored in TUF metadata, so it must dial the exact external
+	// URL once ingress is enabled. ResolveExternalServiceUrl requires CTlog.Ready, but this
+	// action runs before Ready (server and monitor share one CR), which would deadlock.
+	ctlogServerHost, err := actions.ResolveUrl(ctx, i.Client, instance)
 	if err != nil {
 		return i.Error(ctx, err, instance)
 	}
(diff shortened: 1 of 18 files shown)


bouskaJ:
why we need this? Is it just for unification with other components or there is a customer request?

osmman:
> why we need this? Is it just for unification with other components or there is a customer request?

Mainly for unification. There's no customer request, but it fixes an issue where our CTlog isn't accessible from external monitors outside the cluster, making it behave the same way as the Rekor tlog.

internal/controller/ctlog/actions/monitor/statefulset.go
```diff
@@ -68,7 +67,7 @@ func (i statefulSetAction) Handle(ctx context.Context, instance *rhtasv1.CTlog)
 			ObservedGeneration: instance.Generation,
 		})
 	}
-	ctlogServerHost, err := serviceresolver.Resolve(instance)
+	ctlogServerHost, err := actions.ResolveUrl(ctx, i.Client, instance)
```
bouskaJ, line 76: We should stay consistent with Rekor: https://github.com/securesign/secure-sign-operator/blob/main/internal/controller/rekor/actions/monitor/statefulset.go#L68. If you want to use Ingress (when enabled), you can use this: https://github.com/securesign/secure-sign-operator/blob/main/internal/serviceresolver/service_ref_resolver.go#L49. Although it pulls the URL from the status, it will fall back to the internal address if Ingress is enabled.

osmman: I tried both `ResolveUrl` and `ResolveExternalServiceUrl`, but neither works for the CTlog monitor as long as the monitor lives in the CTlog CR. The blocker is the tight coupling of the CTlog **server** and **monitor** in one CRD plus the ctlog-monitor's internal implementation.

**Why Rekor's approach doesn't transfer:** the Rekor monitor doesn't verify the tlog by URL — it validates the **signed checkpoint**, so the internal cluster-DNS address is fine and `serviceresolver.Resolve` works. The CTlog monitor is different: it compares the log URL **byte-for-byte** against what's stored in the TUF metadata. When ingress is enabled, TUF holds the **external** URL, so the monitor must dial that exact external URL, an internal address mismatches and the check fails. That rules out the Rekor-style internal resolver for CTlog.

- **`ResolveExternalServiceUrl`** returns the correct external URL, but gates on `CTlog.Ready == True`. The monitor StatefulSet action runs *before* the Ready transition, and server+monitor share the one CR, so Ready can't flip until the monitor succeeds → deadlock. The StatefulSet is never created (that's the CI timeout).
- **`ResolveUrl`**  no readiness gate, so no deadlock, but it still can't satisfy the monitor's internal byte-for-byte URL comparition with Trusted root data.

bouskaJ: can you add a comment that you re-implement the URL lookup on purpose?

osmman: @bouskaJ comment added

bouskaJ (approved):

PR #2231, opened 2026-08-18
SECURESIGN-5150: Migrate TUF spec from secret-based keys to service b…

knrc:
…inding references


knrc:
@bouskaJ Please be aware, I had to relax the crdify rules to get these changes to pass.  We will need to revert these prior to release.

crdify.yaml
```diff
@@ -1,11 +1,19 @@
 # Matches OLM v1 default configuration for CRD upgrade safety preflight checks.
 # Reference: https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/rukpak/preflights/crdupgradesafety/crdupgradesafety.go
+# v1 has not yet shipped, so we can tolerate breaking changes at this time.
+# The following configuration must be reverted just prior to the v1 release.
```
bouskaJ: Do not modify the configuration. We are deliberately ignoring the CRDIFy linter failures; it is safer to keep them failing than to suppress them and forget to re-enable them.

api/v1/common.go
```diff
@@ -113,6 +113,29 @@ type LocalObjectReference struct {
...
+type TrustRootBindingWithOIDC struct {
+	TrustRootBinding `json:",inline"`
+	// OIDCIssuers is a list of OIDC issuer URLs to include in the Fulcio signing configuration.
+	// Use for manual configuration; when specified, these values take precedence over auto-loaded OIDC configuration from the Fulcio service reference.
+	//+optional
+	//+listType=set
+	//+kubebuilder:validation:items:Pattern=`^[a-zA-Z][a-zA-Z0-9+.-]*://.+$`
+	OIDCIssuers []string `json:"oidcIssuers,omitempty"`
+}
+
```
bouskaJ: All of these are trustRoot-related—not shared with other components—so we can keep them in `tuf_types.go`.

api/v1/tuf_types.go
```diff
@@ -48,62 +37,57 @@ type TufSpec struct {
...
 	Pvc Pvc `json:"pvc,omitempty"`
-	// Ctlog service configuration
+	// Ctlog service and trust material binding.
 	//+optional
-	//+kubebuilder:validation:XValidation:rule="!has(self.url) || size(self.url) == 0 || self.url.matches('^([a-zA-Z][a-zA-Z0-9+.-]*://[^/]+/.+|//[^/]*/.+)$')",message="url must follow the pattern scheme://host[:port]/path or //[:port]/path"
-	Ctlog ServiceReference `json:"ctlog,omitempty"`
-	// Fulcio service configuration
+	// At most one entry is allowed today; this ceiling is deliberate and temporary,
+	// pending future multi-instance support.
+	//+kubebuilder:validation:MaxItems:=1
```
bouskaJ, line 49: Just a note: Future removal of this rule is safe (https://github.com/kubernetes-sigs/crdify/blob/main/docs/validations.md#maximum-maxlength-maxitems-and-maxproperties) - tested.

api/v1/common.go
```diff
@@ -113,6 +113,29 @@ type LocalObjectReference struct {
 	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
 }
 
+// TrustRootBinding identifies a component's service binding and, optionally, a
+// reference to the Secret holding its trust material (public key for
+// Rekor/CTlog, cert chain for Fulcio/TSA). If SecretRef is unset, the operator
+// resolves material from the referenced or autodiscovered component CR's
+// status.
+type TrustRootBinding struct {
```
bouskaJ: The API doesn't prevent Ref + SecretRef being set together, which is semantically odd (why ref the component if you're not using its status?). The XValidation only prevents Ref + URL

bouskaJ (changes requested):
Just a few minor issues. Otherwise LGTM. Good work.

crdify.yaml
knrc: Sounds good, I can revert that.

knrc: I've reverted this in the latest push, please take a look

api/v1/common.go
knrc: These types have been moved in the latest push, please take a look

api/v1/common.go
knrc: I decided not to add the extra validation rule for now, as this would be an additional restriction on potentially existing v1alpha1 configurations.  From what I can see there was nothing tying the Keys to the ServiceReference fields, they were previously independent and this PR is only unifying the config into a shared hierarchy.

We should talk more about this

bouskaJ: We need to distinguish between two cases here:

- **No Ref (Autoloading)**: In this scenario, the secret binding remains valid to preserve legacy behavior from the alpha API.

- **Ref is set to an object**: Adding a validation rule to ensure secretRef is not set at the same time won't break any existing functionality. Because Ref could not be set in the alpha API, this case only applies to newly created instances.

We should implement the strictest possible validation rules from the start. While we can always relax these rules later, failing to enforce them in the API now means the field will have to remain permissive indefinitely (due to OLM 1 crdify validation constraints).

knrc: @bouskaJ Updated to match the above, please take a look

bouskaJ (approved):

PR #2235, opened 2026-08-20
feat: re-reconcile managed controllers on cluster TLS profile change

anithapriyanatarajan:
## Summary

When a cluster admin changes `spec.tlsSecurityProfile` on the OpenShift
`config.openshift.io` `APIServer/cluster` singleton, every managed controller now
**re-reconciles** so all TLS-sensitive outputs — Route annotations, the `redis.conf`
ConfigMap, and operand `GODEBUG` env vars — converge on the new cluster TLS security
profile within one reconcile cycle.

This PR adds **only the trigger** (the Watch). It does not change any reconcile
business logic. The reconcile bodies already resolve the cluster profile
(SECURESIGN-4993) and are idempotent via `kubernetes.CreateOrUpdate` + the `ensure.*`
helpers, so re-enqueuing is sufficient for convergence.

Jira: **SECURESIGN-4996**
Builds on **SECURESIGN-4993** (cluster TLS profile resolver) and **#2040** (RBAC + cache scaffolding).

## What changed

**1. `Watch` on `configv1.APIServer` in all 7 controllers**
`ctlog`, `fulcio`, `rekor`, `trillian`, `tsa`, `tuf`, `securesign` each register a watch
on the cluster APIServer singleton via a new shared helper. The handler maps the
singleton change → enqueues **every instance** of that controller's own CR kind; the
APIServer object itself is never reconciled. All existing `For` / `Owns` / `Watches`
entries are left intact.

**2. Shared helper — `internal/utils/controller/apiserver_watch.go`**
- `WatchAPIServer(...)` — registers the watch, guarded by `kubernetes.IsOpenShift()`,
  with `predicate.ResourceVersionChangedPredicate{}`.
- `EnqueueAllOnClusterConfigChange(...)` — lists the controller's CR kind and enqueues
  one request per instance.

## OpenShift-only guard

`configv1.APIServer` does not exist on vanilla Kubernetes, and watching a missing CRD
makes the manager fail at startup. Registration is therefore wrapped in
`kubernetes.IsOpenShift()` (the same pattern already used across the codebase). On
vanilla k8s the watch is simply never added and the operator starts normally.

## Storm prevention

- `ResourceVersionChangedPredicate` drops no-op cache resyncs, so periodic informer
  resyncs of the singleton never trigger a reconcile.
- The watch only fires on real APIServer changes — unrelated changes to a Securesign (or
  any operand) CR do **not** cause an APIServer re-read.
- Reconcile bodies remain idempotent (`CreateOrUpdate` + `ensure.*`), so a redundant
  enqueue is a cheap no-op when nothing changed.

## Cache singleton restriction (already present, verified)

The APIServer informer cache is pinned to `metadata.name=cluster` in
`cmd/main.go` (inside the existing `IsOpenShift()` block, alongside `configv1.Ingress`).
The operator's ClusterRole is scoped to `resourceName=cluster`, so a cluster-wide list
would be forbidden — the singleton restriction keeps the cache legal.

## RBAC (already present, verified)

The `config.openshift.io` `apiservers` `get;list;watch` marker
(`internal/controller/types.go`) and the generated cluster-scoped ClusterRole rule
(`config/rbac/role.yaml`) already exist from #2040. `make manifests` produces **no diff**
to `role.yaml` in this PR.

## Tests

- New unit tests in `internal/utils/controller/apiserver_watch_test.go`:
  - `EnqueueAllOnClusterConfigChange` fans out to every CR instance across namespaces and
    ignores the triggering object.
  - `WatchAPIServer` registers the watch **only** when `IsOpenShift()` is true
    (0 registrations on vanilla k8s, exactly 1 on OpenShift, on `*configv1.APIServer`).
- All 7 controller envtest suites pass.

```
make manifests generate fmt vet   # clean; git diff config/rbac/role.yaml is empty
go build ./...                    # ok
go test ./internal/controller/... ./cmd/...   # pass
```

## Non-goals

- No changes to reconcile business logic, Route annotation writing, `redis.conf`
  generation, or `GODEBUG` injection (separate stories).
- No process-restart watcher — this is re-reconcile, not restart.
- All new behavior is OpenShift-guarded; vanilla Kubernetes startup is unaffected.

### DIFF
--- a/internal/controller/ctlog/ctlog_controller.go
+++ b/internal/controller/ctlog/ctlog_controller.go
@@ -163,7 +163,7 @@ func (r *ctlogReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return err
 	}
 
-	return ctrl.NewControllerManagedBy(mgr).
+	b := ctrl.NewControllerManagedBy(mgr).
 		WithEventFilter(pause).
 		For(&rhtasv1.CTlog{}, builder.WithPredicates(tasPredicate.ConfigurationChangedOnFailurePredicate[*rhtasv1.CTlog]())).
 		Owns(&v1.Deployment{}).
@@ -202,6 +202,11 @@ func (r *ctlogReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			ctrlutil.ServiceRefWatch(mgr.GetClient(), &rhtasv1.CTlogList{}, func(o client.Object) rhtasv1.ServiceReference {
 				return o.(*rhtasv1.CTlog).Spec.Monitoring.Tuf
 			}),
-		), builder.WithPredicates(crpredicate.GenerationChangedPredicate{})).
-		Complete(r)
+		), builder.WithPredicates(crpredicate.GenerationChangedPredicate{}))
+
+	// Re-reconcile all CTlog instances when the cluster-wide TLS security profile
+	// (config.openshift.io APIServer/cluster) changes. OpenShift-only; no-op on vanilla k8s.
+	ctrlutil.WatchAPIServer(b, mgr.GetClient(), &rhtasv1.CTlogList{})
+
+	return b.Complete(r)
 }
--- a/internal/controller/fulcio/fulcio_controller.go
+++ b/internal/controller/fulcio/fulcio_controller.go
@@ -147,7 +147,7 @@ func (r *fulcioReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return err
 	}
 
-	return ctrl.NewControllerManagedBy(mgr).
+	b := ctrl.NewControllerManagedBy(mgr).
 		WithEventFilter(pause).
 		For(&rhtasv1.Fulcio{}, builder.WithPredicates(predicate.ConfigurationChangedOnFailurePredicate[*rhtasv1.Fulcio]())).
 		Owns(&v1.Deployment{}).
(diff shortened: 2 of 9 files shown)


anithapriyanatarajan:
@osmman @bouskaJ - This PR is to address https://redhat.atlassian.net/browse/SECURESIGN-4996. Scope is restricted to reconcile every component controller when the security profile is updated. There will be one more follow up that wires the values derived in PR#2040 to the respective components. This PR simply focuses in watch API Server and reconcile for security profile change. Thank you

osmman:
@anithapriyanatarajan 

I don't think we need this because the restart path of operator already covers it.

`main.go:334-353` already restarts the operator (`cancel()`) the moment the TLS profile or adherence policy changes. On restart, every controller does a fresh informer List, which reconciles all existing CRs anyway. That's just controller-runtime's normal startup behavior.

So this watch reacts to the same event that's already killing the process. Either it loses the race with `cancel()` and never fires, or it wins and enqueues reconciles that get thrown away when the pod dies moments later. Either way, it can't add anything the restart doesn't already give us for free.

Given that, I'd lean toward closing this rather than merging ~200 lines to duplicate existing behavior — unless I'm missing a case the restart doesn't cover.

anithapriyanatarajan:
@osmman  Thank you very much. Apologise to have over worked this. :disappointed: Will close and work the wiring part.

PR #2238, opened 2026-08-24
fix: disable cluster-wide caching of Secret and ConfigMap data (backport to release-1.4)

kdacosta0:
## What
Backport of the Secret/ConfigMap cache-disable fix from `main` (`5d047ed7`, `ffbdc504`) to `release-1.4`, plus a follow-up fix for additional call sites found during cluster testing.

- `cmd/main.go`: `Client.Cache.DisableFor` for `corev1.Secret`/`corev1.ConfigMap` — reads go straight to the API server instead of the shared informer cache.
- New `kubernetes.Create` helper (Create without the existence-check Get) used for the 6 `GenerateName` Secret/ConfigMap objects covered by upstream (trillian DB secret, ctlog/fulcio config, rekor sharding config, redis password, tsa NTP config).
- **Follow-up**: `release-1.4` still has 5 pre-generic-signer-action call sites (Fulcio cert, Rekor signer, Rekor pub key, CTlog keys, TSA signer) with the same `GenerateName` + `CreateOrUpdate` pattern. These weren't in the upstream commit since `main` had already replaced them via a later generic-signer refactor. Deploying this backport on a live cluster surfaced it immediately: Fulcio/Rekor/TSA/CTlog all failed reconciliation with `resource name may not be empty`. Fixed the same way (`Create` instead of `CreateOrUpdate`), with a regression test per call site.

## Why
Secret/ConfigMap have no `ByObject` label filter (unlike Deployment/Pod/Service), so any Get/List against them cached **every** Secret/ConfigMap in the cluster, including decrypted key material — flagged by a customer as a data-exposure issue. Label-filtering isn't an option here since CRDs allow unlabeled, user-provided secret/configmap refs (BYO keys/certs).

Disabling the cache means `CreateOrUpdate`'s probing `Get` on any `GenerateName` object (empty name) now hits the API server and gets `BadRequest` instead of a cache `NotFound`, breaking creation. `kubernetes.Create` skips that Get entirely for these always-new objects.

## Not included
1.5.0 also has `649b3bd2`/`1f7a7d0e` (metadata-only Secret reads) and the later generic-signer-action refactor — skipped, they depend on helpers/refactors not present on `release-1.4` and aren't needed for the customer fix.

## Test plan
- [x] `make test` — full suite green (envtest included), all 11 affected call sites covered
- [x] `go vet`, `gofmt`, `golangci-lint` clean
- [x] Deployed to a live OpenShift cluster: full `Securesign` CR (Trillian, CTlog, Fulcio, Rekor, TSA) reconciles cleanly, no `resource name may not be empty` errors
- [x] Live cluster memory test: 150 Secrets + 150 ConfigMaps, ~900KB payload each (~270MB total), fresh operator pod per run:
  - `v1.4.3` (unfixed): **OOMKilled**
  - This backport: **33.3 MiB**, no crash

REF: [SECURESIGN-5387](https://redhat.atlassian.net/browse/SECURESIGN-5387)

[SECURESIGN-5387]: https://redhat.atlassian.net/browse/SECURESIGN-5387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -235,6 +235,18 @@ func main() {
 			RecoverPanic: ptr.To(true),
 		},
 		Cache: cacheOpts,
+		// Secret/ConfigMap are excluded from cacheOpts.ByObject: a label/field selector on
+		// these types would return a permanent NotFound for any object outside the selector
+		// (no live fallback), breaking user-provided (BYO) secret/configmap refs. DisableFor
+		// keeps reads live instead, avoiding an unfiltered, cluster-wide cache.
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
+					&corev1.Secret{},
+					&corev1.ConfigMap{},
+				},
+			},
+		},
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
--- a/internal/controller/ctlog/actions/handle_keys.go
+++ b/internal/controller/ctlog/actions/handle_keys.go
@@ -236,7 +236,7 @@ func (g handleKeys) generateAndUploadSecret(ctx context.Context, instance *v1alp
 			Namespace:    instance.Namespace,
 		},
 	}
-	if _, err = kubernetes.CreateOrUpdate(ctx, g.Client,
+	if err = kubernetes.Create(ctx, g.Client,
 		secret,
 		ensure.ControllerReference[*v1.Secret](instance, g.Client),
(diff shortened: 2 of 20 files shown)


kdacosta0:
It's connected to Customer Case: 
`RHTAS 1.43 Upgrade Causes rhtas-operator-controller-manager CrashLoopBackOff with OOMKilled (Exit Code 137)`

osmman (approved):

PR #2245, opened 2026-08-25
feat: gate component rollout on cluster TLS profile adherence

anithapriyanatarajan:
## Summary

Adds a per-component gate that decides what happens when a managed component cannot honour the cluster TLS security profile. The effective mode is derived from the cluster-wide TLS adherence policy combined with an optional `rhtas.redhat.com/tlsAdherence` annotation.

The cluster policy is treated as a **floor** — the annotation may match or tighten it, never relax it:

- **strict** → block the rollout with a terminal error condition
- **legacy** → deploy with a `Warning` condition
- **legacy annotation on a strict cluster** → configuration error

On vanilla Kubernetes, or when profile resolution is disabled, the gate is a no-op. The per-component capability check is currently stubbed (`TODO STORY-994`) until upstream binaries expose the signal.

## Changes

- New inheritable `rhtas.redhat.com/tlsAdherence` annotation (propagates Securesign → all components).
- Snapshot the OpenShift `APIServer` TLS adherence policy at startup; gate runs only when the cluster profile is actually resolved.
- New `internal/action/tlsadherence` gate, wired into all seven component controllers.
- Tests for intent/mode resolution, policy mapping, annotation propagation, and every legacy/strict/conflict/disabled outcome.

No CRD schema change. This is an interim control, to be deprecated once all upstream binaries honour the cluster profile.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -54,6 +54,7 @@ import (
 	consolev1 "github.com/openshift/api/console/v1"
 	v1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
+
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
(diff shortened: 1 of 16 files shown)


anithapriyanatarajan:
Superseded by #2246 (branch now lives in the base repo).

PR #2246, opened 2026-08-25
feat: gate component rollout on cluster TLS profile adherence (SECURESIGN-4995)

anithapriyanatarajan:
## Summary

Adds a per-component gate that decides what happens when a managed component cannot honour the cluster TLS security profile. The effective mode is derived from the cluster-wide TLS adherence policy combined with an optional `rhtas.redhat.com/tlsAdherence` annotation.

The cluster policy is treated as a **floor** — the annotation may match or tighten it, never relax it:

- **strict** → block the rollout with a terminal error condition
- **legacy** → deploy with a `Warning` condition
- **legacy annotation on a strict cluster** → configuration error

On vanilla Kubernetes, or when profile resolution is disabled, the gate is a no-op and any stale adherence condition is removed. The per-component capability check is currently stubbed until upstream binaries expose the signal.

## Changes

- New inheritable `rhtas.redhat.com/tlsAdherence` annotation (propagates Securesign → all components).
- Snapshot the OpenShift `APIServer` TLS adherence policy at startup; gate runs only when the cluster profile is actually resolved.
- Wrap both cluster TLS fetches in a shared transient-error retry so a momentary API blip at startup is not mistaken for a persistent failure.
- New `internal/action/tlsadherence` gate, wired into all seven component controllers.
- Extend `ConditionsAwareObject` with `RemoveCondition` so the gate can clear its condition once a component is no longer gated.
- Tests for intent/mode resolution, policy mapping, annotation propagation, transient-retry behaviour, condition cleanup, and every legacy/strict/conflict/disabled outcome.

No CRD schema change. This is an interim control, to be deprecated once all upstream binaries honour the cluster profile.


internal/action/tlsadherence/action.go
```diff
@@ -0,0 +1,98 @@
...
+	// per-resource annotation to decide the effective enforcement mode.
+	effective, conflict := resolveMode(config.ClusterTLSAdherenceStrict, intentFromAnnotations(instance.GetAnnotations()))
+
+	msg := fmt.Sprintf("component %s cannot honour the cluster TLS security profile", i.component)
+
+	// A legacy annotation on a strict cluster tries to relax below the cluster
+	// mandate: surface it as a configuration error rather than silently upgrading.
+	if conflict {
+		err := reconcile.TerminalError(fmt.Errorf("%s: tlsAdherence=legacy conflicts with the cluster TLS adherence policy (StrictAllComponents); the annotation may not relax below the cluster mandate", msg))
+		return i.Error(ctx, err, instance, metav1.Condition{
```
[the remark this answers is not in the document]
anithapriyanatarajan: Good catch — fixed. The message now says the cluster policy is "strict" instead of naming the enum, since any unrecognised value also maps to strict (fail-secure).

internal/annotations/annotations.go
```diff
@@ -125,6 +125,52 @@
...
+// This is an interim control. Once all upstream binaries honour the cluster
+// profile it will be deprecated and removed (see STORY-994).
+//
+// Supported values:
+//   - "legacy": deploy a non-compliant component and surface a Warning condition;
+//     do not block the rollout. This is the operator default when the cluster
+//     policy is not strict and the annotation is absent or empty.
+//   - "strict": block deployment of any component that cannot honour the cluster
+//     TLS profile, surfacing an error condition instead of rolling out.
+//
```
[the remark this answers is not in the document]
anithapriyanatarajan: Fixed — added a line documenting that any unrecognised value is treated as absent (inherits the cluster policy; strict floor still applies on a strict cluster).

internal/action/tlsadherence/action.go
```diff
@@ -0,0 +1,98 @@
...
+// CanHandle fires only when cluster TLS profile resolution is enforced and the
+// component cannot honour the profile. Compliant components, and any component on
+// a cluster where resolution is disabled (vanilla Kubernetes or
+// DisableClusterTLSProfile), are a no-op.
+func (i adherenceAction[T]) CanHandle(_ context.Context, instance T) bool {
+	if !config.ClusterTLSProfileEnforced() {
+		return false
+	}
+	return !i.canHonourProfile(instance)
+}
```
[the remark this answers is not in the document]
anithapriyanatarajan: Correct in principle, but moot today: CanHonourClusterTLSProfile is a stub that always returns true, so nothing sets this condition in production yet. Clearing/refreshing the condition on recovery is part of the real capability signal tracked under STORY-994.

internal/annotations/annotations.go
```diff
@@ -125,6 +125,56 @@
...
+// setting this annotation to "legacy" is a configuration error and blocks the
+// rollout; an absent annotation inherits the cluster mandate (strict). When the
+// cluster policy is not strict (the common case, including clusters without the
+// TLSAdherence feature gate), this annotation is the operative control.
+//
+// Precedence: if set on the Securesign resource, this annotation is automatically
+// propagated to child resources and overwrites any value set directly on them.
+// Per-component overrides via this annotation are only effective on standalone
+// child CRs that are not managed by a Securesign parent.
+// ([github.com/securesign/operator/api/rhtasv1.Securesign])
```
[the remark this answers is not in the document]
anithapriyanatarajan: Fixed — the real import path is github.com/securesign/operator/api/v1 (rhtasv1 is only a local alias, not part of the path). Corrected this and the 6 other pre-existing api/rhtasv1 links in the same file (Securesign/Rekor/Fulcio/TimestampAuthority) so all godoc references resolve.

internal/action/tlsadherence/doc.go
```diff
@@ -0,0 +1,57 @@
...
+// [github.com/securesign/operator/internal/config.ClusterTLSProfileEnforced]). On
+// vanilla Kubernetes, or when DisableClusterTLSProfile is set, the action is a
+// no-op: no condition is set and the rollout is never blocked.
+//
+// # Capability check
+//
+// Whether a component can honour the profile is decided by the per-component
+// TLS-capability check passed to [NewAction]. Until the upstream binaries
+// expose that signal, [CanHonourClusterTLSProfile] is a stub that always
+// reports compliant (see STORY-994), so the action is a no-op in production
```
osmman: what is STORY-994?

internal/controller/console/console_controller.go
```diff
@@ -82,6 +83,7 @@ func (r *consoleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		consoleapi.NewTlsAction(),
 
 		transitions.NewToCreatePhaseAction[*rhtasv1.Console](),
+		tlsadherence.NewAction[*rhtasv1.Console]("console", tlsadherence.CanHonourClusterTLSProfile[*rhtasv1.Console]),
```
osmman, line 85: We can register `tlsadherence` action under `Pending` phase

internal/action/tlsadherence/doc.go
anithapriyanatarajan: :innocent:  Cleared Now. It meant https://redhat.atlassian.net/browse/SECURESIGN-3609

internal/controller/console/console_controller.go
anithapriyanatarajan: Moved before create

cmd/main.go
```diff
@@ -432,17 +439,40 @@ func resolveClusterTLSProfile(ctx context.Context, cli client.Client, openshift,
 	tlsAdherence, err := ostls.FetchAPIServerTLSAdherencePolicy(ctx, cli)
```
knrc: This is really a comment for changes in the earlier PR, but it would be worth fixing in this one.

The FetchAPIServerTLSAdherencePolicy implementation just issues a Get, which would be susceptible to transient API failures similar to what we saw with the OpenShift detection, it would be worth updating this method to use backoff.  Take a look at `DetectOpenShiftPlatform` and `detectOpenShiftWithRetry`, perhaps these can be abstracted and used for both?

I'm also happy considering this as a necessary follow-on PR, if it keeps this one cleaner.

internal/action/tlsadherence/action.go
```diff
@@ -0,0 +1,98 @@
...
+func (i adherenceAction[T]) Name() string {
+	return fmt.Sprintf("check %s TLS profile adherence", i.component)
+}
+
+// CanHandle fires only when cluster TLS profile resolution is enforced and the
+// component cannot honour the profile. Compliant components, and any component on
+// a cluster where resolution is disabled (vanilla Kubernetes or
+// DisableClusterTLSProfile), are a no-op.
+func (i adherenceAction[T]) CanHandle(_ context.Context, instance T) bool {
+	if !config.ClusterTLSProfileEnforced() {
```
knrc, line 45: I think we want to extend this so we can clean up the condition should the CLI flag be disabled, as it stands the condition would not be removed should someone switch the flag back to false.

knrc (changes requested):
Looks good overall, just a couple of considerations that may need addressing.

cmd/main.go
anithapriyanatarajan: Thank you @knrc . I have incorporated the changes to this PR .

internal/action/tlsadherence/action.go
anithapriyanatarajan: Thank you. Extended `ConditionsAwareObject` with `RemoveCondition` and the gate now removes the TLSProfileAdherence condition when it no longer applies (flag disabled or component compliant). Added a test for the cleanup path

osmman (approved):

knrc (approved):

PR #2276, opened 2026-09-02
feat(tsa): add signer type discriminator enum

kdacosta0:
## What changed
Added a `signer.type` discriminator (`file` / `kms` / `tink`) to `TimestampAuthoritySigner` in the `v1` API, and moved TSA onto the shared `SignerType*` constants (`api/v1/common.go`, now including `SignerTypeTink`).

## Why
Fulcio, Rekor and CTlog already select their signer backend via a `type` enum. TSA was the only component still inferring the backend from struct presence and keeping its own local `FileType`/`KmsType`/`TinkType` constants. This aligns TSA with the rest of the API and centralizes the constants (per earlier review feedback).

## Before
No `type` field — the backend was inferred purely from which sub-struct was set:

```yaml
spec:
  signer:
    kms:
      keyResource: gcp-kms://...
    certificateChain:
      certificateChainRef:
        name: my-chain
```

## Now
`type` is the discriminator; CEL rules validate it against the matching sub-struct. It defaults to `file`.

```yaml
spec:
  signer:
    type: kms
    kms:
      keyResource: gcp-kms://...
    certificateChain:
      certificateChainRef:
        name: my-chain
```


osmman (approved):

api/v1/timestampauthority_types.go
```diff
@@ -50,9 +50,18 @@ type TimestampAuthoritySpec struct {
 }
 
 // TimestampAuthoritySigner defines the desired state of the Timestamp Authority Signer
-// +kubebuilder:validation:XValidation:rule="(has(self.file) || has(self.kms) || has(self.tink)) == has(self.certificateChain.certificateChainRef)",message="external signer (file/kms/tink) and certificateChainRef must be configured together"
 // +kubebuilder:validation:XValidation:rule="[has(self.file), has(self.kms), has(self.tink)].filter(x, x).size() <= 1",message="only one signer should be configured at any time"
+// +kubebuilder:validation:XValidation:rule="(has(self.file) || has(self.kms) || has(self.tink)) == has(self.certificateChain.certificateChainRef)",message="external signer (file/kms/tink) and certificateChainRef must be configured together"
```
bouskaJ, line 54: Securesign umbrella is not defaulted. Isn't the empty type rejected on Securesign?

kdacosta0: Good catch, but an empty type still passes CEL: the guarded rules `(!has(self.type) || ...)` are skipped, and `(has(self.type) && self.type == 'kms') || !has(self.kms)` passes because there's no `kms`/`tink` sub-struct 


So empty type -> file is accepted the same on Securesign as standalone -only kms/tink need an explicit type.

Added envtest for it: 
1. empty embedded TSA type is accepted
2. kms without type is rejected

bouskaJ (approved):

PR #2278, opened 2026-09-02
docs: add PKCS#11 SecureSign sample CR

sampras343:
## Summary
Adds `docs/samples/rhtas_v1_securesign_pkcs11.yaml` — a sample SecureSign CR with PKCS#11/HSM signer configuration for both Fulcio and CTLog using SoftHSM.

This enables the Konflux PKCS#11 E2E pipeline to read the CR from the operator repo (same pattern as file-mode uses `rhtas_v1_securesign.yaml`) instead of hardcoding the CR inline. When the CRD changes, the sample is updated in the same PR — keeping the pipeline in sync automatically.

## Prerequisites (created before applying this CR)
- `softhsm-config` ConfigMap
- `hsm-credentials` Secret (PIN)
- `fulcio-pkcs11-config` Secret (crypto11 JSON)
- `fulcio-root-ca` Secret (root CA certificate from key ceremony)
- `ctlog-public-key` Secret (public key from key ceremony)
- `hsm-tokens-pvc` and `ctlog-hsm-tokens-pvc` PVCs

### DIFF
--- a/docs/samples/rhtas_v1_securesign_pkcs11.yaml
+++ b/docs/samples/rhtas_v1_securesign_pkcs11.yaml
@@ -1,0 +1,108 @@
+apiVersion: rhtas.redhat.com/v1
+kind: Securesign
+metadata:
+  labels:
+    app.kubernetes.io/name: securesign-sample
+    app.kubernetes.io/instance: securesign-sample
+    app.kubernetes.io/part-of: trusted-artifact-signer
+  name: securesign-sample
+spec:
+  fulcio:
+    ingress:
+      enabled: true
+    config:
+      oidcIssuers:
+        - clientID: "trusted-artifact-signer"
+          issuerURL: "https://your-oidc-issuer-url"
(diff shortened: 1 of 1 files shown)


osmman (changes requested):
Please do not add any specific samples into `config/samples` this is mainly only for OLM and all of these are embedded to bundle resources in annotation which is limited on size. 

it is better to put examples as part of doc

sampras343:
> Please do not add any specific samples into `config/samples` this is mainly only for OLM and all of these are embedded to bundle resources in annotation which is limited on size.
> 
> it is better to put examples as part of doc

@osmman Understood. 
This is mainly for e2e for pipelines repo to pick it up. Do you recommend putting it under `docs` or `test/e2e/samples` ?

osmman:
you can in docs/samples/

sampras343:
> you can in docs/samples/

Updated

osmman (approved):

PR #2296, opened 2026-09-08
Kube api linter

bouskaJ:


bouskaJ:
I’m not sure how best to approach the optional/required field migration, which is why I disabled the linter for now.

  The linter enforces Kubernetes API conventions: required fields should use non-pointer types, while  optional scalar fields should use pointers so unset values can be distinguished from zero values.

  I propagated the required-field changes here:
  https://github.com/securesign/secure-sign-operator/pull/2296/changes/976857f0752eda9285e8c1d5acfa868d0
b9a0007

  However, the optional-field migration is not implemented yet, since it would require changing optional scalars throughout the API—for example, string to *string—and updating all affected code and tests.

  Reference: [Kubernetes API conventions: Optional vs. Required](  https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions
  .md#optional-vs-required
)

osmman (approved):

PR #2300, opened 2026-09-09
fix: simplify ctlog convertor

osmman:


api/v1/ctlog_types.go
```diff
@@ -22,24 +22,20 @@ import (
...
-	Signer CTlogSigner `json:"signer"`
 
-	// List of secrets containing root certificates that are acceptable to the log.
-	// The certs are served through get-roots endpoint. Optional in mirrors.
-	//+optional
-	// +listType=atomic
-	RootCertificates []SecretKeySelector `json:"rootCertificates,omitempty"`
+	// Logs defines the list of certificate transparency logs (active and frozen shards).
+	// Each entry represents either the active log or a frozen shard.
+	// +optional
```
bouskaJ:
```suggestion
// +  required
// + kubebuilder:validation:MinItems=1
```

api/v1/ctlog_types.go
```diff
@@ -22,24 +22,20 @@ import (
 )
 
 // CTlogSpec defines the desired state of CTlog component
+// +kubebuilder:validation:XValidation:rule="has(self.logs) && size(self.logs) > 0",message="at least one log is required"
+// +kubebuilder:validation:XValidation:rule="!has(self.logs) || self.logs.filter(x, has(x.active) && x.active == true).size() == 1",message="exactly one log should be active"
 type CTlogSpec struct {
```
bouskaJ: move validation on the `logs` field

api/v1/ctlog_types.go
```diff
@@ -105,6 +93,84 @@ type CTlogPKCS11Config struct {
...
+// +kubebuilder:validation:XValidation:rule="!has(self.signer) || !has(self.signer.file) || !has(self.signer.file.publicKeyRef) || has(self.signer.file.privateKeyRef)",message="privateKeyRef is required when publicKeyRef is set (CTFE validates key consistency for all logs, including readonly)"
+// +kubebuilder:validation:XValidation:rule="(has(self.active) && self.active == true) || has(self.logId)",message="logId is required for non-active logs"
+// +kubebuilder:validation:XValidation:rule="(has(self.active) && self.active == true) || has(self.signer)",message="signer is required for non-active logs"
+// +kubebuilder:validation:XValidation:rule="(has(self.active) && self.active == true) || (has(self.rootCerts) && size(self.rootCerts) > 0)",message="rootCerts is required for non-active logs"
+// +kubebuilder:validation:XValidation:rule="!(has(self.active) && self.active == true && (has(self.readonly) && self.readonly == true))",message="a log cannot be both active and readonly"
+// +kubebuilder:validation:XValidation:rule="!(has(self.active) && self.active == true && (has(self.mirror) && self.mirror == true))",message="a log cannot be both active and mirror"
```
bouskaJ: There is a room for simplification. Try patterns like
`(has(self.active) && self.active == true) || has(self.logId)` -> `self.?active.orValue(false) || has(self.logId)`

`// +kubebuilder:validation:XValidation:rule="!self.?active.orValue(false) || (!self.?readonly.orValue(false) && !self.?mirror.orValue(false))",message="an active log cannot be readonly or a mirror"`

internal/testing/fuzzer/time.go
bouskaJ: Why we have a new fuzzer, one is already implemented as a default fuzzer https://github.com/securesign/secure-sign-operator/blob/f80542ac959df93134a50a683db4618f4bcc4c94/internal/testing/fuzzer/time.go

tommyd450 (approved):

api/v1alpha1/tuf_conversion.go
```diff
@@ -25,6 +25,19 @@ func (src *Tuf) ConvertTo(dstRaw conversion.Hub) error {
...
+	if len(dst.Spec.Ctlog) > 0 && dst.Spec.Ctlog[0].URL == "///trusted-artifact-signer" { //nolint:goconst
+		if len(restored.Spec.Ctlog) == 0 || restored.Spec.Ctlog[0].Ref == nil {
+			// Fresh conversion: set Ref using the source Tuf's name and namespace
+			dst.Spec.Ctlog[0].Ref = &rhtasv1.ServiceReferenceRef{
+				Name:      src.Name,
+				Namespace: src.Namespace,
+			}
+			dst.Spec.Ctlog[0].URL = ""
+		}
+	}
```
bouskaJ: Do not set Ctlog ref namespaceName to the Tuf's resource name:
- v1aplha API has static prefix binding (https://github.com/securesign/secure-sign-operator/blob/main/api/v1alpha1/common.go#L91) so users do not expect autoloading
- ctlog instance can have different name (not created by the Securesign umbrella). This silently breaks the default (static) connection.

PR #2340, opened 2026-09-18
fix: revert alphav1 updates

bouskaJ:


osmman (approved):

osmman (changes requested):
There still changes in v1alpha1 which broke upgrade

osmman (approved):

PR #2360, opened 2026-09-21
fix: migration fixes for ConsoleCLIDownload cache bypass and cli-server cleanup

sampras343:
## What

Two fixes for the 1.4.x to 1.5.0 upgrade migration path.

### 1. Bypass cache for ConsoleCLIDownload in migration

Added `ConsoleCLIDownload` to the manager's `DisableFor` cache list so the migration component uses direct API calls instead of the cached client.

### 2. Remove legacy cli-server resources on upgrade

Extended the `MigrationComponent` to delete the obsolete cli-server Deployment, Service, Ingress, and `trusted-artifact-signer` namespace on operator startup. All deletions are guarded by `app.kubernetes.io/part-of: trusted-artifact-signer` label check.

### Cache bypass

The migration component only has `get`/`delete` RBAC on specific `ConsoleCLIDownload` resource names (e.g., `cosign`, `rekor-cli`). The manager's cached client requires `list`/`watch` permissions to populate its informer cache — permissions the migration intentionally does not have to minimize RBAC exposure. Without bypassing the cache, `Get` calls return `NotFound` even though the resources exist on the cluster, causing the migration to silently skip deletion.

### cli-server cleanup

In 1.4.x, the operator created a cli-server Deployment, Service, and Ingress in the `trusted-artifact-signer` namespace to serve CLI binaries from inside the cluster. In 1.5.0, CLI downloads were moved to the Red Hat Developer Portal — the cli-server is no longer needed. However, the upgrade does not remove these resources, leaving stale binaries from the old release exposed on the cluster. This can lead to users downloading unsupported or vulnerable CLI versions.


internal/clidownload/component.go
```diff
@@ -59,7 +79,52 @@ func (c *MigrationComponent) Start(ctx context.Context) error {
...
+		}
+
+		if err := c.Client.Delete(ctx, obj); err != nil {
+			c.Log.Error(err, "failed to delete legacy cli-server resource", "name", obj.GetName())
+			continue
+		}
+		c.Log.Info("deleted legacy cli-server resource", "kind", obj.GetObjectKind().GroupVersionKind().Kind, "name", obj.GetName())
+	}
+
+	// Delete the namespace last, only if empty and owned by this operator.
```
osmman: lets be defender and **do not delete** namespace. It is problematic because in our oficial documentation we references `trusted-artifact-signer` namespace as target for TAS deployment, plus I know that multiple existing deployment already uses this namespace.

internal/clidownload/component.go
```diff
@@ -9,11 +9,24 @@ import (
...
+const (
+	cliServerNamespace = "trusted-artifact-signer"
+	cliServerName      = "cli-server"
+)
+
 //+kubebuilder:rbac:groups=console.openshift.io,resources=consoleclidownloads,resourceNames=cosign;rekor-cli;gitsign;ec;fetch-tsa-certs;createtree;updatetree;tuftool,verbs=get;delete
+//+kubebuilder:rbac:groups=apps,resources=deployments,resourceNames=cli-server,verbs=get;delete
+//+kubebuilder:rbac:groups=core,resources=services,resourceNames=cli-server,verbs=get;delete
+//+kubebuilder:rbac:groups=networking.k8s.io,resources=ingresses,resourceNames=cli-server,verbs=get;delete
+//+kubebuilder:rbac:groups=core,resources=namespaces,resourceNames=trusted-artifact-signer,verbs=get;delete
```
osmman: remove **delete** role

osmman (changes requested):

internal/clidownload/component.go
osmman: It is better to keep that namespace in cluster even if it is empty. Users can delete it manually if they wanted.

sampras343: Agreed and fixed.

internal/clidownload/component.go
sampras343: Done

config/rbac/role.yaml
```diff
@@ -46,6 +46,15 @@ rules:
   - persistentvolumeclaims/finalizers
   verbs:
   - update
+- apiGroups:
+  - ""
+  resourceNames:
+  - cli-server
```
osmman: This is unnecessary, operator already has rights to get and delete any service, deployment and ingress in cluster.

sampras343: fixed

osmman (approved):

